PR #14, opened 2020-02-10
Fix the reporting issue in MEC by setting cmdID to 0 for all the

yalcinmelihyasin:
initial commands while transforming commands.

b/148923131

### DIFF
--- a/gapis/api/vulkan/find_issues.go
+++ b/gapis/api/vulkan/find_issues.go
@@ -63,7 +63,6 @@ func isValidationLayer(n string) bool {
 type findIssues struct {
 	replay.EndOfReplay
 	state           *api.GlobalState
-	numInitialCmds  int
 	issues          []replay.Issue
 	reportCallbacks map[VkInstance]VkDebugReportCallbackEXT
 }
@@ -71,7 +70,6 @@ type findIssues struct {
 func newFindIssues(ctx context.Context, c *capture.GraphicsCapture, numInitialCmds int) *findIssues {
 	t := &findIssues{
 		state:           c.NewState(ctx),
-		numInitialCmds:  numInitialCmds,
 		reportCallbacks: map[VkInstance]VkDebugReportCallbackEXT{},
 	}
 	t.state.OnError = func(err interface{}) {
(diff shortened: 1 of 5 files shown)


gapis/api/vulkan/find_issues.go
```diff
@@ -286,18 +284,19 @@ func (t *findIssues) Flush(ctx context.Context, out transform.Writer) error {
 			var issue replay.Issue
 			msg := eMsg.GetMsg()
 			label := eMsg.GetLabel()
-			if int(label) < t.numInitialCmds {
+			issue.Command = api.CmdID(label)
+			issue.Severity = service.Severity(uint32(eMsg.GetSeverity()))
+
+			if label == 0 {
```
hevrard: Should this be `int(label) == 0` ?

gapis/api/vulkan/find_issues.go
```diff
@@ -286,18 +284,19 @@ func (t *findIssues) Flush(ctx context.Context, out transform.Writer) error {
...
 				// The debug report is issued for state rebuilding command
 				// TODO: Fix all the errors reported for initial commands.
-				issue.Command = api.CmdID(0)
-				issue.Error = fmt.Errorf("[State rebuilding command, linearized ID: %d]: %s", label, msg)
+				issue.Error = fmt.Errorf("[State rebuilding command : %s", msg)
```
hevrard: We don't have the id/index of the state rebuilding command that triggers an error. Would there be a way to keep this information?
ALso, missing closing `]`

gapis/api/vulkan/find_issues.go
yalcinmelihyasin: No, previously it was casted into int becausethe type of t.numInitialCmds were int. Currently you don't need it. It's original type is uint64 and it is better to compare as it is.

gapis/api/vulkan/find_issues.go
yalcinmelihyasin: I will add the `]`.

We don't have because it actually does not mean anything. They are the commands produced by AGI to rebuild state rather than the real commands. Currently, they have a command id 0. Please refer @pmuetschard for further information.

PR #15, opened 2020-02-10
Disable GLES support in AGI.

pmuetschard:


gapis/trace/android/trace.go
```diff
@@ -268,9 +268,11 @@ func NewTracer(dev bind.Device) tracer.Tracer {
 // TraceConfiguration returns the device's supported trace configuration.
 func (t *androidTracer) TraceConfiguration(ctx context.Context) (*service.DeviceTraceConfiguration, error) {
 	apis := make([]*service.TraceTypeCapabilities, 0, 3)
+	/* Disabled in AGI. Use GAPID instead.
```
pau-baiget: Shouldn't we just remove this code??

pmuetschard: Eventually, yes, but we should do that at the time we remove all the gles code. Right now it serves as an explanation of why this tracer doesn't claim to support gl.

PR #16, opened 2020-02-10
Create Pipeline Strip in Pipeline View

hliatis:
Bugs: 147872472, 147872901, 147870682

### DIFF
--- a/gapic/src/main/com/google/gapid/views/PipelineView.java
+++ b/gapic/src/main/com/google/gapid/views/PipelineView.java
@@ -15,6 +15,7 @@
  */
 package com.google.gapid.views;
 
+import static com.google.gapid.util.Colors.lerp;
 import static com.google.gapid.util.Loadable.MessageType.Error;
 import static com.google.gapid.util.Loadable.MessageType.Info;
 import static com.google.gapid.util.Logging.throttleLogRpcError;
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -101,12 +112,17 @@ public PipelineView(Composite parent, Models models, Widgets widgets) {
     loading = LoadablePanel.create(this, widgets,
         panel -> createComposite(panel, new FillLayout(SWT.VERTICAL)));
 
-    stagesContainer = createComposite(loading.getContents(), new FillLayout());
+    stagesContainer = createComposite(loading.getContents(), new GridLayout(1, false));
+
+    RGB selectRGB = getDisplay().getSystemColor(SWT.COLOR_BLUE).getRGB();
+    RGB unselectRGB = getDisplay().getSystemColor(SWT.COLOR_LIST_BACKGROUND).getRGB();
+    hoverColor = new Color(getDisplay(), lerp(selectRGB, unselectRGB, 0.5f));
```
hliatis: Can I get a Windows user to do a comparison with 0.25f instead? Attached is how it looks on linux with 50% and 25%

Screenshot from 2020-02-10 12-32-30
Screenshot from 2020-02-10 12-32-55

johnplate: This looks different even on Linux, for example if you use a dark theme. We address the color theme issue separately, please use 0.5 for now.

PR #18, opened 2020-02-10
Force prerelease driver for tracing and replay

apbodnar:
Also requires that the device support GPU profiling in prep for the
validation UI

### DIFF
--- a/gapii/client/adb.go
+++ b/gapii/client/adb.go
@@ -72,6 +72,23 @@ func Start(ctx context.Context, p *android.InstalledPackage, a *android.Activity
 		abi = p.Device.Instance().GetConfiguration().PreferredABI(nil)
 	}
 
+	driver, err := d.PrereleaseGraphicsDriver(ctx)
+	if err != nil {
+		return nil, nil, log.Err(ctx, err, "Failed to locate pre-release driver package")
+	}
+
+	cleanup := app.Cleanup(func(ctx context.Context) {})
+	// Force the traced app to use the pre-release driver, or fail early
+	if driver.Package != "" && a != nil && a.Package != nil {
+		nextCleanup, err := adb.SetupPrereleaseDriver(ctx, d, a.Package)
+		cleanup = cleanup.Then(nextCleanup)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


core/os/android/adb/device.go
```diff
@@ -114,18 +113,18 @@ func Devices(ctx context.Context) (DeviceList, error) {
 	return out, nil
 }
 
-func SetupPrereleaseDriver(ctx context.Context, d Device, p *android.InstalledPackage) (app.Cleanup, error) {
+func SetupPrereleaseDriver(ctx context.Context, d Device, packageName string) (app.Cleanup, error) {
```
chrisforbes: I'm not sure about this part of the change. Passing an entire InstalledPackage here makes this harder to misuse.

gapii/client/adb.go
```diff
@@ -72,6 +72,23 @@ func Start(ctx context.Context, p *android.InstalledPackage, a *android.Activity
 		abi = p.Device.Instance().GetConfiguration().PreferredABI(nil)
 	}
 
+	driver, err := d.GraphicsDriver(ctx)
```
lpy: What's returned here does not necessarily match what we use in `SetupPrereleaseDriver` function. We should have a separate function that returns the prerelease driver apk without the override. The override should be removed or renamed in the future because it's misleading now

gapis/perfetto/android/trace.go
```diff
@@ -62,6 +62,10 @@ func hasRenderStageEnabled(perfettoConfig *perfetto_pb.TraceConfig) bool {
 
 // SetupProfileLayersSource configures the device to allow packages to be used as layer sources for profiling
 func SetupProfileLayersSource(ctx context.Context, d adb.Device, packageName string, abi *device.ABI) (app.Cleanup, error) {
+	cleanup, err := adb.SetupPrereleaseDriver(ctx, d, packageName)
```
chrisforbes: If you want the entire android.InstalledPackage here, you can get that for GAPIR from gapidapk.EnsureInstalled(...).apk

May be more replumbing than you want to do though..

lpy:
I don't think this patch is ready to be submitted.

PR #20, opened 2020-02-11
Fix vulkan event linking issue when zoomed too out.

stellama0208:
- Allow overriding quantized slices's SQL query handle.
 - Query submission_id for quantized GPU Queue slices.
 - Related ticket: http://b/148952158.


stellama0208:
Previous PR in GAPID repo: https://github.com/google/gapid/pull/3767

PR #25, opened 2020-02-11
Fix Vulkan sample cube distortion.

lpy:
Previously the submission will wait inside BOTTOM_OF_PIPE for the
swapchain_image_ready_semaphores. This means that it will do all the rendering
and writing out to the image before the swapchain image is actually ready.
Instead we need to wait for the submission inside TOP_OF_PIPE so that the work
is blocked until the image is ready.

Bug: b/149113751


lpy:
PTAL!

lpy:
mac build is soooooooo sloooooooow

PR #29, opened 2020-02-11
Revert "Fix the reporting issue in MEC by setting cmdID to 0 for all the"

apbodnar:
Reverts google/agi#14

This change broke replay profiling.  We use `numInitialCommands` to set the wait fence in the correct place in the command stream.


apbodnar:
Closing as a fix PR has been opened

PR #30, opened 2020-02-11
Have the enum tooltips display in Pipeline View

hliatis:
Bug: 148873707


gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -448,6 +453,15 @@ public void update(ViewerCell cell) {
 
                 super.update(cell);
               }
+
+              @Override
+              public String getToolTipText(Object element) {
+                if (dv != null) {
```
hliatis: I am checking this out of paranoia. Let me know if it's unnecessary, and if I should remove it.

johnplate: It is correct and necessary.

PR #31, opened 2020-02-11
Grey-out KVPs in Pipeline View that depend on another value

hliatis:
Also adds a tooltip for affected KVPs

Note: For enum values, the enum tooltip will overtake the active/disabled tooltip when label is active.

BUG: 147872480


gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -364,6 +364,14 @@ private Composite createStage(Composite parent, API.Stage currentStage) {
 
             Label keyLabel = withLayoutData( createBoldLabel(keyComposite, kvp.getName() + (kvp.getDynamic() ? "*:" : ":")),
                 new GridData(SWT.RIGHT, SWT.CENTER, true, true));
+            
+            if (!kvp.getDependee().equals("")) {
+              keyLabel.setToolTipText((kvp.getActive() ? "Activated" : "Deactivated") + " by " + kvp.getDependee());
+              
+              if (!kvp.getActive()) {
+                keyLabel.setForeground(getDisplay().getSystemColor(SWT.COLOR_TITLE_INACTIVE_FOREGROUND));
```
johnplate: This does not work on Windows, the color looks unchanged.

hliatis: Do they change with setEnabled()?

johnplate: Yes

hliatis: I went with setEnabled() to be safe, but with tables I went with DARK_GRAY. If the system doesn't display tables as expected, then I strongly suggest to keep it but add a message about deactivation in the group title like we do with dynamic tables.

johnplate: That's not a good solution as a fixed color does not play well with different color themes. Also under Linux currently the label background color seems to be slightly different than the composite background.
Maybe we just keep it for now until we have discussed a general solution for the color themes.

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -378,9 +387,15 @@ private Composite createStage(Composite parent, API.Stage currentStage) {
               withLayoutData( createLink(valueComposite,"<a>" + dv.displayValue + "</a>", e -> models.follower.onFollow(dv.link)),
                   new GridData(SWT.LEFT, SWT.CENTER, true, true));
             } else {
-              Label valueLabel  = withLayoutData( createLabel(valueComposite, dv.displayValue),
+              Label valueLabel = withLayoutData( createLabel(valueComposite, dv.displayValue),
                   new GridData(SWT.LEFT, SWT.CENTER, true, true));
-              valueLabel.setToolTipText(dv.tooltipValue);
+
+              if (kvp.getActive()) {
+                valueLabel.setToolTipText(kvp.getDependee().equals("") ? "Activated by " + kvp.getDependee() : dv.tooltipValue);
```
johnplate: Something is broken here somewhere. Looks like all active KVP values now have a "Activated by" tooltip.

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: I can't use setEnabled() with tables. At least on Linux, it doesn't gray out the cells.

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: The ternary was flipped. Should be fixed now.

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: There is also not coloring the table cells at all and instead using the group title to say it is disabled like we do when table data is set dynamically

gapic/src/main/com/google/gapid/views/PipelineView.java
johnplate: Seems to be fixed

gapic/src/main/com/google/gapid/views/PipelineView.java
johnplate: Ok, we keep it for now.

PR #32, opened 2020-02-11
Fix replay profiling after command id changes

apbodnar:


gapis/api/vulkan/wait_for_perfetto.go
```diff
@@ -46,7 +47,12 @@ func addVkDeviceWaitIdle(ctx context.Context, out transform.Writer) {
...
 
 func (t *WaitForPerfetto) waitTest(ctx context.Context, id api.CmdID, cmd api.Cmd) bool {
-	if id == t.cmdId {
+	// Set the command id we care about once we're in the "real" commands
+	if id.IsReal() && t.cmdId == api.CmdNoID {
+		t.cmdId = id
+	}
+
+	if id.IsReal() && id == t.cmdId {
 		return true
```
chrisforbes: combine these? This second condition is going to be true exactly once, in the same invocation as the first is?

PR #38, opened 2020-02-12
GPU Counter Selection Dialog Improvements

pmuetschard:
- Respect the "select by default" property
- Allow selecting the default counters
- Allow filtering the counters


pmuetschard:
Yes, it looks like a certain producer doesn't yet populate this value.

PR #41, opened 2020-02-12
Fix vkEnumerateDeviceExtensionProperties in CPUTiming layer

rschiu:
The layer added new extension and increased the pPropertyCount by 2.
pPropertyCount is expected to be reduced back to the original value
when vkEnumerateDeviceExtensionProperties is called.  Therefore, it
needs to increased back to the requested value.

Bug: b/149116549


lpy:
Bug: b/xxxxx

rschiu:
Added Bug number

lpy:
`VK_EXT_debug_utils` is an instance extension, let's do a follow up to fix this

PR #42, opened 2020-02-13
Clamp the number that is entered into a spinner.

stellama0208:


stellama0208:
Had some idea and tried it out after seeing the discussion in #34. This change will get number clamped between min&max, immediately after user entering the number with keyboard. Maybe this will reduce some surprise.

stellama0208:
> I would say no to this change as I wouldn't like the spinner changing while I'm typing in it.

OK, then I will close this PR for now, we can later find out a good way to handle it.

PR #44, opened 2020-02-13
Fix error where last row tooltip in PV tables were showing

hliatis:
Bug: b/148873707


johnplate:
You can ignore that CI-linux-test failed.

PR #46, opened 2020-02-13
Fix the framelooping bug after the command id changes

yalcinmelihyasin:
Fixes the regression in frame looping caused by #14


yalcinmelihyasin:
Even though we cannot prove its working due to other bugs in frame looping, I would advise to merge this any way, as it is failing to export replay with frame looping without this fix.

yalcinmelihyasin:
@purvisa-at-google-com

PR #47, opened 2020-02-13
Fix the issue of first command of the trace has CmdNoID

yalcinmelihyasin:


AWoloszyn:
Looks good, someone else should approve though

PR #48, opened 2020-02-13
Remove the numInitialCmds from transforms

yalcinmelihyasin:
as it is not needed now


AWoloszyn:
Seems ok, someone else should approve though

PR #49, opened 2020-02-13
Do the app cleanup for validation in the background context

silence-do-good:
Force stopping the app after validation is currently happening in the same context as the validation process. This is fine in the normal case, but if the validation is interrupted by switching to another device, the old context is cancelled, and while calling ForceStop(ctx, app), the cancelled context is used. The force stop hence, fails, and the app keeps running on the first device forever.
Doing the force stop on the background context, which never stops, seems to be one way to address this.

Bug: 149394413


silence-do-good:
Created a separate PR for this to not block the validation UI. We can track this and discuss alternate approaches separately.

PR #50, opened 2020-02-14
Add perfetto protos to gofuse.

lpy:
gofuse commnad has output all generated protobuf go files of the project to
gofuse directory, however, Perfetto protos are in the third_party directory
which is not scanned at all. This patch adds Perfetto directory in the mapping
path so that all Perfetto protos are output to gofuse directory and indexable
by IDE.

Bug: N/A
Test: bazel run //cmd/gofuse -- -dir GOFUSE_OUTPUT_DIR

### DIFF
--- a/cmd/gofuse/main.go
+++ b/cmd/gofuse/main.go
@@ -208,8 +208,17 @@ func run() error {
 		extMapping = append(extMapping, m...)
 	}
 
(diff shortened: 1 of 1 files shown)


lpy:
ptal

lpy:
With this patch, Intellij works like a charm on my Linux machine :)

PR #54, opened 2020-02-14
Draw vulkan event linking arrow based on slices' distance.

stellama0208:
- Extend the arrow to the start of the linked gpu queue slices group.
 - Bug: http://b/148964552.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/VulkanEventTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/VulkanEventTrack.java
@@ -53,7 +53,7 @@ public class VulkanEventTrack extends Track.WithQueryEngine<VulkanEventTrack.Dat
           "where ts >= %d - dur and ts <= %d order by ts";
   private static final String QUEUE_GROUP_START_SQL =
       "select submission_id, min(ts) start from gpu_track t left join gpu_slice s " +
-          "on (t.id = s.track_id) where t.name like \'queue%%\' group by submission_id";
+          "on (t.id = s.track_id) where t.scope = 'gpu_render_stage' group by submission_id";
   private static final String SLICES_WITH_DIST_SQL =
       "with basics as (" + SLICES_SQL + ")," +
       "queue_starts as ("+ QUEUE_GROUP_START_SQL + ") " +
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/perfetto/models/VulkanEventTrack.java
```diff
@@ -51,6 +51,14 @@
   private static final String SLICES_SQL =
       "select " + BASE_COLUMNS + " from %s " +
           "where ts >= %d - dur and ts <= %d order by ts";
+  private static final String QUEUE_GROUP_START_SQL =
+      "select submission_id, min(ts) start from gpu_track t left join gpu_slice s " +
+          "on (t.id = s.track_id) where t.name like \'queue%%\' group by submission_id";
```
chrisforbes: This should probably find the correct tracks by scope rather than by name?

rschiu: Yes, please never rely on the name of the track (or almost anything for that matter) as an identifier.

gapic/src/main/com/google/gapid/perfetto/models/VulkanEventTrack.java
```diff
@@ -95,13 +104,14 @@ public VulkanEventTrack(QueryEngine qe, GpuInfo.VkApiEvent vkApiEvent) {
             data.commandBuffers[i] = row.getLong(5);
             data.submissionIds[i] = row.getLong(6);
             data.args[i] = args.getOrDefault(row.getLong(7), ArgSet.EMPTY);
+            data.dists[i] = row.getLong(8);
```
rschiu: Maybe a comment explaining what dists is.

gapic/src/main/com/google/gapid/perfetto/models/VulkanEventTrack.java
stellama0208: Thanks for pointing this out!

gapic/src/main/com/google/gapid/perfetto/models/VulkanEventTrack.java
stellama0208: Thanks! Done in a new commit.

PR #56, opened 2020-02-17
Trigger a swarming test from Kokoro

hevrard:


hevrard:
I've moved the swarming logic into AGI source code, now only the APKs are stored on x20.

PTAL

PR #58, opened 2020-02-17
Fix the boundary issues in Commands()

yalcinmelihyasin:
b/149010824

### DIFF
--- a/gapis/resolve/commands.go
+++ b/gapis/resolve/commands.go
@@ -34,22 +34,22 @@ func Commands(ctx context.Context, p *path.Commands, r *path.ResolveConfig) (*se
 	if err != nil {
 		return nil, err
 	}
+	count := uint64(len(c.Commands))
+	if count == 0 {
+		return &service.Commands{List: []*path.Command{}}, nil
+	}
 	cmdIdxFrom, cmdIdxTo := p.From[0], p.To[0]
 	if len(p.From) > 1 || len(p.To) > 1 {
 		return nil, fmt.Errorf("Subcommands currently not supported for Commands") // TODO: Subcommands
 	}
-	count := uint64(len(c.Commands))
-	if count == 0 {
-		return nil, fmt.Errorf("No commands in capture")
-	}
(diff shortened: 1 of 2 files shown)


gapis/resolve/commands.go
```diff
@@ -45,11 +45,11 @@ func Commands(ctx context.Context, p *path.Commands, r *path.ResolveConfig) (*se
 	cmdIdxFrom = u64.Min(cmdIdxFrom, count-1)
 	cmdIdxTo = u64.Min(cmdIdxTo, count-1)
 	if cmdIdxFrom > cmdIdxTo {
-		cmdIdxFrom, cmdIdxTo = cmdIdxTo, cmdIdxFrom
```
purvisa-at-google-com: If these are inclusive boundaries you should state it with a comment!

gapis/resolve/commands.go
```diff
@@ -34,22 +34,22 @@ func Commands(ctx context.Context, p *path.Commands, r *path.ResolveConfig) (*se
...
 	}
-	count := uint64(len(c.Commands))
-	if count == 0 {
-		return nil, fmt.Errorf("No commands in capture")
-	}
 	cmdIdxFrom = u64.Min(cmdIdxFrom, count-1)
 	cmdIdxTo = u64.Min(cmdIdxTo, count-1)
 	if cmdIdxFrom > cmdIdxTo {
-		cmdIdxFrom, cmdIdxTo = cmdIdxTo, cmdIdxFrom
+		return nil, fmt.Errorf("Invalid command boundaries")
```
ben-clayton: nit: It would be good to keep with the standard, localized error messages. See `errPathOOB` for an example.

gapis/resolve/commands.go
```diff
@@ -34,22 +34,22 @@ func Commands(ctx context.Context, p *path.Commands, r *path.ResolveConfig) (*se
...
 	}
-	count = cmdIdxTo - cmdIdxFrom
+	count = cmdIdxTo - cmdIdxFrom + 1
```
ben-clayton: Good catch

PR #61, opened 2020-02-18
Use Android NDK r21 and VK_LAYER_KHRONOS_validation

hevrard:
Upgrade the NDK to r21. This let us use the unified
VK_LAYER_KHRONOS_validation meta validation layer, which is available
on both Desktop and Android.

This change does a bit of refactoring of the code handling the
validation layers, to make use of the new unified one.

Bug: b/148519068

### DIFF
--- a/gapis/api/templates/vulkan_gfx_api_extras.tmpl
+++ b/gapis/api/templates/vulkan_gfx_api_extras.tmpl
@@ -73,7 +73,7 @@
     return std::any_of(std::begin(kValidationLayerNames),
                        std::end(kValidationLayerNames),
                        [layer](const char* val) -> bool {
-                         return strcmp(layer, val) == 0;
+                         return std::strcmp(layer, val) == 0;
                        });
   }
 ¶
--- a/gapis/api/templates/vulkan_gfx_api_extras.tmpl
+++ b/gapis/api/templates/vulkan_gfx_api_extras.tmpl
@@ -55,8 +55,11 @@
 ¶
   const char kVirtualSwapchainLayerName[] = "VirtualSwapchain";
   const char kGraphicsSpyLayerName[] = "GraphicsSpy";
-  const char kValidationMetaLayerName[] = "VK_LAYER_LUNARG_standard_validation";
   const char* kValidationLayerNames[] = {
+    // Meta layers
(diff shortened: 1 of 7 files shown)


gapis/api/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -55,8 +55,11 @@
 ¶
   const char kVirtualSwapchainLayerName[] = "VirtualSwapchain";
   const char kGraphicsSpyLayerName[] = "GraphicsSpy";
-  const char kValidationMetaLayerName[] = "VK_LAYER_LUNARG_standard_validation";
   const char* kValidationLayerNames[] = {
+    // Meta layers
+    "VK_LAYER_KHRONOS_validation",
+    "VK_LAYER_LUNARG_standard_validation",
```
yalcinmelihyasin: Do we need both?

hevrard: Yes, because we use these names to search for validation layers loaded by the app, and it may use any validation layer combination.

gapis/api/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -65,6 +68,14 @@
   };
   const char kDebugReportExtensionName[] = "VK_EXT_debug_report";
 
+¶
+  bool isValidationLayerName(const char* layer) {
+    return std::any_of(std::begin(kValidationLayerNames),
+                       std::end(kValidationLayerNames),
+                       [layer](const char* val) -> bool {
+                         return strcmp(layer, val) == 0;
```
johnplate: std::strcmp

johnplate:
And we should announce the PR with the team before merging it, so everyone knows that they have to update their NDKs.

gapis/api/templates/vulkan_gfx_api_extras.tmpl
hevrard: done

gapis/api/templates/vulkan_gfx_api_extras.tmpl
yalcinmelihyasin: :+1:

PR #63, opened 2020-02-18
Make slice renaming robust to missing mappings

chrisforbes:


gapis/trace/android/adreno/profiling_data.go
```diff
@@ -196,6 +196,10 @@ func processGpuSlices(ctx context.Context, processor *perfetto.Processor, captur
 			Value: &service.ProfilingData_GpuSlices_Slice_Extra_IntValue{IntValue: uint64(hwQueueIds[i])},
 		})
 
+		if names[i] == "Surface" && groupIds[i] != -1 {
```
lpy: We probably can't rely on the name

chrisforbes: Long term, no. This code is specific to the current Adreno producers though.

PR #66, opened 2020-02-18
Add the option to allow users to mark flags along the timeline

silence-do-good:
This PR adds a new feature that allows users to drop markers along the timeline. 

To add a flag:
 Click in the bottom half on the timeline panel

To toggle drawing vertical line along the canvas:
 Click again on the flag

To remove the flag:
 Right click on the flag

Bug: 143048804


gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
```diff
@@ -359,6 +360,19 @@ protected void finishSelection(int mods) {
 
   @Override
   public Hover onMouseMove(Fonts.TextMeasurer m, double x, double y, int mods) {
+    if (mouseMode == MouseMode.Flagging) {
+      return new Hover() {
```
stellama0208: It would be great if you could keep this onMouseMove method concise, you could return this Hover object through a private method.

gapic/src/main/com/google/gapid/perfetto/views/State.java
```diff
@@ -122,6 +129,28 @@ public long deltaPxToDuration(double px) {
     return Math.round(px * nanosPerPx);
   }
 
+  protected SortedMap<Double, Long> searchForFlag(double x) {
+    return flags.subMap(x - 12, x + 12);
```
stellama0208: Currently, there's a UX problem that when a flag is created and some zoomings happens afterwards, it may be difficult to remove a flag. I think it is related to this fixed number 12, I would suggest using something like 3 * deltaPxToDuration(1) instead, to denote a dynamic 3 pixel flexible range.

gapic/src/main/com/google/gapid/perfetto/views/StyleConstants.java
```diff
@@ -43,6 +43,7 @@
   public static final double SELECTION_THRESHOLD = 0.333;
   public static final double ZOOM_FACTOR_SCALE = 0.05;
   public static final double ZOOM_FACTOR_SCALE_DRAG = 0.01;
+  public static final double FLAGS_Y = 24;
```
stellama0208: Nit: I would suggest put this one in RootPanel.java along with those highlight constants, since this one is not shared cross different files, or extremely global, like the other constants here.

gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
```diff
@@ -466,6 +481,15 @@ protected void preMainUiRender(RenderContext ctx, Repainter repainter) {
       }
     }
 
+    private void renderFlags(RenderContext ctx) {
+      state.getFlags().forEach(e -> {
```
stellama0208: Maybe use a more meaningful 't' instead of 'e' as the variable name. 'e' is used in lambdas across gapid because they means "events" in those contexts.

gapic/src/main/com/google/gapid/perfetto/views/State.java
```diff
@@ -68,6 +73,7 @@
   private final AtomicInteger lastSelectionUpdateId = new AtomicInteger(0);
   private HashMultimap<Long, Long> selectedThreads;     // upid -> utids
   private TimeSpan highlight = TimeSpan.ZERO;
+  private TreeMap<Double, Long> flags;
```
stellama0208: If you could provide a comment briefly tell what's the key-value pair, like in L74, it would be great.

rschiu:
What's the reason to create a new mode?  Perfetto UI does this pretty nicely by letting someone add flags on the timeline bar.

silence-do-good:
> What's the reason to create a new mode? Perfetto UI does this pretty nicely by letting someone add flags on the timeline bar.

So by default, if we drag on the timeline in any mode, it becomes a timeline select. If we allow simply clicking on the timeline to drop a flag, it conflicts with the timeline select and can cause unintentional bad experience. 
We can hide the flagging mode and just press a hotkey and click to drop flags, but if we want to drop multiple flags, we need to press every single time.
Considering these things, I decided to make it a mode so that the users are aware that they are in the flagging mode and also lets them know about such a feature easily.

stellama0208:
> > What's the reason to create a new mode? Perfetto UI does this pretty nicely by letting someone add flags on the timeline bar.
> 
> So by default, if we drag on the timeline in any mode, it becomes a timeline select. If we allow simply clicking on the timeline to drop a flag, it conflicts with the timeline select and can cause unintentional bad experience.
> We can hide the flagging mode and just press a hotkey and click to drop flags, but if we want to drop multiple flags, we need to press every single time.
> Considering these things, I decided to make it a mode so that the users are aware that they are in the flagging mode and also lets them know about such a feature easily.
> 
> The other advantage here is that once in flagging mode, the user can click anywhere on the canvas, to flag the corresponding time. So if you are looking at an event and just want to flag it, you enter the flag mode, click the interesting point and it will automatically drop a flag on the timeline.

A piece of information: the click behavior and drag behavior is separate in agi system profiler, so it's feasible to make flagging a click event in timeline but with no bad user experience (click for flagging and de-selection, drag for time highlight). The click event handling in timeline panel lies at Rootpanel.java L384, if you wanna go this way you can do the change there. 

A new data point for UX preference: I would prefer a flagging in timeline panel and no new mode as well. Because it's simple and fast, I don't have to switch mode to do it and I can click whenever I want.

gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
silence-do-good: Done

gapic/src/main/com/google/gapid/perfetto/views/State.java
silence-do-good: Done

silence-do-good:
> A piece of information: the click behavior and drag behavior is separate in agi system profiler, so it's feasible to make flagging a click event in timeline but with no bad user experience (click for flagging and de-selection, drag for time highlight). The click event handling in timeline panel lies at Rootpanel.java L384, if you wanna go this way you can do the change there.
> 
> A new data point for UX preference: I would prefer a flagging in timeline panel and no new mode as well. Because it's simple and fast, I don't have to switch mode to do it and I can click whenever I want.

The advantage here is that, when you're looking at events, doing selection or whatnot, you dont need to go to the timeline panel and click at some time hoping to match it with the event you are looking at. You can instead press 5, click at the event you're looking at and then press 1/2 again to go back to selection/ pan mode. Even if you had some selection previously, it wouldn't be disturbed and I verified this.

stellama0208:
> 
> The advantage here is that, when you're looking at events, doing selection or whatnot, you dont need to go to the timeline panel and click at some time hoping to match it with the event you are looking at. You can instead press 5, click at the event you're looking at and then press 1/2 again to go back to selection/ pan mode. Even if you had some selection previously, it wouldn't be disturbed and I verified this.
> 
> We can **add** your suggestion as well. In addition to having a mode, by default, if you click in the timeline, it will add a flag.

This scenario makes sense to me. Then I'm fine with adding a new mode. :)

gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
```diff
@@ -212,6 +212,10 @@ public Dragger onDragStart(double sx, double sy, int mods) {
       return Dragger.NONE;
     }
 
+    if (mouseMode == MouseMode.Flagging) {
```
stellama0208: We shouldn't do an early return here, just let returning Dragger.NONE happen at Line 237. 
Reason: the next several lines will handle mode overriding: basically we should let [shift-dragging -> selection; ctrl-dragging -> time higlight; space-dragging- > pan] to work in any modes. Early return here will disable the mode overriding in flagging mode.

All other code looks fine for me now, just this one last bit introduced in the new commit.

rschiu:
> So by default, if we drag on the timeline in any mode, it becomes a timeline select. If we allow simply clicking on the timeline to drop a flag, it conflicts with the timeline select and can cause unintentional bad experience.

Frankly, I didn't even know the drag behavior is a feature.  Personally, I would prefer to remove it unless someone actually uses that feature and object to its removal.  If we do keep this feature, then the discovery needs to be made better.  For example, the mouse icon can change while on the timeline, but that's another bug.

I would prefer to remove the drag behavior, but if we do keep it, then what Stella suggested would work well.  Ideally, we should also do something like Perfetto UI to make discovery better.

> We can hide the flagging mode and just press a hotkey and click to drop flags, but if we want to drop multiple flags, we need to press every single time.

Nah, don't do that.

> Considering these things, I decided to make it a mode so that the users are aware that they are in the flagging mode and also lets them know about such a feature easily.

The mode system is there for backward compatibility for some systrace user.  I personally do not use modes and would prefer to be able avoid modes going forward.  New features should avoid modes as much as possible.

> The other advantage here is that once in flagging mode, the user can click anywhere on the canvas, to flag the corresponding time. So if you are looking at an event and just want to flag it, you enter the flag mode, click the interesting point and it will automatically drop a flag on the timeline.

That's definitely a desirable feature, but not quite enough to overcome the disadvantages.  If we implement something like Perfetto UI where we draw a vertical line, I am not sure if it's necessary.

If user feedback suggest that is important and we couldn't think of better ways to implement this other than modes, then we can consider adding the mode.  But that should be done after we make this feature available without mode.  Also, if we do add mode, we need to avoid using the modifier key.  A big advantage of using modes is to avoid the need to use modifier keys.  Making ctrl+click remove the flag removes a key advantage of using modes and make the UX a lot more complicated.

gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
silence-do-good: Yes I intentionally made the override in flagging mode. Because I accidentally went into timing select when flagging because of misclicks and that started becoming a bit messy. But we can change this behavior after we finalize something for the UX.

silence-do-good:
Updated the PR based on our meeting.

WIP:
Adding hover flag similar to perfetto.

silence-do-good:
Done with the hover work. You'll be able to see the flags hover in the bottom half of the timeline panel. Please take a look @stellama0208 @rschiu

gapic/src/main/com/google/gapid/util/Colors.java
```diff
@@ -26,6 +26,8 @@
   public static final int DARK_LUMINANCE8_THRESHOLD = 165;
   public static final RGB BLACK = new RGB(0, 0, 0);
   public static final RGB WHITE = new RGB(255, 255, 255);
+  public static final RGBA BLACK_RGBA = new RGBA(0, 0, 0, 255);
```
stellama0208: AGI's color management system has moved to perfetto/views/StyleConstants.java, please do this part there.

gapic/src/main/com/google/gapid/perfetto/canvas/Panel.java
```diff
@@ -197,6 +203,12 @@ public boolean click() {
           boolean r1 = Hover.this.click(), r2 = onClick.getAsBoolean();
           return r1 || r2;
         }
+
+        @Override
+        public boolean rightClick() {
+          boolean r1 = Hover.this.rightClick(), r2 = onClick.getAsBoolean();
+          return r1 || r2;
```
stellama0208: Would suggest simply return Hover.this.rightClick() as well for the same reason. This function withClick() was designed as a helper function, so that we can add new actions to left-click with a lightweight lambda.

gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
```diff
@@ -196,6 +200,7 @@ public void render(RenderContext ctx, Repainter repainter) {
 
   protected abstract void preTopUiRender(RenderContext ctx, Repainter repainter);
   protected abstract void preMainUiRender(RenderContext ctx, Repainter repainter);
+  protected abstract void renderFlags(RenderContext ctx, Panel panel);
```
stellama0208: No need as an abstract method, refer to renderVSync.

gapic/src/main/com/google/gapid/perfetto/canvas/Panel.java
```diff
@@ -135,6 +135,7 @@ public default void stop() { /* empty */ }
 
     /** Returns whether the screen should be redrawn. */
     public default boolean click() { return false; }
+    public default boolean rightClick() { return click(); }
```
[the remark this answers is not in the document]
silence-do-good: If you right click on events, it currently goes to the selection. I didn't want to break this behavior by adding something new. I suggest keeping the behavior as is, and cleaning up post GDC when we can discuss and finalize if right click should be selection or not. What do you think?

gapic/src/main/com/google/gapid/perfetto/views/State.java
```diff
@@ -68,6 +73,9 @@
   private final AtomicInteger lastSelectionUpdateId = new AtomicInteger(0);
   private HashMultimap<Long, Long> selectedThreads;     // upid -> utids
   private TimeSpan highlight = TimeSpan.ZERO;
+  private TreeMap<Long, Boolean> flags;
+  private boolean flagHovered = false;
+  private double flagHoverXpos;
```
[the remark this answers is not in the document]
silence-do-good: Good idea. Matter of fact, I think the entire flag related stuff (including the TreeMap) can be moved to the RootPanel. I put the TreeMap here in State for the first version of the flag feature, since it involved using the TimelinePanel's onMouseMove() and RootPanel's rendering. But now, given that the entire flag mechanism happens within the RootPanel, I think its better to move everything there.

gapic/src/main/com/google/gapid/perfetto/canvas/Panel.java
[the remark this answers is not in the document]
stellama0208: Yes, I'm fine with further discussion. It's a detail compared to the great flag implementation.

gapic/src/main/com/google/gapid/util/Colors.java
silence-do-good: Done.

gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
```diff
@@ -359,27 +364,61 @@ protected void finishSelection(int mods) {
...
-          return true;
-        }
-        return false;
-      });
-    }
-    if (checkHighlightEdgeHovered(x)) {
-      result = result.withRedraw(Area.FULL);
+    if (y >= (timeline.getPreferredHeight()/2) && y <= timeline.getPreferredHeight() && x > LABEL_WIDTH) {
+      return flagHover(x);
+    } else {
```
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
silence-do-good: This needs to be abstract because it is similar to the preTop and preMain render functions and not the renderVSync. The flags have to be rendered after the top and bottom have finished rendering, for it to be on top of all other events. This means calling renderFlags from within the render() of RootPanel. We don't want flags in the Replay Profiler, so this has to be abstract and only the ForSystemTrace will implement it.

silence-do-good:
Added support for dark theme now and removed the State involvement in flags. @stellama0208  PTAL :)

gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
```diff
@@ -106,6 +119,36 @@ public void setSize(double w, double h) {
     state.setMaxScrollOffset(bottomHeight - h + topHeight);
   }
 
+  private SortedMap<Long, Boolean> searchForFlag(double x) {
+    long time = state.pxToTime(x);
+    // If the click falls within the flag icon's boundaries, the flag is considered to be hit.
+    // Leave an additional two pixels to the left of the flag to get a better hit box
+    long rightOffset = state.deltaPxToDuration(FLAG_WIDTH);
+    long leftOffset = state.deltaPxToDuration(2);
```
rschiu: optional: Might want to increase this a bit to 5 or 6.  It feels a little odd that there is such an imbalance between how close you can set a flag on the left vs on the right.  Ignore me if you don't feel the same.

silence-do-good: The imbalance is because the 'x' is the pole of the flag. The flag is facing right, so the the right offset is a bit more. The left offset of 2 pixels is just to make sure that clicking on the exact pole of the flag is considered as well.

gapic/src/main/com/google/gapid/perfetto/views/RootPanel.java
stellama0208: I still hold the opinion that it's closer to renderVsync than to preTop/preMain but it's not a blocker for this PR, so I'm fine with having it checked in.

- PreTop/preMain emphasize function's execution time, and what a real panel do at preTop time may indeed vary a lot so it's reasonable to declare them abstract. 
- RenderVsync/renderFlagBut emphasize the actual function of the module, if a panel decide to render flags, how they render wouldn't vary that much, so it's not that straightforward for me to mark it abstract. 

I'm not sure whether we want to toggle off flag function in replay profiler, but even if we do so, we can still override renderFlags() method to empty even if it's not declared as abstract.

PR #68, opened 2020-02-18
Fix GPU counter's config update issue.

stellama0208:
- Populate the remembered counter selection from the most updated
 perfetto settings, rather than from a possibly outdated copy.
 - Bug: http://b/149737345.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
@@ -488,8 +488,8 @@ public class TraceConfigDialog extends DialogBase {
 
           gpuCountersLabels[0] = createLabel(counterGroup, sGpu.getCounterIdsCount() + " selected");
           gpuCountersSelect = Widgets.createButton(counterGroup, "Select", e -> {
-            GpuCountersDialog dialog =
-                new GpuCountersDialog(getShell(), theme, caps, sGpu.getCounterIdsList());
+            List<Integer> currentIds = settings.perfetto().getGpuOrBuilder().getCounterIdsList();
+            GpuCountersDialog dialog = new GpuCountersDialog(getShell(), theme, caps, currentIds);
             if (dialog.open() == Window.OK) {
               List<Integer> newIds = dialog.getSelectedIds();
               settings.writePerfetto().getGpuBuilder()


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -488,8 +488,8 @@ public BasicInputArea(Composite parent, Settings settings, Theme theme,
 
           gpuCountersLabels[0] = createLabel(counterGroup, sGpu.getCounterIdsCount() + " selected");
```
silence-do-good: How is the CounterIdsCount() updated properly though?

stellama0208: It's updated on L499.

silence-do-good:
> I seems to work as intended, except in one use case:
> 
> * open the counter selection window, have some counters already selected
> * click on "none" to deselect all
> * click OK
> * re-open the counter selection window: the counters that were selected before are still selected
> 
> @stellama0208 any idea how to fix this?

stellama0208:
> I seems to work as intended, except in one use case:
> 
> * open the counter selection window, have some counters already selected
> * click on "none" to deselect all
> * click OK
> * re-open the counter selection window: the counters that were selected before are still selected
> 
> @stellama0208 any idea how to fix this?

Hey Hugues, thanks for testing out! 
In fact, this PR should cover this case, and I couldn't reproduce the error following these steps on my glinux and mac. To double check, I asked Adi to help test out this case and neither could he observe the same result.
Would you mind try it again Hugues? Thanks!

rschiu:
Is there another ARM device in mountain view?
I think what's happening is that the dialog is selecting the default counters when no counters are selected.

stellama0208:
> Is there another ARM device in mountain view?
> I think what's happening is that the dialog is selecting the default counters when no counters are selected.

Yes, it should be this reason. Just saw that AGI will select default counters when there's no selection at all. (According to code here: https://github.com/google/agi/blob/669abddb4690d26b2ae82f711f4940a4c8918132/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java#L811)

But this behavior is out of this bug's scope. Hugues, if you think it is OK, I will have this PR merged in and create a new bug tracking the desired UX for choosing default counters. :)

PR #69, opened 2020-02-18
Change the success validation icon to check mark

silence-do-good:
The current success validation icon is a smile image. It might be a bit informal to have a smiley, so changing it into a check.

Bug: 149783603


silence-do-good:
validation_check

PR #72, opened 2020-02-19
Consider if handle types match when replacing the replay handles

yalcinmelihyasin:
with trace handles

b/147979675

### DIFF
--- a/gapis/trace/android/adreno/profiling_data.go
+++ b/gapis/trace/android/adreno/profiling_data.go
@@ -54,6 +54,29 @@ func ProcessProfilingData(ctx context.Context, processor *perfetto.Processor, ca
 	return &service.ProfilingData{Slices: slices, Counters: counters}, nil
 }
 
+func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHandleType string, handleMapping *map[uint64][]service.VulkanHandleMappingItem) {
+	for i, v := range *replayHandles {
+		handles, ok := (*handleMapping)[uint64(v)]
+		if !ok {
+			log.E(ctx, "%v not found in replay: %v", replayHandleType, v)
+			continue
+		}
+
+		found := false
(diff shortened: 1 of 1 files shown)


gapis/trace/android/adreno/profiling_data.go
```diff
@@ -54,6 +54,29 @@ func ProcessProfilingData(ctx context.Context, processor *perfetto.Processor, ca
 	return &service.ProfilingData{Slices: slices, Counters: counters}, nil
 }
 
+func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHandleType string, handleMapping *map[uint64][]service.VulkanHandleMappingItem) {
```
chrisforbes: Not new in this change, but can we eventually be consistent about signedness of handle types?

gapis/trace/android/adreno/profiling_data.go
```diff
@@ -54,6 +54,29 @@ func ProcessProfilingData(ctx context.Context, processor *perfetto.Processor, ca
...
+	for i, v := range *replayHandles {
+		handles, ok := (*handleMapping)[uint64(v)]
+		if !ok {
+			log.E(ctx, "%v not found in replay: %v", replayHandleType, v)
+			continue
+		}
+
+		found := false
+		for _, handle := range handles {
```
chrisforbes: This seems OK given we expect the number of handle collisions to be very low.

gapis/trace/android/adreno/profiling_data.go
yalcinmelihyasin: I completely agree with you on this one but we have an issue there.

In all AGI, Vulkan Handles are treated as unsigned but in perfetto.proto, we don't have unsigned values. I think some preliminary work is required in perfetto side to send values unsigned.

I am happy to change this in another pr after that work has been done.

PR #73, opened 2020-02-19
Remove the issues in initial commands from report

yalcinmelihyasin:
and log them instead

b/149014498

### DIFF
--- a/gapis/api/vulkan/find_issues.go
+++ b/gapis/api/vulkan/find_issues.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/gapid/gapis/api"
 	"github.com/google/gapid/gapis/api/transform"
 	"github.com/google/gapid/gapis/capture"
+	"github.com/google/gapid/gapis/config"
 	"github.com/google/gapid/gapis/replay"
 	"github.com/google/gapid/gapis/replay/builder"
 	"github.com/google/gapid/gapis/service"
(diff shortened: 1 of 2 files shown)


AWoloszyn:
Can we also not just FIX these issues?

yalcinmelihyasin:
> Can we also not just FIX these issues?

Currently we don't have issues in our test apps but it does not mean we will not have in the future. I consider keeping this as if we get issues in the future, this log would warn us.

gapis/api/vulkan/find_issues.go
```diff
@@ -290,16 +290,12 @@ func (t *findIssues) Flush(ctx context.Context, out transform.Writer) error {
 			if issue.Command == api.CmdNoID {
 				// The debug report is issued for state rebuilding command
 				// TODO: Fix all the errors reported for initial commands.
-				// TODO: Provide a way for the UI to distinguish these issues
-				// from issues on command 0.
-				issue.Command = api.CmdID(0)
-				issue.Error = fmt.Errorf("[State rebuilding command : %s]	", msg)
+				log.E(ctx, "Error in state rebuilding command : %s", msg)
```
AWoloszyn: We should probably log with the same severity as the validation layers?

yalcinmelihyasin: I am not sure about this one. As these are our issues instead of the applications, maybe it might be better to consider them as errors regardless of severity. We may still print the severity in the log message though.

yalcinmelihyasin: I have decided to put it behind a config flag to not confuse our users.

gapis/api/vulkan/find_issues.go
```diff
@@ -290,16 +291,14 @@ func (t *findIssues) Flush(ctx context.Context, out transform.Writer) error {
 			if issue.Command == api.CmdNoID {
 				// The debug report is issued for state rebuilding command
```
hevrard: remove this

PR #75, opened 2020-02-19
Remove internet permission of the apk.

lpy:
Previously gathering gpu slices require internet permission because the data
producer created TCP/IP socket to communicate with the driver. Now that the
architecture has changed, the data producer no longer needs to create it and
hence we can remove internet permission.

Bug: b/144872032


lpy:
ptal

PR #76, opened 2020-02-20
Trace and replay Vulkan sample app using SwiftShader in CI

paulthomson:
This work was started by @hevrard

### DIFF
--- a/kokoro/linux-test/test.sh
+++ b/kokoro/linux-test/test.sh
@@ -41,7 +41,7 @@ function bazel {
     "$@"
 }
 
-function test {
+function bazel-test {
     echo $(date): Starting test for $@...
     $BUILD_ROOT/bazel/bin/bazel \
         --output_base="${TMP}/bazel_out" \
@@ -68,14 +68,26 @@ export VK_LOADER_DEBUG=all
 bazel build -c dbg //cmd/vulkan_sample:vulkan_sample
 
 # Run it.
+
 SAMPLE_PATH="$(bazel run -c dbg --run_under "echo" //cmd/vulkan_sample:vulkan_sample)"
+
+# Allow non-zero exit status.
+set +e
+
 timeout --preserve-status -s INT -k 10 5 xvfb-run -e xvfb.log -a "${SAMPLE_PATH}"
+APP_EXIT_STATUS="${?}"
+
+set -e
+
+# This line will exit with status 1 if the app's exit status
+# was anything other than 130 (128+SIGINT).
+test "${APP_EXIT_STATUS}" -eq 130
 
 # Running all the tests in one go leads to an out-of-memory error on Kokoro, hence the division in smaller test sets
-test tests-core
-test tests-gapis-api
-test tests-gapis-replay-resolve
-test tests-gapis-other
-test tests-gapir
-test tests-gapil
-test tests-general
+bazel-test tests-core
+bazel-test tests-gapis-api
+bazel-test tests-gapis-replay-resolve
+bazel-test tests-gapis-other
+bazel-test tests-gapir
+bazel-test tests-gapil
+bazel-test tests-general
--- a/kokoro/linux-test/test.sh
+++ b/kokoro/linux-test/test.sh
@@ -35,6 +35,14 @@ export CC=/usr/bin/gcc-8
 cd $SRC
 BUILD_SHA=${KOKORO_GITHUB_COMMIT:-$KOKORO_GITHUB_PULL_REQUEST_COMMIT}
 
+function bazel {
+  echo $(date): Starting bazel command $@...
+  $BUILD_ROOT/bazel/bin/bazel \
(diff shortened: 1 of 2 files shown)


kokoro/linux-test/test.sh
```diff
@@ -35,13 +35,7 @@ export CC=/usr/bin/gcc-8
...
 BUILD_SHA=${KOKORO_GITHUB_COMMIT:-$KOKORO_GITHUB_PULL_REQUEST_COMMIT}
 
-function bazel {
-  "${BUILD_ROOT}/bazel/bin/bazel" \
-    --output_base="${TMP}/bazel_out_dbg" \
-    "$@"
-}
-
-function bazel-test {
+function test {
```
chrisforbes: Can we not call this the same name as a shell builtin which you also use?

paulthomson: That is gone now

paulthomson: (I was just reverting the changes in test.sh.) Please view all "files changed", not per commit.

kokoro/linux/build.sh
```diff
@@ -81,6 +81,46 @@ do
...
+  cat xvfb.log
+fi
+
+# This line will exit with status 1 if the app's exit status
+# was anything other than 130 (128+SIGINT).
+test "${APP_EXIT_STATUS}" -eq 130
+
+xvfb-run -e xvfb.log -a bazel-bin/pkg/gapit trace -device host -disable-coherentmemorytracker -disable-pcs -disable-unknown-extensions -record-errors -no-buffer -api vulkan -start-at-frame 5 -capture-frames 10 -observe-frames 1 -out vulkan_sample.gfxtrace bazel-bin/cmd/vulkan_sample/vulkan_sample
+
+xvfb-run -e xvfb.log -a bazel-bin/pkg/gapit video -gapir-nofallback -type sxs -frames-minimum 3 -out vulkan_sample.mp4 vulkan_sample.gfxtrace
```
AWoloszyn: I dont think you need -type sxs

AWoloszyn: Also why do we have -frames-minumum? Don't expect exactly 10 frames ?

paulthomson: Do you mean it should be `-frames-minimum 10`? Sounds fair to me. I think 3 is just to safeguard against the case where no frames were captured. But you would hope `trace` would fail in this case. And indeed, it would be bad if only 3 frames were captured.

Without "-type sxs", I think the default will usually be sxs anyway, but it may switch to regular video if no framebuffer observations were found. Using "-type sxs" ensures we fail if there are no framebuffer observations (although we have a safeguard against this already!).

kokoro/linux/build.sh
```diff
@@ -81,6 +81,46 @@ do
...
+# Use SwiftShader.
+export VK_ICD_FILENAMES="$(pwd)/swiftshader/lib/vk_swiftshader_icd.json"
+export VK_LOADER_DEBUG=all
+
+# Just try running the app first.
+
+# Allow non-zero exit status.
+set +e
+
+xvfb-run -e xvfb.log -a timeout --preserve-status -s INT -k 10 5 bazel-bin/cmd/vulkan_sample/vulkan_sample
```
hevrard: `-k 10` should probably be `-k 1`, it is **extra** time (in seconds) before sending sigkill, i.e.:
`timeout -s INT -k 1 5 foo` runs `foo` for 5s, sends SIGINT, waits an **extra** 1s for `foo` to terminate, and sends SIGKILL otherwise.

From the man page:
```
       -k, --kill-after=DURATION
              also send a KILL signal if COMMAND is still running
              this long after the initial signal was sent
```

kokoro/linux/build.sh
```diff
@@ -81,6 +81,46 @@ do
...
 $SRC/kokoro/linux/package.sh $BUILD_ROOT/out
+
+##
+## Test capture and replay of the Vulkan Sample App.
+##
+
+# Install the Vulkan loader and xvfb (X virtual framebuffer).
+sudo apt-get -qy install libvulkan1 xvfb
+
+# Get prebuilt SwiftShader.
```
hevrard: Add a comment saying why we use this particular commit, or else to say that we just take the latest commit available and this should be updated from time to time.

kokoro/linux/build.sh
paulthomson: Thanks! I changed to 5, which is what I intended.

kokoro/linux/build.sh
paulthomson: Done thanks. I also added a comment explaining that we have to disable the coherent memory tracker with SwiftShader.

kokoro/linux/build.sh
paulthomson: I changed to `-frames-minimum 10`

PR #79, opened 2020-02-21
Fix several bugs in counter panel.

stellama0208:
- Identify a counter selection by counter id, rather than the combination of name + start time ts.
- Query a counter selection by finding a row with fitted time range, rather than with the exact time.
- These change is to combat time quantization happening in counter track.
 - Bug 1: http://b/147922469, #5.
 - Bug 2: http://b/149959272. #1.
 - Bug 3: http://b/149959272, #2.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
@@ -50,7 +50,7 @@ public class CounterTrack extends Track.WithQueryEngine<CounterTrack.Data> {
         "(select *, first_value(id) over (partition by quantum_ts order by dur desc) as best_id from %s) " +
       "group by quantum_ts";
   private static final String COUNTER_SQL = "select ts, ts + dur, value, id from %s";
-  private static final String VALUE_SQL = "select ts, ts + dur, value, id from %s where ts = %d";
+  private static final String VALUE_SQL = "select ts, ts + dur, value, id from %s where id = %d";
   private static final String RANGE_SQL =
       "select ts, ts + dur, value, id from %s " +
       "where ts + dur >= %d and ts <= %d order by ts";
@@ -123,8 +123,8 @@ public class CounterTrack extends Track.WithQueryEngine<CounterTrack.Data> {
     return format(COUNTER_SQL, tableName("span"));
   }
 
-  public ListenableFuture<Data> getValue(long t) {
-    return transform(expectOneRow(qe.query(valueSql(t))), row -> {
+  public ListenableFuture<Data> getValue(long id) {
+    return transform(expectOneRow(qe.query(valueSql(id))), row -> {
       Data data = new Data(null, new long[2], new double[2], new long[2]);
       data.ts[0] = row.getLong(0);
       data.ts[1] = row.getLong(1);
@@ -156,8 +156,8 @@ public class CounterTrack extends Track.WithQueryEngine<CounterTrack.Data> {
     });
   }
 
-  private String valueSql(long t) {
-    return format(VALUE_SQL, tableName("vals"), t);
+  private String valueSql(long id) {
+    return format(VALUE_SQL, tableName("vals"), id);
   }
 
   private String rangeSql(TimeSpan ts) {
--- a/gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
@@ -31,7 +31,6 @@ import com.google.gapid.perfetto.TimeSpan;
 import com.google.gapid.perfetto.views.CountersSelectionView;
 import com.google.gapid.perfetto.views.State;
 
-import java.util.stream.Collectors;
 import org.eclipse.swt.widgets.Composite;
 
 import java.util.Arrays;
@@ -40,7 +39,7 @@ import java.util.function.Consumer;
 
 public class CounterTrack extends Track.WithQueryEngine<CounterTrack.Data> {
   private static final String VIEW_SQL_DELTA =
-      "select ts + 1 ts, lead(ts) over win - ts dur, lead(value) over win value, id " +
+      "select ts + 1 ts, lead(ts) over win - ts dur, lead(value) over win value, lead(id) over win id " +
       "from counter where track_id = %d window win as (order by ts)";
   private static final String VIEW_SQL_EVENT =
       "select ts, lead(ts, 1, (select end_ts from trace_bounds)) over win - ts dur, value, id " +
@@ -206,7 +205,7 @@ public class CounterTrack extends Track.WithQueryEngine<CounterTrack.Data> {
 
     private void initKeys() {
       for (long[] keys : ids) {
-        valueKeys.addAll(Arrays.stream(keys).boxed().collect(Collectors.toSet()));
+        Arrays.stream(keys).forEach(valueKeys::add);
       }
     }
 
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
```diff
@@ -136,19 +145,21 @@ private String counterSQL() {
         return Data.empty(null);
       }
 
-      Data data = new Data(null, new long[rows + 1], new double[rows + 1]);
+      Data data = new Data(null, new long[rows + 1], new double[rows + 1], new long[rows + 1]);
       res.forEachRow((i, r) -> {
         data.ts[i] = r.getLong(0);
         data.values[i] = r.getDouble(2);
+        data.ids[i] = r.getLong(3);
```
silence-do-good: Can we remove the (1) element from the rangeSql if its not getting used anywhere?

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
```diff
@@ -39,17 +40,21 @@
...
+      "select min(ts), max(ts + dur), avg(value), best_id from " +
+        "(select *, first_value(id) over (partition by quantum_ts order by dur desc) as best_id from %s) " +
+      "group by quantum_ts";
+  private static final String COUNTER_SQL = "select ts, ts + dur, value, id from %s";
+  private static final String VALUE_SQL =
+      "select ts, ts + dur, value, id from %s " +
+      "where ts <= %d and ts + dur > %d";
   private static final String RANGE_SQL =
-      "select ts, ts + dur, value from %s " +
+      "select ts, ts + dur, value, id from %s " +
```
silence-do-good: Is the ts + dur under select getting used? If not, we can remove that from the query.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
```diff
@@ -99,13 +104,15 @@ private String viewSql() {
...
-      Data data = new Data(req, new long[rows + 1], new double[rows + 1]);
+      Data data = new Data(req, new long[rows + 1], new double[rows + 1], new long[rows + 1]);
       res.forEachRow((i, r) -> {
         data.ts[i] = r.getLong(0);
         data.values[i] = r.getDouble(2);
+        data.ids[i] = r.getLong(3);
       });
       data.ts[rows] = res.getLong(rows - 1, 1, 0);
       data.values[rows] = data.values[rows - 1];
+      data.ids[rows] = data.ids[rows - 1];
```
silence-do-good: Is it possible to add a comment to why we are adding an extra index everywhere? Would help if any one of the functions had a comment.  Ignore this if its too trivial.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
```diff
@@ -295,7 +313,7 @@ public Values combine(Values other) {
     }
 
     @Override
-    public Selection<Values.Key> build() {
+    public Selection<Long> build() {
```
silence-do-good: Delete the Key class below this function if its not getting used anywhere.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
stellama0208: It's actually used at `data.ts[rows] = res.getLong(rows - 1, 1, 0);`.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
stellama0208: It's actually used at `data.ts[rows] = res.getLong(rows - 1, 1, 0);`.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
stellama0208: Thanks for reminding, forgot this bit! Done in the new change.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
stellama0208: It's actually for transforming (ts, ts + dur, value) scheme to (ts, value) scheme. The added row is for the (very last timestamp - very last value) pair. 

Since this comes from the design of CounterTrack at the very beginning and done by Pascal, I might not be the best person to explain it. Will omit the comment for this PR, maybe we can re-check this thing when Pascal is back.

stellama0208:
Addressed the comments above. PTAL!

Another change in the updated PR is that now I query counter selection based on id rather than time range, this will further solve another bug that you couldn't select counter when zoomed too **out** in a long trace.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
```diff
@@ -158,44 +167,46 @@ private String rangeSql(TimeSpan ts) {
...
     }
 
     private void initKeys() {
-      for (String name : names) {
-        // Skip the last dummy entry for the range end.
-        Arrays.stream(ts, 0, ts.length - 1)
-            .boxed()
-            .forEach(t -> valueKeys.add(new Values.Key(name, t)));
+      for (long[] keys : ids) {
+        valueKeys.addAll(Arrays.stream(keys).boxed().collect(Collectors.toSet()));
```
silence-do-good: Nit: Didn't notice this before. The old code seems to ignore the last dummy entry. I. don't think the new one does that. However, I'm not too sure of whether it will affect the end result or not.

silence-do-good: Nit: Didn't notice this before. The old code seems to ignore the last dummy entry. I. don't think the new one does that. However, I'm not too sure of whether it will affect the end result or not.

rschiu: I think java.util.Collections.addAll would work and that would be best if it does.
If it doesn't, then you actually want to use forEach like before.  Using collect will unnecessary create a Set to contain the value of the stream.
Also, I don't think you need to boxed it if you are using forEach, but not 100% sure.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
```diff
@@ -39,17 +40,19 @@
 
 public class CounterTrack extends Track.WithQueryEngine<CounterTrack.Data> {
   private static final String VIEW_SQL_DELTA =
-      "select ts + 1 ts, lead(ts) over win - ts dur, lead(value) over win value " +
+      "select ts + 1 ts, lead(ts) over win - ts dur, lead(value) over win value, id " +
```
rschiu: This should be 'lead(id) over win id' instead of id.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
rschiu: It's probably better this way in the new code.  The old code use name + ts as the key. The last value is artificially created and doesn't actually exist in the database.  In the new code, the key is the id.  The last value is just a duplicate of the previous value.  Since it's being inserted into a Set, the duplicate just get ignored.

stellama0208: Yes, it's taking advantage of set+long to keep code shorter.

stellama0208: Rewrote it to avoid set creation.

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
stellama0208: Nice catch, thanks! Done.

PR #80, opened 2020-02-21
Don't automatically select default counters at empty selection.

stellama0208:
- Bug: http://b/150017682.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
@@ -810,8 +810,7 @@ public class TraceConfigDialog extends DialogBase {
         table.setInput(caps.getGpuProfiling().getGpuCounterDescriptor().getSpecsList());
         table.setCheckedElements(
             caps.getGpuProfiling().getGpuCounterDescriptor().getSpecsList().stream()
-                .filter(currentIds.isEmpty() ?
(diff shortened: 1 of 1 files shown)


stellama0208:
> Ideally, there would be a way to differentiate between none vs default vs manually selecting all default counters. Then it would make it possible to change device and have the correct default counters selected. But that's another bug and has low priority. Can you file a bug for that?

A bug is filed for it: https://b.corp.google.com/issues/150038262.

lpy:
> Ideally, there would be a way to differentiate between none vs default vs manually selecting all default counters. Then it would make it possible to change device and have the correct default counters selected. But that's another bug and has low priority.

we already have a `default` button, we should leave it blank and let users select what they want. And do our best to remember the selection. The real bug I have encountered is when I unplug the only device and plug another device with a different GPU, the number of selected counters shown doesn't change but actually there are no counters selected.

rschiu:
> we already have a default button, we should leave it blank and let users select what they want. 

Since we named it default, it just seems more intuitive those counters will be selected on initialization instead of selecting nothing at all.  Although I can be convinced otherwise.

> The real bug I have encountered is when I unplug the only device and plug another device with a different GPU, the number of selected counters shown doesn't change but actually there are no counters selected.

The purpose of differentiating between default vs manually selecting the same set of counters is to solve this bug for the common use case.  The counters are different between devices, so there is no good way to translate the counter selections.  The trivial solution is to reset the counter selection to empty every time we change devices, but it will be a better UX if we select the default counters when the user indicated they just want default counters.

PR #81, opened 2020-02-22
Use Perfetto Client API in tracing.

lpy:
Previously we used Perfetto CLI to trace, however, when tracing is started,
some ftrace are not set up yet, which results in a gap of empty data at the
beginning of the trace. With Perfetto Client API, we now can do deferred start
by setting deferred_start field to true, and always make sure we only start the
tracing session when things are ready.

Bug: b/147490759

### DIFF
--- a/gapis/perfetto/android/trace.go
+++ b/gapis/perfetto/android/trace.go
@@ -15,15 +15,19 @@
 package android
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
 	"os"
 	"sync/atomic"
+	"time"
 
 	perfetto_pb "protos/perfetto/config"
 
+	"github.com/golang/protobuf/proto"
 	"github.com/google/gapid/core/app"
+	"github.com/google/gapid/core/app/crash"
 	"github.com/google/gapid/core/event/task"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/os/android"
@@ -32,11 +36,13 @@ import (
 	"github.com/google/gapid/core/os/file"
 	"github.com/google/gapid/core/text"
 	"github.com/google/gapid/gapidapk"
+	"github.com/google/gapid/gapis/perfetto"
 	"github.com/google/gapid/gapis/service"
 )
 
 const (
-	gpuRenderStagesDataSourceDescriptorName = "gpu.renderstages"
+	ftraceDataSourceName          = "linux.ftrace"
+	gpuRenderStagesDataSourceName = "gpu.renderstages"
 
 	// perfettoTraceFile is the location on the device where we'll ask Perfetto
 	// to store the trace data while tracing.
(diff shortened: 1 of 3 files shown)


lpy:
I still need to figure out what to check to know things are ready, it seems to me `/sys/kernel/debug/tracing/tracing_on` doesn't work at all.

lpy:
ptal

gapis/perfetto/android/trace.go
```diff
@@ -208,3 +226,63 @@ func (p *Process) Capture(ctx context.Context, start task.Signal, stop task.Sign
...
+	// TODO(b/150141871) Figure out a robust way to determine whether tracing is
+	// ready. Setting up ftrace trace points takes time, hence sometimes there's
+	// a gap at the beginning of the trace. In order to avoid this issue, we
+	// need to know when ftrace trace points are ready. This workaround checks
+	// the sysfs node of the ftrace event task_newtask that is always enabled
+	// regardless of the user selection. Skip if ftrace is not enabled.
+	if hasDataSourceEnabled(p.config, ftraceDataSourceName) {
+		sessionReadySignal, sessionReadyFunc := task.NewSignal()
+		crash.Go(func() {
+			for {
```
rschiu: We should have a timemout.

lpy:
ptal

lpy:
Updated the patch based on our offline discussion:

1. If the trace config has write into file, use CLI
2. If the trace is too long (> 10 seconds), use CLI
3. Otherwise, try to use consumer protocol, if it fails to connect, use CLI

I still keep the sysfsnode check given that the gap will increase bigger and bigger.

PTAL!

gapis/perfetto/android/trace.go
```diff
@@ -167,15 +175,33 @@ func Start(ctx context.Context, d adb.Device, a *android.ActivityAction, opts *s
 		}
 	}
 
+	// With the comsumer protocol, in Android 10 it causes a stall when
+	// traced_pros writes into file or the trace is too long. To be more
+	// conservative we consider traces longer than 10 seconds long traces.
+	// In this case, don't create perfetto client, instead fall back to CLI.
+	var c *perfetto.Client
+	if !opts.PerfettoConfig.GetWriteIntoFile() && opts.PerfettoConfig.GetDurationMs() <= maxInMemDuration {
```
pmuetschard: We don't need the duration check. If write into file is not set there is no benefit of using the command line even for long traces. The command line does exactly the same as we do and downloads the data at the end if write-into-file is disabled.

gapis/perfetto/android/trace.go
```diff
@@ -208,3 +234,76 @@ func (p *Process) Capture(ctx context.Context, start task.Signal, stop task.Sign
...
 }
+
+func (p *Process) captureWithClientApi(ctx context.Context, start task.Signal, stop task.Signal, ready task.Task, w io.Writer, written *int64) (int64, error) {
+	defer p.perfettoClient.Close(ctx)
+
+	p.config.DeferredStart = proto.Bool(true)
+	var buf bytes.Buffer
+	ts, err := p.perfettoClient.Trace(ctx, p.config, &buf)
+	if err != nil {
+		return 0, log.Err(ctx, err, "Failed to call EnableTrace")
```
pmuetschard: The message shouldn't mention `EnableTrace`, that is a client internal and is likely already part of the error. Simply "Failed to setup Perfetto trace" or something like that, is better.

lpy:
ping for review

pmuetschard:
> ping for review

Please address my comments

lpy:
ops, sorry about that, forgot to push the patch.

PTAL!

gapis/perfetto/android/trace.go
```diff
@@ -167,15 +174,33 @@ func Start(ctx context.Context, d adb.Device, a *android.ActivityAction, opts *s
 		}
 	}
 
+	// With the comsumer protocol, in Android 10 it causes a stall when
+	// traced_pros writes into file or the trace is too long. To be more
+	// conservative we consider traces longer than 10 seconds long traces.
```
pmuetschard: the comment is now out-of-date.

lpy:
I am going to go ahead and commit the patch given this is a flaky test.

PR #84, opened 2020-02-26
DO NOT MERGE Generate links to renderpasses from gpu slices

apbodnar:
@AWoloszyn Opening as a PR to solicit discussion.  This does what is probably an unnecessary amount of state tracking inside of a transform.  Since we're already generating links to the command buffers in a queue submit, i'm hoping it might be possible to walk the nodes in the existing command tree to find the render passes rather than what's happening here.

### DIFF
--- a/gapis/api/cmd_id.go
+++ b/gapis/api/cmd_id.go
@@ -53,5 +53,7 @@ func (id CmdID) String() string {
 // TODO(apbodnar) find a more appropriate place for this
 type CommandSubmissionKey struct {
 	SubmissionOrder uint64
-	CommandBuffer   int64
+	CommandBuffer   uint64
+	RenderPass      uint64
+	Framebuffer     uint64
 }
--- a/gapis/api/vulkan/slice_command_mapper.go
+++ b/gapis/api/vulkan/slice_command_mapper.go
@@ -22,16 +22,36 @@ import (
 	"github.com/google/gapid/gapis/api/transform"
 )
 
+type beginRenderPassIndex struct {
+	index       uint64
+	renderPass  uint64
+	framebuffer uint64
+}
+
 type sliceCommandMapper struct {
-	subCommandIndicesMap *map[api.CommandSubmissionKey][]uint64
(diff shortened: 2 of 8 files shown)


gapis/api/cmd_id.go
```diff
@@ -53,5 +53,7 @@ func (id CmdID) String() string {
 // TODO(apbodnar) find a more appropriate place for this
 type CommandSubmissionKey struct {
 	SubmissionOrder uint64
-	CommandBuffer   int64
+	CommandBuffer   uint64
+	RenderPass      uint64
+	Framebuffer     uint64
```
AWoloszyn: This is probably the wrong place for this (For all of them in practice).
It leaks Vulkan-isms out of the Vulkan package.

Renderpasses/submission indices / renderpasses/etc are all handled currently in vulkan.go when generating the sync-data. I think we can probably get this to work well with that data, rather than re-implementing it outside

gapis/api/vulkan/slice_command_mapper.go
```diff
@@ -22,16 +22,36 @@ import (
...
+	framebuffer uint64
+}
+
 type sliceCommandMapper struct {
-	subCommandIndicesMap *map[api.CommandSubmissionKey][]uint64
+	subCommandIndicesMap *map[api.CommandSubmissionKey][][]uint64
 	submissionCount      uint64
+	commandBuffersMap    map[uint64]uint64
+	beginRenderPassMap   map[uint64][]beginRenderPassIndex
 }
```
AWoloszyn: I may have missed this on my first view of this CL.

This transform doesn't actually transform anything does it? I can't see where it is actually modifying the commands.
If so, this should probably just be a loop over the commands outside of the transform interface. (That way it wont cost every time we run it, and we can cache the results)

gapis/api/cmd_id.go
apbodnar: If you can suggest an alternative place for this type, I'd appreciate it.  Given the scope of profiling_data.go, importing the vulkan package causes a circular dependency.

PR #85, opened 2020-02-26
Enum tooltip takes precedence over activation tooltip

hliatis:
Bug: b/150000414


gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -475,7 +475,7 @@ public void update(ViewerCell cell) {
               public String getToolTipText(Object element) {
                 DataValue dv = convertDataValue(((API.Row)element).getRowValues(col));
                 if (dv != null) {
-                  if (!dataTable.getDependee().equals("")) {
+                  if (dv.tooltipValue == null && !dataTable.getDependee().equals("")) {
                     if (dataTable.getActive()) {
                         return (dv.tooltipValue != null ? dv.tooltipValue : "Activated by " + dataTable.getDependee());
```
johnplate: In this line dv.tooltipValue is now always null, so the ternary operation is redundant

PR #88, opened 2020-02-26
Do not store junk values into the state for imageview part of sampler descriptors

chrisforbes:


yalcinmelihyasin:
In general, it seems pretty good.

Are we sure we do not leak any garbage information in other descriptor set types?

AWoloszyn:
> In general, it seems pretty good.
> 
> Are we sure we do not leak any garbage information in other descriptor set types?

I think we can run into the same problems with the reverse. (ImageViews with garbage samplers for example)

yalcinmelihyasin:
> > In general, it seems pretty good.
> > Are we sure we do not leak any garbage information in other descriptor set types?
> 
> I think we can run into the same problems with the reverse. (ImageViews with garbage samplers for example)

Then should we make this pr a bigger one with all the fixes? Or should we followed it with other prs? In that case, we should create an issue for it.

chrisforbes:
Yes, there are various other case like this.

chrisforbes:
I'll keep fixing those as I find them, in follow-ups.

PR #89, opened 2020-02-26
Add option to right truncate the track name

silence-do-good:
By default, we left truncate the track name. This makes sense for processes where the truncated part has only the system path. But for GPU counters, left truncate cuts out important identifying information. This PR adds a right truncate option and sets the GPU counters to use right truncate instead.

Bug : 149936845


gapic/src/main/com/google/gapid/perfetto/canvas/RenderContext.java
```diff
@@ -272,6 +272,36 @@ private void drawTextLeftTruncate(
     }
   }
 
+  // draws text centered vertically, right truncated to fit into the given width.
```
rschiu: Move new code below drawTextLeftTruncate so that all the left truncate code are together.

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
```diff
@@ -52,20 +52,40 @@
...
+    };
+
+    public abstract void drawTextTruncate(RenderContext ctx, Fonts.Style style, String text,
+        double x, double y, double w, double h);
+  }
+
   public static <T extends TrackPanel<T>> TrackConfig.Track.UiFactory<Panel> single(
-      TrackConfig.Track.UiFactory<T> track, boolean sep) {
-    return state -> new Single<T>(state, track.createPanel(state), sep, null, true);
+      TrackConfig.Track.UiFactory<T> track, boolean sep, Truncate truncate) {
```
rschiu: Would be simpler if this is just a boolean.  You can also overload the 'single' function to default to truncate left.

silence-do-good: Yes but it will look very hacky. This would be a cleaner approach and we can add more if we want.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -173,7 +174,7 @@ private Tracks() {
       for (CounterInfo counter : counters) {
         CounterTrack track = new CounterTrack(data.qe, counter);
         data.tracks.addTrack(parent, track.getId(), counter.name,
-            single(state -> new CounterPanel(state, track, DEFAULT_COUNTER_TRACK_HEIGHT), true));
+            single(state -> new CounterPanel(state, track, DEFAULT_COUNTER_TRACK_HEIGHT), true, Truncate.right));
```
lpy: indent this properly

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -232,7 +233,7 @@ private Tracks() {
         for (CounterInfo counter : counters) {
           CounterTrack track = new CounterTrack(data.qe, counter);
           data.tracks.addTrack(parentId, track.getId(), counter.name,
-              single(state -> new CounterPanel(state, track, PROCESS_COUNTER_TRACK_HIGHT), false));
+              single(state -> new CounterPanel(state, track, PROCESS_COUNTER_TRACK_HIGHT), false, Truncate.left));
```
lpy: ditto

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -251,11 +252,11 @@ private Tracks() {
         boolean isIdleThread = hasIdleThreads && track.getThread().totalDur < idleCutoffThread;
         TrackConfig.Track.UiFactory<Panel> ui;
         if (track.getThread().maxDepth == 0) {
-          ui = single(state -> new ThreadPanel(state, track, false), false);
+          ui = single(state -> new ThreadPanel(state, track, false), false, Truncate.left);
         } else {
           boolean expanded = !isIdleProcess && !isIdleThread;
           ui = single(state ->
-            new ThreadPanel(state, track, expanded), false, ThreadPanel::setCollapsed, !expanded);
+            new ThreadPanel(state, track, expanded), false, ThreadPanel::setCollapsed, !expanded, Truncate.left);
```
lpy: ditto

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
lpy: I think this is more readable

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
```diff
@@ -52,20 +52,40 @@
   private TrackContainer() {
   }
 
+  public static enum Truncate {
+    left {
```
lpy: Left

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
```diff
@@ -52,20 +52,40 @@
...
+      @Override
+      public void drawTextTruncate(RenderContext ctx, Fonts.Style style, String text,
+          double x, double y, double w, double h) {
+        ctx.drawTextLeftTruncate(Fonts.Style.Normal, text, x, y, w, h);
+      }
+    },
+    right {
```
lpy: Right

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
```diff
@@ -52,20 +52,40 @@
...
+      @Override
+      public void drawTextTruncate(RenderContext ctx, Fonts.Style style, String text,
+          double x, double y, double w, double h) {
+        ctx.drawTextRightTruncate(Fonts.Style.Normal, text, x, y, w, h);
+      }
+    };
+
+    public abstract void drawTextTruncate(RenderContext ctx, Fonts.Style style, String text,
```
lpy: Just `drawText`

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
silence-do-good: I had it as Left, but the eclipse language suggestion warned that usually enum constants should start with lowercase. So I reverted back to all lowercase. I'll change it to Left if needed. What do you think?

lpy: I did a brief check in our codebase, I think `Left` is better

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
mattyclarkson: Old but gold: https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

rschiu: Yes, unfortunately, Java does not support named argument.  But that is an easily solvable problem.  We can just add a comment so that it's look like `single(..., true /*truncateLeft*/)`.  Or pass a constant enum value so it looks like `single(..., Truncate.LEFT)`.  Even the page you link suggest different ways to named the argument instead of turning the boolean argument into a functor.

The PR unnecessarily complicate the code for flexibility that is not currently needed.  You have to go through multiple object inheritance to determine what truncate.drawText actually do.  I mean...if you think about it, it's not even named well.  The functor class is named Truncate, but the actual main purpose is drawing text.

Not worth arguing over it if that doesn't convince you, but I do think it's make the code more complicated than it needs to be.

lpy: maybe just change the name to `DrawTextTruncate`

silence-do-good: I had it first as a simple enum. My idea to move this to a enum method was because the caller is actually a long line of code. It will look a little ugly if we have multiple lines looking exactly same except for the change of the function name. Another argument was that, if we want to do a different justification, we can just add a new LeftJustified or RightJustified in the enum and call whatever function needs to be called there. I can change it to just a simple enum with if else statement if its needed. I believe this just looks cleaner.

stellama0208: Would also prefer it to go with a simple boolean value + if else statement at rendering. 
1.The code style of `single()` track creation method. The other lightweight variable for initialization are in boolean, e.g. filtered or not, expanded or not, with separator or not.
2.The code style of AGI, or JAVA. So far the enum definition only happens when there's more than two possibilities for a state. Since it's either left or right, the creation of a new class carrying function inside is over-killing.

rschiu: > Another argument was that, if we want to do a different justification, we can just add a new LeftJustified or RightJustified in the enum and call whatever function needs to be called there.

That's what I mean by complicating code for flexibility that is not currently needed.  And frankly, using a simple enum does the same thing.  If you use simple enum, adding new draw function is just as easy.  You would just add a new case in a switch statement.

The typical reason to use a functor is to allow the caller to implement it.  But this is an enum, so caller cannot implement their own draw function.  Nor do we actually to give this flexibility to the caller.

> I'll change it to just a simple enum with if else statement if you guys think so. I believe this just looks cleaner.

I have to disagree, but not strongly enough that I will block it if you insist.

silence-do-good: I've removed the enum method and made it simple if-else inside RenderContext now. But kept the enum for defining left or right.

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
```diff
@@ -52,20 +52,40 @@
   private TrackContainer() {
   }
 
+  public static enum Truncate {
+    Left {
```
rschiu: I thought about more, this really should be LEFT and RIGHT.  This is an enum, so it should be named as such.

As a side note, the fact that we have to discuss how to name this partly shows why I think this way of doing this is more confusing than it needs to be.

rschiu: https://engdoc.corp.google.com/eng/doc/devguide/java/style/index.md?cl=head#s4.8.1-enum-classes

silence-do-good: I gave it some more thought and decided to do away with enums, since importing stuff from RenderContext into models is not something we have done anywhere else.

silence-do-good:
Got rid of the enum after our long conversation here. Didn't want to break any existing structure we have in AGI. If we plan on adding more entries to truncation, we can rework this to be enums. For now, changing it to boolean to be simple. 
PTAL @lpy @rschiu @stellama0208

gapic/src/main/com/google/gapid/perfetto/canvas/RenderContext.java
```diff
@@ -284,6 +284,45 @@ public void drawTextLeftTruncate(
...
+      if (l <= 0) {
+        break;
+      }
+      toDisplay = text.substring(0, l) + "...";
+    }
+  }
+
+  public void drawTextTruncated(
+      Fonts.Style style, String text, double x, double y, double w, double h, Truncate truncate) {
+    if (truncate == Truncate.Left) {
```
stellama0208: I still think it should be a boolean than a two possibility enum. 
Doing a search about `public static enum` in AGI we can see that enum are mostly defined when there's a lot of options or when the items are highly descriptive.
We do `if(truncate == Truncate.Left) else` rather than `if(truncate == Truncate.Left) else if (truncate == Truncate.Right)` here further makes me feel boolean is enough for this task. 
I see the benefit of easy readibility at the panel creation time, but that part of code is not modified that often and it's easy to go one step further to see the meaning of the boolean value, so I still think boolean is better in this case.

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
```diff
@@ -101,31 +101,33 @@ private TrackContainer() {
...
     private final PinState pinState;
+    private final boolean truncate; // False -> Left truncate, True -> Right truncate
 
     protected boolean filtered;
     protected boolean hovered = false;
 
     public Single(State.ForSystemTrace state, T track, boolean sep, BiConsumer<T, Boolean> filter,
-        boolean filtered) {
-      this(track ,sep, filter, filtered, new PinState(state));
+        boolean filtered, boolean truncate) {
```
lpy: should be rightTruncate? Same applies to others

PR #91, opened 2020-02-26
Fix wrong counter selection count in config window.

stellama0208:
- Bug: http://b/150290967.


silence-do-good:
Are the counter ids interchangeable across vendors though?

lpy:
> Are the counter ids interchangeable across vendors though?

Yes, it is.

PR #92, opened 2020-02-27
Have ImageView links in PV go to Texture View

hliatis:
Todo: Have it go to both Texture and State View

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Follower.java
+++ b/gapic/src/main/com/google/gapid/models/Follower.java
@@ -231,6 +231,9 @@ public class Follower {
       case STATE:
         listeners.fire().onStateFollowed(path);
         break;
+      case IMAGE_VIEW:
+        listeners.fire().onTextureFollowed(path.getImageView());
+        break;
       default:
         LOG.log(WARNING, "Unknown follow path result: " + path);
     }
@@ -259,6 +262,8 @@ public class Follower {
      * Event indicating that a link with the given path to a memory region was followed.
      */
     public default void onMemoryFollowed(Path.Memory path)  { /* empty */ }
+
+    public default void onTextureFollowed(Path.ImageView path) {}
   }
 
   public static interface Prefetcher<K> {
--- a/gapic/src/main/com/google/gapid/models/Follower.java
+++ b/gapic/src/main/com/google/gapid/models/Follower.java
@@ -25,6 +25,7 @@ import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.gapid.proto.service.api.API;
 import com.google.gapid.proto.service.path.Path;
+import com.google.gapid.proto.service.Service;
 import com.google.gapid.rpc.Rpc;
 import com.google.gapid.rpc.RpcException;
 import com.google.gapid.rpc.UiCallback;
@@ -61,11 +62,13 @@ public class Follower {
 
   private final Shell shell;
   private final Client client;
+  private final Resources resources;
   private final ListenerCollection<Listener> listeners = Events.listeners(Listener.class);
 
-  public Follower(Shell shell, Client client) {
+  public Follower(Shell shell, Client client, Resources resources) {
     this.shell = shell;
     this.client = client;
+    this.resources = resources;
   }
 
   /**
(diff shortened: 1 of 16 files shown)


gapis/api/vulkan/resources.go
```diff
@@ -1115,8 +1117,12 @@ func commonShaderDataGroups(ctx context.Context,
 							currentSetData = append(currentSetData, api.CreateLinkedDataValue("url", samplerPath, api.CreatePoDDataValue("VkSampler", samplerHandle)))
 
 							viewHandle := descInfo.ImageView()
-							viewPath := path.NewField("ImageViews", resolve.APIStateAfter(path.FindCommand(cmd), ID)).MapIndex(viewHandle)
-							currentSetData = append(currentSetData, api.CreateLinkedDataValue("url", viewPath, api.CreatePoDDataValue("VkImageView", viewHandle)))
+							//viewPath := path.NewField("ImageViews", resolve.APIStateAfter(path.FindCommand(cmd), ID)).MapIndex(viewHandle)
+
+							imageView, _ := vkState.ImageViews().Lookup(viewHandle)
+							imageHandle := imageView.Image().VulkanHandle()
+							imageViewPath := path.NewImageView(fmt.Sprintf("Image<%d>", imageHandle))
```
hliatis: This is the hacky part I want to discuss

gapis/service/path/path.go
```diff
@@ -119,6 +120,7 @@ func (n FramebufferObservation) Parent() Node    { return n.Command }
 func (n Field) Parent() Node                     { return oneOfNode(n.Struct) }
 func (n GlobalState) Parent() Node               { return n.After }
 func (n ImageInfo) Parent() Node                 { return nil }
+func (n ImageView) Parent() Node                 { return nil }
```
AWoloszyn: I am not sure making ImageViews their own type in the API is the right way to go.

gapis/api/vulkan/resources.go
AWoloszyn: We used to return a state path. (CmdID->ImageViews->Handle). 
I think we should create something similar to "Follow" (FollowResource) which takes that existing pah (CmdID->ImageViews->Handle) and returns a new Resource Path. This resource could contain the type (Texture in this case) and the information the UI would need for it. (Handle I believe) From that ResourcePath, we could determine which windows need to be in front (texture view in this case) and go to the correct object.

This lets us do this from anywhere in the UI, instead of only being able to do it from the TextureView. (For example, if someone clicked on an ImageView in the state-view we could also do this, or a pipeline->pipeline View, or buffer -> Model view etc etc).

gapis/service/path/path.go
hliatis: What do you recommend? In this approach, I need to send something new to the UI

hliatis:
The latest commit simply is proof of concept of a link going to multiple locations (e.g. ImageView links going to both Texture View and State View)

gapis/service/path/path.go
pmuetschard: I agree. I think it would make sense to create a path called `Resource`:
```
message Resource {
  Capture capture = 1;
  ID id =2;
}
```
We can then use it for and link to any type of resource.

hliatis:
Last push was to simply keep it up to date

hliatis:
Latest commit simply updates the current solution to use ResourceMeta. I'll begin updating to match what Andrew wanted.

hliatis:
Added a second commit that shows a solution that uses Path.ResourceData. Since ResourceType is in the API, not Path, this solution does a grab for a Service.Resource to determine the type.

gapic/src/main/com/google/gapid/models/Follower.java
```diff
@@ -231,6 +234,21 @@ public void onFollow(Path.Any path) {
...
+      case RESOURCE_DATA:
+        onFollowResource(path.getResourceData());
+        break;
+      default:
+        LOG.log(WARNING, "Unknown follow path result: " + path);
+    }
+  }
+
+  private void onFollowResource(Path.ResourceData path) {
+    Service.Resource r = resources.getResource(path).resource;
```
pmuetschard: `resources.getResource(path)` can return `null`. We should check it here. I think for now it's OK to simply log it.

gapis/service/path/path.proto
```diff
@@ -44,27 +44,28 @@ message Any {
...
-    Result result = 33;
-    Slice slice = 34;
-    State state = 35;
-    StateTree state_tree = 36;
-    StateTreeNode state_tree_node = 37;
-    StateTreeNodeForPath state_tree_node_for_path = 38;
-    Stats stats = 39;
-    Thumbnail thumbnail = 40;
-    Type type = 41;
+    ImageView image_view = 21;
```
pmuetschard: revert this file.

gapic/src/main/com/google/gapid/models/Follower.java
pmuetschard: This is still de-referencing a null. It needs to be:
```
Resources.Resource res = resources.getResource(path);
if (res != null) {
  switch (res.resource.getType()) {
    ...
  }
}
```

PR #93, opened 2020-02-27
Add selection behavior for Memory Panel.

stellama0208:
- Bug: http://b/147922469. #1.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/BatterySummaryTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/BatterySummaryTrack.java
@@ -20,6 +20,7 @@ import static com.google.gapid.perfetto.models.QueryEngine.createView;
 import static com.google.gapid.perfetto.models.QueryEngine.createWindow;
 import static com.google.gapid.perfetto.models.QueryEngine.dropTable;
 import static com.google.gapid.perfetto.models.QueryEngine.dropView;
+import static com.google.gapid.perfetto.models.QueryEngine.expectOneRow;
 import static com.google.gapid.perfetto.views.TrackContainer.single;
 import static com.google.gapid.util.MoreFutures.transform;
 import static com.google.gapid.util.MoreFutures.transformAsync;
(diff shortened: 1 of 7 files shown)


stellama0208:
This PR is ready for review. 
It will only cover selection behavior for memory panel and battery panel, because they are handled in a similar fashion. The selection behavior for other panels mentioned in the linked bug will be submitted in future PRs.

gapic/src/main/com/google/gapid/perfetto/models/BatterySummaryTrack.java
```diff
@@ -139,26 +159,187 @@ private String counterSQL() {
...
+        if (at == bt) {
+          ai++;
+          bi++;
+        } else if (at < bt) {
+          ai++;
+        } else {
+          bi++;
+          selectThis = false;
+        }
+        if (selectThis) {
```
silence-do-good: Is there a good reason behind choosing arrays for the members of Values? I've seen this in all the tracks. The logic here is the same as merge sort merging phase, which I feel can be much simplified if we use containers instead of arrays. I'm not being very picky here, but just curious.

gapic/src/main/com/google/gapid/perfetto/models/BatterySummaryTrack.java
```diff
@@ -139,26 +159,187 @@ private String counterSQL() {
     return data;
   }
 
+  public ListenableFuture<Values> getValue(long id) {
```
rschiu: Maybe I missed something, couldn't this be just a List<Value>?  Similar to getSices in your other PR.

gapic/src/main/com/google/gapid/perfetto/views/BatterySummaryPanel.java
```diff
@@ -168,9 +191,24 @@ public Area getRedraw() {
...
+          state.setSelection(Kind.Battery, track.getValue(id));
+        }
+        return true;
+      }
     };
   }
 
+  @Override
+  public void computeSelection(CombiningBuilder builder, Area area, TimeSpan ts) {
+    builder.add(Selection.Kind.Battery, transform(track.getValues(ts), v -> v));
```
rschiu: What does transform do when the input and output is the same?

gapic/src/main/com/google/gapid/perfetto/models/BatterySummaryTrack.java
stellama0208: So far the Selection objects in Counter track, Memory track, Battery track are a little different from others, you can see that they don't differentiate single selection and multi selection: 
1. There's only one `Values` class, while in other panels there's `Slice` class and `Slices` class.
2. The information window beneath is the same for single selection or multi selection.
Since there's no single selection then there wouldn't be a List<SingleSelectionClass>.

Note: This pattern comes from the original design in Counter track. I guess it's done because counter data is much simpler than the slices data. I'm a supporter to apply this pattern to other tracks in the future, seeing single selection just a special case of mult selection.

gapic/src/main/com/google/gapid/perfetto/models/BatterySummaryTrack.java
stellama0208: I think what you says makes great sense! I wrote it basically following the existing patterns in AGI, especially the counter track. 
After a closer look I believe Pascal's original choice for array was to lend convenience to CounterSelectionView. With array, each array can directly act as a real column in the selection info window, with an elegant implementation using `createTableColumn` and `setInput` . A tradeoff for simplicity here and the simplicity there have to be made...

gapic/src/main/com/google/gapid/perfetto/views/BatterySummaryPanel.java
stellama0208: It's a little hacky way to solve a java generic type issue here I have to admit... Replacing it with only `track.getValues(ts)` will somehow trigger a type unmatch error, because it couldn't sense the class inside the ListenableFuture which indeed implements the correct interface.

gapic/src/main/com/google/gapid/perfetto/models/BatterySummaryTrack.java
rschiu: +1, we really should't be re-inventing the wheels like this.

stellama0208:
According to our discussion in this PR, a buganizer ticket is created to track the selection implementation clean up: http://b/150716421.

PR #95, opened 2020-02-27
Disable Frame lifecycle by default

silence-do-good:
During the feedback from Blizzard, they were using android R and were confused due to the amount of tracks with Frame Lifecycle. Disabling it by default would be a better solution in the current state of frame lifecycle.
We can enable it if we want, to test it here.

Bug : 150403599

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
@@ -512,7 +512,7 @@ public class TraceConfigDialog extends DialogBase {
 
         if (gpuCaps.getHasFrameLifecycle()) {
           gpuFrame = createCheckbox(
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -512,7 +512,7 @@ public BasicInputArea(Composite parent, Settings settings, Theme theme,
 
         if (gpuCaps.getHasFrameLifecycle()) {
           gpuFrame = createCheckbox(
-              gpuGroup, "Frame Lifecycle", sGpu.getSurfaceFlinger(), e -> updateGpu());
+              gpuGroup, "Frame Lifecycle", false, e -> updateGpu());
```
rschiu: Default settings should be set in gapic/src/main/com/google/gapid/models/Settings.java.

PR #96, opened 2020-02-28
Add selection behavior for CPU Panel under time quantization.

stellama0208:
- In CPU Panel, when zoom in, the selection behavior for "CPU slices"
 already exists; but when zoom out, selection behavior for "CPU graph" is
 missing, this PR address this issue.
 - A very useful use case enabled by this PR would be 1.zooming out and
 have an overview, 2.select a peak point denoting high CPU utilization
 in graph, 3.click 'f'or 'm' to quickly zoom/mark.
 - Bug: http://b/147922469. #1.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/CpuTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/CpuTrack.java
@@ -36,6 +36,7 @@ import com.google.gapid.perfetto.views.CpuSliceSelectionView;
 import com.google.gapid.perfetto.views.CpuSlicesSelectionView;
 import com.google.gapid.perfetto.views.State;
 
+import java.util.Arrays;
 import org.eclipse.swt.widgets.Composite;
 
 import java.util.Collections;
@@ -99,7 +100,10 @@ public class CpuTrack extends Track.WithQueryEngine<CpuTrack.Data> {
   private ListenableFuture<Data> computeSummary(DataRequest req, Window w) {
     return transform(qe.query(summarySql(w.bucketSize)), result -> {
       int len = w.getNumberOfBuckets();
-      Data data = new Data(req, w.bucketSize, new long[len], new long[len], new double[len]);
+      long[] ids = new long[len], utids = new long[len];
+      Arrays.fill(ids, -1);
+      Arrays.fill(utids, -1);
+      Data data = new Data(req, w.bucketSize, ids, utids, new double[len]);
       result.forEachRow(($, r) -> {
         data.ids[r.getInt(0)] = r.getLong(1);
         data.utids[r.getInt(0)] = r.getLong(2);
--- a/gapic/src/main/com/google/gapid/perfetto/models/CpuTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/CpuTrack.java
@@ -49,8 +49,10 @@ import java.util.function.Consumer;
  */
 public class CpuTrack extends Track.WithQueryEngine<CpuTrack.Data> {
   private static final String SUMMARY_SQL =
-      "select quantum_ts, sum(dur)/cast(%d as float) " +
-      "from %s where cpu = %d and utid != 0 " +
+      "select quantum_ts, id, utid, sum(dur)/cast(%d as float) util from(" +
+          "select *, first_value(row_id) over win id, first_value(utid) over win utid " +
+          "from %s where cpu = %d and utid != 0 " +
+          "window win as (partition by quantum_ts order by dur desc)) " +
       "group by quantum_ts";
   private static final String SLICES_SQL =
       "select ts, dur, utid, row_id from %s where cpu = %d and utid != 0";
@@ -96,8 +98,13 @@ public class CpuTrack extends Track.WithQueryEngine<CpuTrack.Data> {
 
   private ListenableFuture<Data> computeSummary(DataRequest req, Window w) {
     return transform(qe.query(summarySql(w.bucketSize)), result -> {
-      Data data = new Data(req, w.bucketSize, new double[w.getNumberOfBuckets()]);
-      result.forEachRow(($, r) -> data.utilizations[r.getInt(0)] = r.getDouble(1));
+      int len = w.getNumberOfBuckets();
+      Data data = new Data(req, w.bucketSize, new long[len], new long[len], new double[len]);
+      result.forEachRow(($, r) -> {
+        data.ids[r.getInt(0)] = r.getLong(1);
+        data.utids[r.getInt(0)] = r.getLong(2);
+        data.utilizations[r.getInt(0)] = r.getDouble(3);
+      });
       return data;
     });
   }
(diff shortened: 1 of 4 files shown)


rschiu:
It crash when you select an space with no data, can you check that?

W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] Unhandled exception in the UI thread.
W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] java.lang.NullPointerException
W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] 	at com.google.gapid.perfetto.views.State.setSelectedThread(State.java:290)
W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] 	at com.google.gapid.perfetto.views.CpuPanel$2.click(CpuPanel.java:305)
W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] 	at com.google.gapid.perfetto.canvas.Panel$Hover$2.click(Panel.java:169)
....

gapic/src/main/com/google/gapid/perfetto/models/CpuTrack.java
```diff
@@ -49,8 +49,10 @@
  */
 public class CpuTrack extends Track.WithQueryEngine<CpuTrack.Data> {
   private static final String SUMMARY_SQL =
-      "select quantum_ts, sum(dur)/cast(%d as float) " +
-      "from %s where cpu = %d and utid != 0 " +
+      "select quantum_ts, id, utid, sum(dur)/cast(%d as float) util from(" +
+          "select *, first_value(row_id) over win id, first_value(utid) over win utid " +
```
rschiu: This will select the longest slice.  Is that by design?
Intuitively, I kind of expect all the all slices in the quantum to be selected.

stellama0208: Hey Raymond, thanks for giving the suggestion, the new commit implements the whole selection, feel free to check out.

stellama0208:
> It crash when you select an space with no data, can you check that?
> 
> W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] Unhandled exception in the UI thread.
> W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] java.lang.NullPointerException
> W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] at com.google.gapid.perfetto.views.State.setSelectedThread(State.java:290)
> W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] at com.google.gapid.perfetto.views.CpuPanel$2.click(CpuPanel.java:305)
> W20200227-174802198[main][Main$UI.lambda$registerWindowExceptionHandler$0] at com.google.gapid.perfetto.canvas.Panel$Hover$2.click(Panel.java:169)
> ....

Thanks for noticing that, it's addressed in the new commit, PTAL!

silence-do-good:
When you click at a peak in the graph, most often, the adjacent ones also get highlighted. Is that an intended behavior?

rschiu:
> When you click at a peak in the graph, most often, the adjacent ones also get highlighted. Is that an intended behavior?

This makes sense.  Selected slices often span multiple quanta.

gapic/src/main/com/google/gapid/perfetto/models/ProcessSummaryTrack.java
```diff
@@ -35,8 +36,11 @@
   // "where cpu < %numCpus%" is for performance reasons of the window table.
   private static final String PROCESS_VIEW_SQL = "select * from sched where utid in (%s)";
   private static final String SUMMARY_SQL =
-      "select quantum_ts, sum(dur)/cast(%d * %d as float) " +
-      "from %s group by quantum_ts";
+      "select quantum_ts, id, utid, sum(dur)/cast(%d * %d as float) util from(" +
+          "select *, first_value(row_id) over win id, first_value(utid) over win utid " +
+          "from %s " +
+          "window win as (partition by quantum_ts order by dur desc)) " +
+      "group by quantum_ts";
```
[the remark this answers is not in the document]
stellama0208: Make sense, it's the same in Process Panel now.

gapic/src/main/com/google/gapid/perfetto/models/CpuTrack.java
```diff
@@ -96,8 +101,14 @@ public CpuTrack(QueryEngine qe, CpuInfo.Cpu cpu) {
 
   private ListenableFuture<Data> computeSummary(DataRequest req, Window w) {
     return transform(qe.query(summarySql(w.bucketSize)), result -> {
-      Data data = new Data(req, w.bucketSize, new double[w.getNumberOfBuckets()]);
-      result.forEachRow(($, r) -> data.utilizations[r.getInt(0)] = r.getDouble(1));
+      int len = w.getNumberOfBuckets();
+      String[] concatedIds = new String[len];
+      Arrays.fill(concatedIds, "");
```
silence-do-good: fill with empty string is redundant I think. The new operator would initialize with empty string I believe.

silence-do-good: Nevermind.

stellama0208: It initialized my array with null sadly, went through a bug fix so that I added this line. T_T

silence-do-good: Yeah I realized it doesn't happen. My bad

stellama0208:
> When you click at a peak in the graph, most often, the adjacent ones also get highlighted. Is that an intended behavior?



> > When you click at a peak in the graph, most often, the adjacent ones also get highlighted. Is that an intended behavior?
> 
> This makes sense. Selected slices often span multiple quanta.

Yes, it's related with three facts happening in this process:
1. A selection on a quanta bucket marks up all the slices ids inside.
2. Sometimes the last slice id in a quanta is the same as the first slice id in the next quanta resulted with slice cutting.
3. If in a quanta there's at least one slice marked as selected, we see the quanta as selected and render the highlight.

I gave some thoughts around solving this issue. 
One solution would be to change assumptions 1 or 3 above, but changing either one doesn't look that good to me, they are comparatively the best options to me so far. 
Another approach could be special dealing with the first slice and the last slice in a quanta, but it may be fragile or complicate the code too much. 

So far I think it will not confuse our user that much, since it basically serves as a preview and as user zoom in the precision will go higher and higher. So I'd like to keep it this way to check in with the above considerations.

silence-do-good:
> So far I think it will not confuse our user that much, since it basically serves as a preview and as user zoom in the precision will go higher and higher. So I'd like to keep it this way to check in with the above considerations.

I'm okay with leaving it how it is now. I got a little confused initially because we draw a small circle when hovering over a peak. But clicking it ends up selecting adjacent tiles as well. But yeah, I agree lets not complicate stuff with minor implications.

silence-do-good:
Is it possible to add a % util for each of the slices under the selection view here? It shows up in the hover but I feel it might be useful to have a % shown up in the selection, so that people can see how much each process was utilizing.

stellama0208:
> Is it possible to add a % util for each of the slices under the selection view here? It shows up in the hover but I feel it might be useful to have a % shown up in the selection, so that people can see how much each process was utilizing.

Ah we couldn't do it actually. The % util is tied to a quanta bucket in the summary view, but not a single slice. The definition we give util here is `duration time of all the slices inside a quanta / the total time of a quanta`, so we actually don't have a correct number to present in the selection view for each slice.

PR #99, opened 2020-02-28
Fix bad indices for debug markers

apbodnar:
AFAIK, we're incorrectly popping multiple markers per `popMarker` in apps like Underlords and ultimately generating bad command indices that kills the server. (Expect a follow up PR that makes the server more robust to bad indices)

### DIFF
--- a/gapis/api/vulkan/vulkan.go
+++ b/gapis/api/vulkan/vulkan.go
@@ -179,29 +179,10 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
 	}
 
 	popMarker := func(ty MarkerType, id uint64, nCommands uint64) {
-		numMarkers := len(markerStack)
-		inStack := false
-		for i := numMarkers - 1; i >= 0 && len(markerStack) > 0; i-- {
+		if len(markerStack) > 0 {
 			marker := markerStack[len(markerStack)-1]
-			if marker.ty == ty {
-				inStack = true
-				break
-			}
-		}
-		if !inStack {
-			return
-		}
-
-		if id < nCommands-1 {
-			id++
-		}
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/vulkan.go
```diff
@@ -179,29 +179,10 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
...
-			return
-		}
-
-		if id < nCommands-1 {
-			id++
-		}
-		for i := numMarkers - 1; i >= 0 && len(markerStack) > 0; i-- {
-			marker := markerStack[len(markerStack)-1]
-			d.SubCommandMarkerGroups.NewMarkerGroup(marker.parent, marker.name, marker.start, id)
+			d.SubCommandMarkerGroups.NewMarkerGroup(marker.parent, marker.name, marker.start, id+1)
```
chrisforbes: For other reviewers: the other creator of marker groups (popping everything that got left open) treats the interval as half-open, we're being consistent with them. Previously we'd end up leaving the pop command floating outside the group if it was the last thing in the command buffer.

Just always popping one thing if the stack is nonempty changes the grouping behavior slightly for ill-formed programs. In the case of improper nesting [push(A) push(B) pop(A) pop(B)] then we'll produce a slightly different result. In the case where a push and a pop don't come from the same *level* of the command buffer then the result is still undefined, but it was before as well.

chrisforbes: I'm happy with the slight behavior changes in weird cases, in exchange for the simplification overall + being more robust to well-defined things with multiple nested markers + secondary command buffers.

chrisforbes:
One possible enhancement we considered (but have not implemented so far) is to ensure that popMarker can't pop a marker from the wrong command buffer level. The unwinding of leftover markers at the end of a command buffer is careful to do this.

PR #103, opened 2020-02-29
Prevent repeated descriptors in PV shader stage tables

hliatis:
Bug: b/150420371

### DIFF
--- a/gapis/api/vulkan/api/pipeline.api
+++ b/gapis/api/vulkan/api/pipeline.api
@@ -227,6 +227,11 @@ class CreatedGraphicsPipelines {
   map!(u32, ref!GraphicsPipelineObject) Objects
 }
 
+
+@internal class MutableSet {
+  map!(u64, bool) Set
+}
+
 @indirect("VkDevice")
 @threadsafe
 cmd VkResult vkCreateGraphicsPipelines(
@@ -274,6 +279,8 @@ cmd VkResult vkCreateGraphicsPipelines(
     }
     obj.Flags = create_info.flags
 
+    seenDescs := MutableSet()
+
     pipeline_shader_stage_create_infos :=
     create_info.pStages[0:create_info.stageCount]
     for j in (0 .. create_info.stageCount) {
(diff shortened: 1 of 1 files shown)


hliatis:
Yes, and I originally tried to see how I can resolve that, but it might take some redesigning of how we populate them in the first place

hliatis:
Yes, but it might involve some heavier re-plumbing, at the moment. I attempted to do so in pipeline.api (vkCreateGraphicsPipelines), but I was really just doing the same thing but in an api file.

chrisforbes:
Andrew: Hernan and I discussed one possible future of *not* mashing the descriptors for every stage together into one global list, only to pull it apart again -- any thoughts there? Perf characteristics would change slightly.

AWoloszyn:
> Andrew: Hernan and I discussed one possible future of _not_ mashing the descriptors for every stage together into one global list, only to pull it apart again -- any thoughts there? Perf characteristics would change slightly.

We go through the list quite often. Being able to make it as small as possible would be better for performance.

AWoloszyn:
https://github.com/google/gapid/blob/62f089301c6ba1245aa155ca055c5a80e75b2ce7/gapii/cc/vulkan_extras.cpp#L653 Should I think be the right place to deduplicate the used descriptors, so that the list never gets overly large in the first place.

hliatis:
It looks that function does the right thing in the sense that it is called per module. Sure, I can still prevent duplication by creating a map/set outside the function and preventing additions. This would create the right answer in PV, but then each descriptor would only exist in ONLY one shader module in state view. 

The problem isn't necessarily from this function, but rather from the creation of the pipeline itself. It combines the results of each individual call to this function, and since each shader module can (correctly, unless I am wrong) share descriptors, this will inevitably create duplicates. As I have said before I wanted to solve this in vkCreateGraphicsPipelines, but I need to be able to create a local map!() to be able to do that. Am I able to do that? If so, how?

Alternatively, I could instead grab the descriptors from the shader modules themselves just like how I already grab the code from them. This would skip over using UsedDesciptors completely or in an edge case where I need to check against it. The one caveat to that approach is whether or not I can guarantee that information won't be deleted by pipeline bind time.

AWoloszyn:
Aha fair. Sorry I forgot that this was done per-shader-module.

Locals are (as you have figured out) immutable, however only the top-level. We have used this elsewhere in the code, but creating a 
```
struct MutableXMap {
   map!(foo, bar) myMap
}

cmd..... () {
    map := MutableXMap().myMap;
}
```
Will give you the map you want.

gapis/api/vulkan/api/pipeline.api
```diff
@@ -305,24 +312,31 @@ cmd VkResult vkCreateGraphicsPipelines(
...
         for _, _, d in moduleDescriptors {
-          obj.UsedDescriptors[len(obj.UsedDescriptors)] = d
-        }
+          if !seenDescs.Set[(as!u64(d.Set)<<32) | as!u64(d.Binding)] {
+            obj.UsedDescriptors[len(obj.UsedDescriptors)] = d
+            seenDescs.Set[(as!u64(d.Set)<<32) | as!u64(d.Binding)] = true
+          }
+        } 
       } else {
         nSets := len(obj.Layout.SetLayouts)
```
chrisforbes: What is this second branch even for?

chrisforbes: @AWoloszyn this looks like "consider all descriptors to be used if we couldn't figure it out" -- is this a path we actually want to take, or should we just treat this as a bug?

AWoloszyn: Used descriptors is an optimisation, so if we can't figure it out, then falling back to slow path is reasonable to me.

Also making the tool useless to our users if we can't parse a shader seems unhelpful, when we could provide the same functionality in a different way.

Secondly since this is an external project we use, getting fixes may take longer than we would like.

PR #104, opened 2020-03-02
Update Android symbols file names

hevrard:
This enables to match all symbol files with `gapir*.sym`

Bug: b/150627448


yalcinmelihyasin:
This requires a change in release tool as well. Please create a cl for it.

PR #107, opened 2020-03-03
Stop overriding GPU queue track names

chrisforbes:
Early versions of the profiling support provided 'ugly' track names, but
that has since been fixed. Just present the GPU queue names as provided
by the driver.

This fixes the 'GPU Queue 48' quirk in the replay profiler,
and prepares us for more descriptive track names from other drivers.

Bug: b/150692802


chrisforbes:
CI-linux failure is almost certainly unrelated, but rerunning.

PR #109, opened 2020-03-04
Generate links to renderpasses from gpu slices

apbodnar:
Redux of https://github.com/google/agi/pull/84 that uses sync.Data to build the slice links


gapis/api/cmd_id.go
```diff
@@ -51,7 +51,9 @@ func (id CmdID) String() string {
 }
 
 // TODO(apbodnar) find a more appropriate place for this
-type CommandSubmissionKey struct {
-	SubmissionOrder uint64
-	CommandBuffer   int64
+type CmdSubmissionKey struct {
+	First  uint64
```
chrisforbes: Not really thrilled about this naming :)

but understand we don't want to leak Vulkanisms everywhere

AWoloszyn: Can you make this
```
type CmdSubmissionKey []uint64
```

Then it is both generic, and does not need individualized named values.

AWoloszyn:
LGTM

PR #111, opened 2020-03-04
Enable compact sched events if available.

pmuetschard:


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -224,7 +224,9 @@ public static String getConfigSummary(Settings settings, Device.PerfettoCapabili
             .getFtraceConfigBuilder()
             .addAllFtraceEvents(Arrays.asList(PROCESS_TRACKING_FTRACE))
             .setDrainPeriodMs(FTRACE_DRAIN_PERIOD)
-            .setBufferSizeKb(FTRACE_BUFFER_SIZE);
+            .setBufferSizeKb(FTRACE_BUFFER_SIZE)
+            .setCompactSched(PerfettoConfig.FtraceConfig.CompactSchedConfig.newBuilder()
```
rschiu: Isn't this supposed to be getCompactSchedConfigBuilder()?

pmuetschard: no, because we want to hold on to the ftrace one, not this one.

PR #112, opened 2020-03-05
Add a global filter for the tracks

silence-do-good:
An important feedback from partners was to have a way to filter the tracks after taking the trace. The use-case was not knowing which counters to enable at the time of taking a trace, thereby selecting all counters. This results in a bucket load of counter tracks and might be hard to navigate through. The filter option here enables/disables the tracks from rendering altogether. 

Bug: b/150172928


silence-do-good:
The position of the filter button is currently in the top right, next to the help button. This can be finalized through a UI discussion if needed.

gapic/src/main/com/google/gapid/perfetto/views/TrackSelector.java
```diff
@@ -61,8 +61,6 @@ public TrackSelector(Composite parent, State.ForSystemTrace state, Theme theme)
     tree = createTree(container, theme);
```
rschiu: nit: Group in line 55 is superfluous.

(I couldn't figure out how to add comment on an unchanged line)

gapic/src/main/com/google/gapid/perfetto/models/TrackConfig.java
```diff
@@ -43,10 +43,28 @@ public TrackConfig(ImmutableList<Element<?>> elements) {
     public final String name;
     protected final T uiFactory;
 
+    // Track filtering info
+    protected boolean checked;
+    public Element<?> parent;
+
```
pmuetschard: This class should not be mutable.

gapic/src/main/com/google/gapid/perfetto/models/TrackConfig.java
```diff
@@ -77,11 +98,30 @@ public Group(String id, String name, ImmutableList<Element<?>> tracks, UiFactory
 
     @Override
     public CopyablePanel<?> createUi(State.ForSystemTrace state) {
+      if (!checked) {
```
pmuetschard: This factory method should never return null.

gapic/src/main/com/google/gapid/perfetto/views/TraceComposite.java
```diff
@@ -56,16 +56,24 @@
...
   private final RootPanel<S> rootPanel;
   private final PanelCanvas canvas;
+  protected final TopBar topBar;
+  protected FilterDialog filterDialog;
 
   public TraceComposite(Composite parent, Analytics analytics, Theme theme) {
     super(parent, SWT.NONE);
     this.state = createState();
     this.rootPanel = createRootPanel();
+    this.filterDialog = createFilterDialog();
```
pmuetschard: The dialog should only be created just before it's shown.

gapic/src/main/com/google/gapid/perfetto/views/TraceComposite.java
```diff
@@ -200,6 +208,7 @@ public TraceComposite(Composite parent, Analytics analytics, Theme theme) {
 
   protected abstract S createState();
   protected abstract RootPanel<S> createRootPanel();
+  protected abstract FilterDialog createFilterDialog();
```
pmuetschard: We don't actually expect FilterDialog to be subclassed, so I don't really understand this abstract method. If you want subclasses to be able to customize the TopBar, create an abstract function that passes it in, or create a method returning a boolean of whether to have a filter dialog or not.

silence-do-good:
It is indeed different from what Carl suggested. I've requested Fran to use this for a couple of days and provide us feedback. Will set up a meeting with Carl and all of us to discuss the UX aspect of this, after Fran gives us his feedback.

pau-baiget:
Is this PR moving somewhere? @silence-do-good @pmuetschard

silence-do-good:
> Is this PR moving somewhere? @silence-do-good @pmuetschard

Not directly. But I'd like to keep this open until I'm able to get to the new Filters, for reference and comparison. I'll close this out when the new one is merged in.

PR #113, opened 2020-03-05
Update frame lifecycle capability based on sdk and preview_sdk versions

silence-do-good:
The name property has been updated to Android 10, for R builds. This is
same as Q and hence cannot be used to distinguish between the two
dessert releases. For now, we depend on sdk and preview_sdk versions to
distinguish them.

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -331,8 +331,7 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
 		result.GpuProfiling = gpu
 	}
 
-	if b.Instance().GetConfiguration().GetOS().GetName() == "R" {
-		// TODO(b/146384733): Change this to API version when it releases
+	if b.Instance().GetConfiguration().GetOS().GetAPIVersion() >= 30 {
 		gpu.HasFrameLifecycle = true
 	}
 
--- a/core/os/device/deviceinfo/cc/android/query.cpp
+++ b/core/os/device/deviceinfo/cc/android/query.cpp
@@ -328,6 +328,9 @@ bool createContext() {
 
   GET_STRING_PROP("ro.build.version.release", gContext.mOSName);
   GET_INT_PROP("ro.build.version.sdk", gContext.mOSVersion);
+  // preview_sdk is used to determine the version for the next OS release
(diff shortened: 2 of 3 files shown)


core/os/android/adb/commands.go
```diff
@@ -331,9 +332,15 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
 		result.GpuProfiling = gpu
 	}
 
-	if b.Instance().GetConfiguration().GetOS().GetName() == "R" {
-		// TODO(b/146384733): Change this to API version when it releases
-		gpu.HasFrameLifecycle = true
+	if sdk,err := b.SystemProperty(ctx, "ro.build.version.sdk"); err == nil {
```
lpy: I think we can wait until `GetAPI()` returns the api level for Android 11

core/os/device/deviceinfo/cc/android/query.cpp
```diff
@@ -328,6 +328,9 @@ bool createContext() {
 
   GET_STRING_PROP("ro.build.version.release", gContext.mOSName);
   GET_INT_PROP("ro.build.version.sdk", gContext.mOSVersion);
+  int previewSdk = 0;
+  GET_INT_PROP("ro.build.version.preview_sdk", previewSdk);
+  gContext.mOSVersion += previewSdk;
```
rschiu: mOSVersion is used in the switch statement a little later. Check how mOSVersionMajor/Minor is used and see if it can break anything and check if we have any weird cases where we have different ways to get the OS version.  I'd imagine you will at least have to add a case 30 for Android 11.

core/os/android/adb/commands.go
rschiu: Charlie is partly blocked on this, so I would prefer this change sooner rather than later.  Also, this change should be able avoid something similar happening again next year.

core/os/device/deviceinfo/cc/android/query.cpp
silence-do-good: Yeah. Peiyong has put up a PR for that. I'll merge this in only after that merges in.

lpy:
Put a comment about we treat preview OS version as next OS release and hence we bump the API level :)

PR #115, opened 2020-03-05
Fixes the pipeline barrier for getting framebuffer during trace

yalcinmelihyasin:


chrisforbes:
(Discussed issues with this offline)

PR #116, opened 2020-03-05
Mali replay profiling

chrisforbes:


gapis/api/vulkan/replay.go
```diff
@@ -1049,7 +1049,7 @@ func (a API) Replay(
 			makeReadable.imagesOnly = true
 			optimize = false
 			transforms.Add(NewWaitForPerfetto(req.traceOptions, req.handler, req.buffer))
-			transforms.Add(&profilingLayers{})
+			//transforms.Add(&profilingLayers{})
```
chrisforbes: WIP :)

AWoloszyn:
So, we have to process the mali data differently than other data? This seems to defeat some of the purpose of the standardized data format?

chrisforbes:
I would like to collapse together most of this. There are only some very minor differences (which slices are interesting, how many of them we expect, etc)

gapis/api/vulkan/replay.go
chrisforbes: This is now fixed.

chrisforbes:
CI-linux-test failure is unrelated flake

gapis/trace/android/mali/profiling_data.go
```diff
@@ -0,0 +1,281 @@
+// Copyright (C) 2019 Google Inc.
```
lpy: 2020

PR #122, opened 2020-03-06
Converts "~" character to user home while saving trace file

yalcinmelihyasin:


yalcinmelihyasin:
This will only solves the problem if the path starts with "~" but I don't expect too many edge cases with normal usage.

PR #128, opened 2020-03-09
Wip various

chrisforbes:
Rollup of experimental changes


gapic/src/main/com/google/gapid/views/CommandTree.java
```diff
@@ -157,7 +157,7 @@ protected void updateModel(Event e) {
           // Copy before loaded. Not ideal, but this is unlikely.
           result.append("Loading...");
         } else {
-          result.append(Formatter.toString(cmd, models.constants::getConstants));
+          result.append(Formatter.toString(cmd, models.constants::getConstants, data.getRepresentation().getIndicesCount() > 1));
```
chrisforbes: None of this is really the right way to do it. We should produce the right thing on the backend.

PR #129, opened 2020-03-10
Only show pipeline view when a draw command is selected

hliatis:
Bug: b/150943958

### DIFF
--- a/gapic/src/main/com/google/gapid/views/PipelineView.java
+++ b/gapic/src/main/com/google/gapid/views/PipelineView.java
@@ -87,6 +87,7 @@ import java.util.List;
 import java.util.HashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Logger;
+import java.util.logging.Level;
 
 /**
  * View the displays the information for each stage of the pipeline.
@@ -154,6 +155,15 @@ public class PipelineView extends Composite
   }
 
   @Override
+  public void onCommandsLoaded() {
+    if (!models.commands.isLoaded()) {
+      loading.showMessage(Error, Messages.CAPTURE_LOAD_FAILURE);
+    } else if (models.commands.getSelectedCommands() == null) {
+      loading.showMessage(Info, Messages.SELECT_DRAW_CALL);
+    }
+  }
+
+  @Override
   public void onCommandsSelected(CommandIndex path) {
     updatePipelines();
   }
@@ -195,7 +205,7 @@ public class PipelineView extends Composite
         }
       });
     } else {
-      loading.showMessage(Info, Messages.SELECT_COMMAND);
+      loading.showMessage(Info, Messages.SELECT_DRAW_CALL);
     }
   }
 
--- a/gapis/resolve/resource_data.go
+++ b/gapis/resolve/resource_data.go
@@ -25,7 +25,9 @@ import (
 	"github.com/google/gapid/gapis/api/sync"
 	"github.com/google/gapid/gapis/capture"
 	"github.com/google/gapid/gapis/database"
+	"github.com/google/gapid/gapis/messages"
 	"github.com/google/gapid/gapis/resolve/initialcmds"
+	"github.com/google/gapid/gapis/service"
 	"github.com/google/gapid/gapis/service/path"
 )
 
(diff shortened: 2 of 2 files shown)


hliatis:
The next steps would be to create a boolean flag in api.command for when a command is a draw call or not. However, I want to ask if I am on the right track first. I know we can use Events, but I'm not seeing a flow logic with them since Events contain a Command and not the other way around.

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -153,9 +154,54 @@ public void onResourcesLoaded() {
...
+      loading.showMessage(Info, Messages.SELECT_DRAW_CALL);
+    } else  {
+      Rpc.listen(models.commands.loadCommand(path.getCommand(), models.devices.getReplayDevicePath()), 
+          new UiErrorCallback<API.Command, String, Loadable.Message>(this, LOG) {
+        @Override
+        protected ResultOrError<String, Loadable.Message> onRpcThread(
+            Rpc.Result<API.Command> result) {
+          try{
+            LOG.log(Level.SEVERE, result.get().getName());
+            if (result.get().getName().equals("vkCmdDrawIndexed")) {
```
hliatis: I don't really want to do it this way. Should we add a bool to the API object?

gapis/resolve/resource_data.go
```diff
@@ -180,6 +180,19 @@ func (r *ResourceDataResolvable) Resolve(ctx context.Context) (interface{}, erro
...
+	if err != nil {
+		return nil, err
+	}
+	state, err := GlobalState(ctx, &path.GlobalState{After: p.After}, r)
+	if err != nil {
+		return nil, err
+	}
+
+	if !cmd.CmdFlags(ctx, api.CmdID(p.After.Indices[0]), state).IsExecutedDraw() {
+		return nil, fmt.Errorf("Select a draw call.")
```
hliatis: This is roughly the same message as Messages.SELECT_DRAW_CALL, but it includes the red error icon. I am open to a solution that doesn't duplicate this string.

AWoloszyn: Anything in the UI should be using messages from the messages file.
gapis/messages/en-us.stb.md

AWoloszyn: I think this is still not QUITE what you want. Since I think it will grab ALL piplines
when you click any draw-call. This will run into the same problems as just clicking on a random command.

I wonder if we could just ONLY present bound pipelines? (That is sad because it prevents us from inspecting other pipelines in the state).

Then we would just return NO pipelines for the  case where you have not clicked on something valid.

Also is does IsExecutedDraw return true for a top-level VkCmdDraw() call? Which is not itself a draw-call but a recording of one?

AWoloszyn: I think this will also drop all Dispatch calls?

hliatis: I am not sure what you are asking me with dispatches. They are also decorated with @executed_draw in the API files.

hliatis: "I wonder if we could just ONLY present bound pipelines? (That is sad because it prevents us from inspecting other pipelines in the state).

Then we would just return NO pipelines for the case where you have not clicked on something valid."

This would still show pipelines between draws, which I believe we don't want

hliatis:
> Have we ensured that IsExecutedDraw is _only_ set on the execution of the draw, not its recording?
> 
> If so, looks good.

Unfortunately, it might seem to be the case that it's considered during recording. However, nothing shows up during the recording.

chrisforbes:
Suggest tweaking to ensure the current command is a subcommand, then.

gapis/resolve/resource_data.go
```diff
@@ -180,6 +182,19 @@ func (r *ResourceDataResolvable) Resolve(ctx context.Context) (interface{}, erro
...
+	cmd, err := Cmd(ctx, p.After, r)
+	if err != nil {
+		return nil, err
+	}
+	state, err := GlobalState(ctx, &path.GlobalState{After: p.After}, r)
+	if err != nil {
+		return nil, err
+	}
+
+	if !cmd.CmdFlags(ctx, api.CmdID(p.After.Indices[0]), state).IsExecutedDraw() || len(p.After.Indices) == 1 {
```
hliatis: Let me know if there is more proper way to make sure only an executing (not recording) draw is being selected. So far, my tests are fine with this change.

chrisforbes: Nope, this is exactly what I would have done. We should probably add a helper method somewhere eventually.

PR #130, opened 2020-03-10
Update the logo to the new version... everywhere.

pmuetschard:


cmd/img2h/main.go
```diff
@@ -0,0 +1,102 @@
...
+	_ "image/jpeg"
+	_ "image/png"
+)
+
+var (
+	out = flag.String("out", "-", "Output file, '-' for stdout")
+)
+
+func main() {
+	app.ShortHelp = "img2h generates C++ headers containing image data"
```
hevrard: I would clarify that the header is Vulkan specific:
`img2h generates C++ headers containing Vulkan image data`

PR #132, opened 2020-03-10
vkUnmap memory of gapii staging buffer

hevrard:
Adding this vkUnmap removes what look like a memory leak during capture.

The spec for vkFreeMemory says:

> If a memory object is mapped at the time it is freed, it is implicitly unmapped.

So this unmap may not be strictly necessary. Yet it does fix a memory explosion
issue for a particular title, on two GPU vendors.

### DIFF
--- a/gapii/cc/vulkan_mid_execution.cpp
+++ b/gapii/cc/vulkan_mid_execution.cpp
@@ -193,6 +193,9 @@ class StagingBuffer {
       device_functions_.vkDestroyBuffer(device_, staging_buffer_, nullptr);
     }
     if (staging_memory_) {
+      if (bound_memory_) {
+        device_functions_.vkUnmapMemory(device_, staging_memory_);
+      }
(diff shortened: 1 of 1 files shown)


AWoloszyn:
Can we not get the gpu vendors in question to fix their buggy drivers?

prahladkg5:
This should be fixed in the driver, not the tool. Does validation layer trigger warnings/errors for this usecase?

prahladkg5:
Following up on my comment here, based on feedback this is just better use of Vulkan in general by gapii so we should just take this change as a way doing tidy cleanup of VKMemory.

AWoloszyn:
> Following up on my comment here, based on feedback this is just better use of Vulkan in general by gapii so we should just take this change as a way doing tidy cleanup of VKMemory.

This is incorrect, there IS a better use of Vulkan by sharing the staging resources, but this change is not that. This is simply a workaround.

PR #133, opened 2020-03-10
Separate vk debug marker layer from cpu timing layer

rschiu:
CPUTiming is often not need together with debug marker.
Separate the two layers make the code much easier to maintain.

Bug: b/148465074


rschiu:
Just to explain a bit more, I have ran into quite a bit of extra maintenance work due to the combined layers.  The original cpu timing layer was a simple layer with a single loop to generate all functions.  Combining the layers means I have special case certain functions and mix the logics together.  Then there are various annoyances like remembering to add the timing scope into the debug marker functions, not knowing which part of the layer causes an issue, etc.

PR #134, opened 2020-03-11
Remove GLES and GVR

chrisforbes:
Tested that trace & replay still work as expected against an Android device

Expect kokoro to find some issues with other targets


chrisforbes:
Macos build is stuck; passed in a previous run.

chrisforbes:
Build times have a lot of noise in them, but this does seem to cut ~10 minutes off.

gapii/cc/spy.h
```diff
@@ -29,7 +27,7 @@
 namespace gapii {
 struct spy_creator;
 class ConnectionStream;
-class Spy : public GlesSpy, public GvrSpy, public VulkanSpy {
+class Spy : public VulkanSpy {
```
AWoloszyn: Yay killing multiple inheritance!

kokoro/macos/build.sh
```diff
@@ -58,9 +58,7 @@ function build {
 
 # Build each API package separately first, as the go-compiler needs ~8GB of
 # RAM for each of the big API packages.
-for api in gles vulkan gvr; do
-  build //gapis/api/$api:go_default_library
-done
+build //gapis/api/vulkan:go_default_library
```
AWoloszyn: We probably want to build everything BUT vulkan first, then vulkan.
The old code built gles (and all dependencies) then started on the rest. This means that vulkan would build by itself, now it will build with all its dependencies, might cause some issues?
But also might not.

test/integration/service/service_test.go
```diff
@@ -113,27 +111,29 @@ var (
 )
 
 func init() {
-	check := func(err error) {
-		if err != nil {
-			panic(err)
-		}
-	}
-	ctx := context.Background()
+	// TODO: this required GLES support, which has since been removed.
```
AWoloszyn: In general we remove commented out code. We can always keep the "Fix this" comment, and remove the code, since we can go back in history to find the code.

chrisforbes: I'll just nuke the body of this function then.

gapii/cc/spy.cpp
```diff
@@ -278,7 +225,7 @@ Spy::~Spy() {
   endTraceIfRequested();
 }
 
-void Spy::resolveImports() { GlesSpy::mImports.resolve(); }
+void Spy::resolveImports() {}
```
pmuetschard: This function is no longer called. Delete it.

kokoro/macos/build.sh
pmuetschard: No, this is fine. This is no different than before, where the deps were built as part of the first API. What kills the VM is building the API package, which is always going to be the last thing built and on its own, as it is the only root of the dependency tree.

gapii/cc/spy.cpp
chrisforbes: Done.

chrisforbes:
[Trivial rebase]

PR #135, opened 2020-03-11
Create only one staging buffer per device

hevrard:
When retrieving the Vulkan state at the start of a mid-execution
capture, we create a new staging buffer for each of the buffers in the
state.

This change creates only one staging buffer per device, and reuses it
to retrieve all buffers of that device.

Bug: none
Test: capture a few Vulkan apps

### DIFF
--- a/gapii/cc/vulkan_mid_execution.cpp
+++ b/gapii/cc/vulkan_mid_execution.cpp
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include "core/cc/make_unique.h"
+
 #include "gapii/cc/state_serializer.h"
 #include "gapii/cc/vulkan_exports.h"
 #include "gapii/cc/vulkan_spy.h"
@@ -344,6 +346,18 @@ void VulkanSpy::serializeGPUBuffers(StateSerializer* serializer) {
     }
   }
 
+  // Retrieve buffers, one by one, chunk by chunk, to prevent OOM.
+  // Prepare one staging buffer of size kChunkSizeLimit per device.
+  std::unordered_map<VkDevice, std::unique_ptr<StagingBuffer>> staging_buffers;
+  for (auto it = mState.Devices.begin(); it != mState.Devices.end(); it++) {
+    VkDevice device = it->first;
+    staging_buffers[device] = make_unique<StagingBuffer>(
+        arena(), mImports.mVkDeviceFunctions[device], device,
+        mState.PhysicalDevices[mState.Devices[device]->mPhysicalDevice]
+            ->mMemoryProperties,
+        kChunkSizeLimit);
+  }
+
   for (auto& buffer : mState.Buffers) {
     VkBuffer buf_handle = buffer.first;
     auto buf = buffer.second;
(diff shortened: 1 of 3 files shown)


gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -344,6 +349,19 @@ void VulkanSpy::serializeGPUBuffers(StateSerializer* serializer) {
     }
   }
 
+  // Retrieve buffers, one by one, chunk by chunk, to prevent OOM.
+  // Prepare one staging buffer of size kChunkSizeLimit per device.
+  std::map<VkDevice, std::shared_ptr<StagingBuffer>> staging_buffers;
```
AWoloszyn: Unless you need these to be ordered, don't use std::map.
std::unordered_map please.

gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -446,8 +460,9 @@ void VulkanSpy::serializeGPUBuffers(StateSerializer* serializer) {
         memory::Observation observation;
         observation.set_pool(deviceMemory->mData.pool_id());
         observation.set_base(bind.mmemoryOffset + offset);
-        serializer->sendData(&observation, true, stage.GetMappedMemory(),
+        serializer->sendData(&observation, true, stage->GetMappedMemory(),
                              chunkSize);
+        stage->UnmapMemory();
```
AWoloszyn: You shouldn't have to unmap this buffer.
Leaving it mapped the entire time is probably simpler.
Just have to flush it before reading it.

gapii/cc/vulkan_mid_execution.cpp
AWoloszyn: Why shared_ptr. I don't see any reason for needing a shared_ptr here.

hevrard: done

gapii/cc/vulkan_mid_execution.cpp
hevrard: done

gapii/cc/vulkan_mid_execution.cpp
hevrard: That's because StagingBuffer has a field which is a reference. How would you do it?

AWoloszyn: std::unique_ptr.
shared_ptrs should be limited where we (incredibly rarely) need shared ownership.

gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -446,7 +455,7 @@ void VulkanSpy::serializeGPUBuffers(StateSerializer* serializer) {
         memory::Observation observation;
         observation.set_pool(deviceMemory->mData.pool_id());
         observation.set_base(bind.mmemoryOffset + offset);
-        serializer->sendData(&observation, true, stage.GetMappedMemory(),
+        serializer->sendData(&observation, true, stage->GetMappedMemory(),
```
AWoloszyn: This also won't quite work. This will end up double-mapping memory. We should just map the memory on staging buffer creation, then we don't have to map it every time we need the contents.

hevrard: Isn't the `if (!bound_memory_)` in GetMappedMemory() forcing to map only once?

AWoloszyn: My mistake, yes that's fine.

gapii/cc/vulkan_mid_execution.cpp
hevrard: done, creating core/cc/make_unique.h along the way.

PR #139, opened 2020-03-13
Kill jdwp

chrisforbes:


chrisforbes:
This builds on top of the previous change to remove the GLES API. Only the final commit is interesting.

gapii/vulkan/vk_graphics_spy/cc/layer.cpp
```diff
@@ -36,21 +36,9 @@ typedef void* (*eglGetProcAddress)(const char* procname);
```
hevrard: on L28, remove: `typedef void* (*eglGetProcAddress)(const char* procname);`

chrisforbes: Done.

chrisforbes:
[trivial rebase]

PR #149, opened 2020-03-18
Remove api.Context machinery

chrisforbes:
This is dead code now that the GUI user of it is gone.


chrisforbes:
I had initially intended to just remove the CommandTree interaction with this, but kept pulling on the string. This removes the notion of API contexts completely, as our one remaining API doesn't have them.

chrisforbes:
Not ready, this breaks some things.

cmd/gapit/common.go
```diff
@@ -47,37 +47,6 @@ import (
 
 func (f CommandFilterFlags) commandFilter(ctx context.Context, client service.Service, p *path.Capture) (*path.CommandFilter, error) {
```
hevrard: This may disappear entirely?

cmd/gapit/flags.go
```diff
@@ -141,8 +141,6 @@ type (
 		CaptureID bool `help:"if true then interpret the capture file argument as a capture ID that is already loaded in gapis"`
 	}
 	CommandFilterFlags struct {
```
hevrard: Shall we just get rid of the whole struct?

cmd/gapit/report.go
```diff
@@ -34,9 +34,7 @@ type reportVerb struct{ ReportFlags }
 func init() {
 	verb := &reportVerb{
 		ReportFlags: ReportFlags{
-			CommandFilterFlags: CommandFilterFlags{
-				Context: -1,
-			},
+			CommandFilterFlags: CommandFilterFlags{},
 		},
```
hevrard: Get rid of all this?

chrisforbes: I want to reuse this machinery for more powerful command filtering soon. Just removing the dead filter criteria

cmd/gapit/common.go
pmuetschard: let's hold off on removing it entirely, since we are in discussion for other filtering.

cmd/gapit/flags.go
pmuetschard: ditto

cmd/gapit/report.go
pmuetschard: i don't think we want to filter the report, though, so here we should remove it.

chrisforbes:
Changes done.

chrisforbes:
Fixed resolve test

cmd/gapit/report.go
```diff
@@ -33,11 +33,7 @@ type reportVerb struct{ ReportFlags }
 
 func init() {
 	verb := &reportVerb{
-		ReportFlags: ReportFlags{
-			CommandFilterFlags: CommandFilterFlags{
-				Context: -1,
-			},
-		},
+		ReportFlags: ReportFlags{},
```
pmuetschard: this is unnecessary, unspecified fields are initialized to their zero value.

chrisforbes: Done.

chrisforbes:
Updated with test fix for macos smoketests

PR #150, opened 2020-03-18
Make sure ready signal is fired.

lpy:
Make sure ready signal is fired when trace session is ready to start.

Bug: b/147490759


chrisforbes:
Needs to be picked to 0.9.x branch

PR #155, opened 2020-03-18
Do get the gfxtrace artifact

hevrard:
I am not sure why we do not get it with the current code.
See experiment with #147  where we do obtain it.

### DIFF
--- a/kokoro/linux/build.sh
+++ b/kokoro/linux/build.sh
@@ -123,6 +123,6 @@ test "${APP_EXIT_STATUS}" -eq 130
 
 # TODO(https://github.com/google/gapid/issues/3163): The coherent memory
 #  tracker must be disabled with SwiftShader for now.
-xvfb-run -e xvfb.log -a bazel-bin/pkg/gapit trace -device host -disable-coherentmemorytracker -disable-pcs -disable-unknown-extensions -record-errors -no-buffer -api vulkan -start-at-frame 5 -capture-frames 10 -observe-frames 1 -out out/dist/vulkan_sample.gfxtrace bazel-bin/cmd/vulkan_sample/vulkan_sample
+xvfb-run -e xvfb.log -a bazel-bin/pkg/gapit trace -device host -disable-coherentmemorytracker -disable-pcs -disable-unknown-extensions -record-errors -no-buffer -api vulkan -start-at-frame 5 -capture-frames 10 -observe-frames 1 -out $KOKORO_ARTIFACTS_DIR/vulkan_sample.gfxtrace bazel-bin/cmd/vulkan_sample/vulkan_sample
 
-xvfb-run -e xvfb.log -a bazel-bin/pkg/gapit video -gapir-nofallback -type sxs -frames-minimum 10 -out vulkan_sample.mp4  out/dist/vulkan_sample.gfxtrace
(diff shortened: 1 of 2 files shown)


kokoro/linux/build.sh
```diff
@@ -84,6 +84,8 @@ set -e
 # Build the release packages.
 mkdir $BUILD_ROOT/out
 $SRC/kokoro/linux/package.sh $BUILD_ROOT/out
+ARTIFACT_DIR=$BUILD_ROOT/out/dist
+mkdir -p $ARTIFACT_DIR
```
pmuetschard: `out/dist` is deleted and re-created by the `package.sh` script, so really no reason to create it here. Better to put these things somewhere else in `out`, since `dist` is meant for the distributable part of the build, not for side-artifacts.

kokoro/linux/build.sh
```diff
@@ -123,6 +125,9 @@ test "${APP_EXIT_STATUS}" -eq 130
 
 # TODO(https://github.com/google/gapid/issues/3163): The coherent memory
 #  tracker must be disabled with SwiftShader for now.
-xvfb-run -e xvfb.log -a bazel-bin/pkg/gapit trace -device host -disable-coherentmemorytracker -disable-pcs -disable-unknown-extensions -record-errors -no-buffer -api vulkan -start-at-frame 5 -capture-frames 10 -observe-frames 1 -out out/dist/vulkan_sample.gfxtrace bazel-bin/cmd/vulkan_sample/vulkan_sample
+xvfb-run -e xvfb.log -a bazel-bin/pkg/gapit trace -device host -disable-coherentmemorytracker -disable-pcs -disable-unknown-extensions -record-errors -no-buffer -api vulkan -start-at-frame 5 -capture-frames 10 -observe-frames 1 -out $ARTIFACT_DIR/vulkan_sample.gfxtrace bazel-bin/cmd/vulkan_sample/vulkan_sample
 
-xvfb-run -e xvfb.log -a bazel-bin/pkg/gapit video -gapir-nofallback -type sxs -frames-minimum 10 -out vulkan_sample.mp4  out/dist/vulkan_sample.gfxtrace
+sync
```
pmuetschard: why? this looks like grasping at straws.

kokoro/linux/build.sh
```diff
@@ -123,6 +125,9 @@ test "${APP_EXIT_STATUS}" -eq 130
...
+ls $ARTIFACT_DIR/*.gfxtrace
```
pmuetschard: remove

kokoro/linux/build.sh
hevrard: I was using out/dist because of the Kokoro config:

```
    # b/151297033: save trace of sample on swiftshader, for debug purposes                                                                                                   
    regex: "out/dist/*.gfxtrace"
    strip_prefix: "out/dist"
```

But I get your point, I just updated to have the gfxtrace artifact outsite of out/dist/

kokoro/linux/build.sh
hevrard: removed

kokoro/linux/build.sh
hevrard: done

PR #156, opened 2020-03-18
Remove more remnants of GLES support

chrisforbes:


chrisforbes:
This removes the remaining libinterceptor plumbing, and the GLES-only dependency graph

gapii/cc/connection_header.cpp
```diff
@@ -60,8 +59,7 @@ bool ConnectionHeader::read(core::StreamReader* reader) {
   if (!reader->read(mObserveFrameFrequency) ||
```
chrisforbes: Should this be bumping the version to 3?

pmuetschard: since you are changing the size of the header, yes we should change the version number.

gapii/client/header.go
```diff
@@ -38,14 +38,13 @@ const version = 2
...
-//     char     mLibInterceptorPath[MAX_PATH]; // Path to libinterceptor.so
 // };
 //
 // All fields are encoded little-endian with no compression, regardless of
 // architecture. All changes must be kept in sync with:
 //   platform/tools/gpu/gapii/cc/connection_header.h
 
-func sendHeader(out io.Writer, options Options, gvrHandle uint64, libInterceptorPath string) error {
+func sendHeader(out io.Writer, options Options) error {
 	const maxPath = 512
```
pmuetschard: remove this constant as well

gapii/cc/connection_header.cpp
chrisforbes: Fixed.

gapii/client/header.go
chrisforbes: Done.

PR #163, opened 2020-03-20
Remove the now unused GLES->GL shader conversion code.

pmuetschard:
This was used for desktop GLES replay.


chrisforbes:
Is any of this still going to be useful for showing GLSL or HLSL source perspective of our SPIRV shaders? I haven't looked at that in any detail yet.

pmuetschard:
No, that stuff I kept. This was just for translating GLES shaders to desktop GL, so we could do desktop replay of GLES apps.

PR #164, opened 2020-03-23
New AGI documentation and Landing Page.

pmuetschard:


about.md
```diff
@@ -1,37 +1,12 @@
...
-permalink: /about/
+permalink: /docs/
 ---
 
-AGI is a developer tool for recording and inspecting calls made by an application to the graphics driver.
+<img class="framed" src="https://raw.githubusercontent.com/google/agi/master/tools/logo/logo.svg?sanitize=true" width="256" height="256"/>
 
-It is open-source: [{{site.project-url}}]({{site.project-url}})
+Android GPU Inspector is an [open source]({{site.project-url}}) GPU profiler and debugger for Android applications. <br>
+Currently, only Vulkan applications are supported. Our vision is to help you profile and debug graphics on Android, regardless of device.
```
hevrard: This seems to say "we may support other things than Vulkan", and "it works on any device", which does not really match our current stance. Maybe rephrase to just say "we support vulkan", and either do not mention device, or explain that only a few devices are planned to be actively supported?

@pau-baiget what do you think?

PR #165, opened 2020-03-23
Update tutorial for system profile

rschiu:


pmuetschard:
We should use Windows for screenshots.

rschiu:
Updated with Windows screenshots

PR #169, opened 2020-03-24
Reserve mapped memory at state initialization

chrisforbes:
Previously, if device memory was mapped prior to MEC start, the
allocator was not informed that the memory range was in use. This
led to other memory allocations and/or mappings being placed over
the top.

This would lead to quiet memory corruption in the common case,
or if we got unlucky and placed another memory mapping over the same
range, bizarre explosions when eventually trying to unmap & delete
pre-MEC-start device memory object.

Fixes Swiftshader 'UnmapMemory' crash on Swarming.

Bug: b/151297033

### DIFF
--- a/gapis/api/vulkan/vulkan.go
+++ b/gapis/api/vulkan/vulkan.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/google/gapid/core/app/status"
 	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/core/math/interval"
 	"github.com/google/gapid/gapis/api"
 	"github.com/google/gapid/gapis/api/sync"
 	"github.com/google/gapid/gapis/api/transform"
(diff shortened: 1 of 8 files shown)


gapis/api/vulkan/vulkan.go
```diff
@@ -66,8 +67,16 @@ func (*State) Root(ctx context.Context, p *path.State, r *path.ResolveConfig) (p
...
 // SetupInitialState recreates the command lamdas from the state block.
 // These are not encoded so we have to set them up here.
-func (s *State) SetupInitialState(ctx context.Context) {
+func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
 	s.customState.init(s)
+
+	// Reserve memory for mapped ranges
+	for _, dm := range s.DeviceMemories().All() {
+		if uint64(dm.MappedLocation()) != uint64(0) {
+			state.ReserveMemory([]interval.U64Range{
```
AWoloszyn: Not sure if its worth doing, but we could raise the interval array out of the loop, so we only have to allocate / expand internal range-list once.

chrisforbes: I don't expect there to be many. Can streamline it when we see an issue.

PR #173, opened 2020-03-24
Update (almost) all our dependencies.

pmuetschard:
Updates all our 3rd party dependencies except for breakpad, llvm, and SWT & JFace.


chrisforbes:
Some widget sizes are broken with the switch to GTK3, see ends of checkbox labels are clipped off:
Screenshot from 2020-03-24 15-57-28

chrisforbes:
Vertical alignment of "Regex" checkboxes with filter textboxes is broken with switch to GTK3:
Screenshot from 2020-03-24 16-01-13

pmuetschard:
I've removed the SWT & JFace update commits from the PR, since they have a bit too much effect.

core/fault/stacktrace/stacktrace_test.go
```diff
@@ -82,7 +82,7 @@ var (
 			{
 				"⇒ core/fault/stacktrace/stacktrace_test.go@36:nested1",
 				"⇒ core/fault/stacktrace/stacktrace_test.go@34:nested3",
-				"⇒ core/fault/stacktrace/stacktrace_test.go@34:nested3",
+				"⇒ core/fault/stacktrace/stacktrace_test.go@36:nested1",
```
chrisforbes: This seems dubious?

gapic/src/platform/linux/com/google/gapid/glcanvas/GlCanvas.java
```diff
@@ -48,7 +48,7 @@
   static {
     boolean loaded = false;
     try {
-      Library.loadSystem("linux_glcanvas");
+      Library.loadSystem("", "linux_glcanvas");
```
hevrard: I don't understand that change? The org.lwjgl.system.Library documentation does not seem to have a loadSystem() function that could take two strings?

https://javadoc.lwjgl.org/org/lwjgl/system/Library.html

pmuetschard: https://github.com/LWJGL/lwjgl3/blob/master/modules/lwjgl/core/src/main/java/org/lwjgl/system/Library.java#L61

pmuetschard:
> Also, would it be possible to update grpc to be >= 1.24

No, please see the comment in the commit.

core/fault/stacktrace/stacktrace_test.go
pmuetschard: Done. Figured out how to make it stable.

PR #174, opened 2020-03-25
Trivial tidy-ups

chrisforbes:


gapic/src/main/com/google/gapid/models/Memory.java
```diff
@@ -725,7 +724,7 @@ private static void appendPointedNodes(StructNode root, Map<Long, StructNode> no
       if (root.getTypeCase() == TyCase.POINTER) {
         long pointedAddress = root.getValue().getPointer().getAddress();
         if (nodes.containsKey(pointedAddress)) {
-          root.getChildren().add(nodes.get(pointedAddress));
+          root.getChildren().add(removeExtraLayers(nodes.get(pointedAddress)));
```
stellama0208: Ah, I was surprised that this line change will make a difference, because `removeExtraLayers` is already done for all the nodes previously and a second run shouldn't make any changes. 

So I looked at my code again and debugged it a little, finally found actually there's a small bug in `removeExtraLayers ` method.

I think an actual fix would be changing this line https://github.com/chrisforbes/agi/blob/7ceaa193ad3cbd339ca45a788a0923dac4cedd2d/gapic/src/main/com/google/gapid/models/Memory.java#L686 to `root = removeExtraLayers(root.children.get(0));`. Sorry I forgot to call the recursive method itself after updating root in this method, and so comes the bug.

If the bug fix is made in method `removeExtraLayers` as above, this line change will no longer be needed, you would get the same result :) And I think this would be a better and cleaner way to solve it.

gapic/src/main/com/google/gapid/models/Memory.java
```diff
@@ -685,10 +685,9 @@ private static StructNode removeExtraLayers(StructNode root) {
           && root.children.get(0).hasChildren()) {
         root = root.children.get(0);
       }
-      // Remove the inner layer for PSEUDONYM type. E.g. {[()]} -> {()}.
-      if (tyCase == TyCase.PSEUDONYM && root.hasChildren() && root.children.size() == 1
-          && root.children.get(0).hasChildren()) {
-        root.children = root.children.get(0).children;
+      // Treat PSEUDONYM types as leaves, and remove any children they may have had.
+      if (tyCase == TyCase.PSEUDONYM) {
+	root.children.clear();
```
stellama0208: Sounds good to treat them as leaves! Just a small nit for indenting here.

gapic/src/main/com/google/gapid/models/Memory.java
chrisforbes: Yes, I like that better. Thanks!

gapic/src/main/com/google/gapid/models/Memory.java
chrisforbes: Misconfigured editor, I usually don't touch the Java... thanks

PR #175, opened 2020-03-25
Implement linking to Vulkan objects in the struct UI

chrisforbes:
- Have the server produce paths when boxing linkable handle types.
  This requires a bit of ugliness to recover the actual type of the
  handle and call through its Link() implementation if present.

- In the client, convert the value column of the struct UI to use
  custom drawing. Move the existing ad-hoc linking for "large arrays" to
  be linked from the value column, and add link following for
  server-provided links as well.

Bug: b/143594054

### DIFF
--- a/gapic/src/main/com/google/gapid/views/MemoryView.java
+++ b/gapic/src/main/com/google/gapid/views/MemoryView.java
@@ -81,11 +81,15 @@ import org.eclipse.jface.viewers.ITreeContentProvider;
 import org.eclipse.jface.viewers.ITreeViewerListener;
 import org.eclipse.jface.viewers.LabelProvider;
 import org.eclipse.jface.viewers.StructuredSelection;
+import org.eclipse.jface.viewers.StyledCellLabelProvider;
 import org.eclipse.jface.viewers.TreeExpansionEvent;
 import org.eclipse.jface.viewers.TreeViewer;
+import org.eclipse.jface.viewers.TreeViewerColumn;
 import org.eclipse.jface.viewers.Viewer;
+import org.eclipse.jface.viewers.ViewerCell;
 import org.eclipse.jface.viewers.ViewerComparator;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.StyleRange;
 import org.eclipse.swt.events.MouseEvent;
 import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.graphics.Color;
@@ -807,11 +811,6 @@ public class MemoryView extends Composite
             }
           }
           for (TreeItem item : treeViewer.getTree().getItems()) {
-            // Give visual hint to the clickable large array.
-            if (item.getData() instanceof StructNode &&
-                ((StructNode)item.getData()).isLargeArray()) {
-              item.setForeground(0, widgets.theme.memoryLinkColor());
-            }
             // Give visual hint to the elements of level 1.
             item.setBackground(widgets.theme.memoryFirstLevelBackground());
           }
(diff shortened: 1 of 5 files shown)


chrisforbes:
Only the final commit is interesting, other two are groundwork.

chrisforbes:
Fixed crash when the type is linkable but we fail to create a link (either an invalid handle, or VK_NULL_HANDLE)

gapic/src/main/com/google/gapid/views/MemoryView.java
```diff
@@ -887,16 +893,32 @@ public void treeCollapsed(TreeExpansionEvent event) {
 
       createTreeColumn(tree, "Type", e -> ((StructNode)e).getTypeFormatted());
       createTreeColumn(tree, "Name", e -> ((StructNode)e).getStructName());
-      createTreeColumn(tree, "Value", e -> ((StructNode) e).getValueFormatted());
+      TreeViewerColumn valueColumn = createTreeColumn(tree, "Value");//, e -> ((StructNode) e).getValueFormatted());
```
chrisforbes: I should remove the commented crud off the end here

chrisforbes: Fixed.

gapis/service/memory_box/box.go
```diff
@@ -248,15 +255,44 @@ func Box(ctx context.Context, d *memory.Decoder, t *types.Type) (*Value, error)
...
+				switch vv := b.Val.(*Value_Pod).Pod.Val.(type) {
+				case (*pod.Value_Uint64):
+					v = reflect.ValueOf(vv.Uint64)
+				case (*pod.Value_Uint32):
+					v = reflect.ValueOf(vv.Uint32)
+				default:
+					return nil, log.Errf(ctx, nil, "Cannot handle pseudonym underlying type")
+				}
+
+				linker := v.Convert(reflType).Interface().(path.Linker)
```
pmuetschard: I think a slightly less convoluted way to do this is:
```
v := reflect.New(reflType)
switch vv := b.Val.(*Value_Pod).Pod.Val.(type) {
  case (*pod.Value_Uint64):
    v.Elem().SetInt(vv.Uint64)
  ...
}
if res, err := v.Interface().(path.Linker).Link(...); err == nil {
}
```
... at least I think that should work...

gapic/src/main/com/google/gapid/views/MemoryView.java
```diff
@@ -868,17 +867,24 @@ public void treeCollapsed(TreeExpansionEvent event) {
...
           StructNode node = ((StructNode)tree.getTree().getItem(new Point(e.x, e.y)).getData());
-          long address = node.getRootAddress();
-          blockPanel.goToObservation(address);
-          folder.setSelection(0);
+          if (node.isLargeArray()) {
+            // Large array links are synthetic on the client side
+            long address = node.getRootAddress();
+            blockPanel.goToObservation(address);
+            folder.setSelection(0);
+          }
```
pmuetschard: nit: extra ws

gapis/service/memory_box/box.go
chrisforbes: Thanks, done. I was initially not thrilled about poking values into the guts like this, but it does simplify a bit.

gapic/src/main/com/google/gapid/views/MemoryView.java
chrisforbes: Fixed

gapis/service/memory_box/box.go
```diff
@@ -248,15 +255,41 @@ func Box(ctx context.Context, d *memory.Decoder, t *types.Type) (*Value, error)
...
+
+			if reflType.Implements(tyLinkable) {
+				// reflectively convert the underlying value back to what its
+				// api type should have been, then attempt to link through it.
+				v := reflect.New(reflType)
+				switch vv := b.Val.(*Value_Pod).Pod.Val.(type) {
+				case (*pod.Value_Uint64):
+					v.Elem().SetUint(vv.Uint64)
+				case (*pod.Value_Uint32):
+					v.Elem().SetUint(uint64(vv.Uint32))
```
pmuetschard: we still should fail if in the default case, no?

chrisforbes: In the default case we'll end up leaving v as its zero value and so we won't produce a link. I'm not sure if blowing up the request is preferable.

pmuetschard: True, ok, no default.

PR #180, opened 2020-03-26
Fix struct typed memory observation's appending issue.

stellama0208:
- Bug: http://b/152407425.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Memory.java
+++ b/gapic/src/main/com/google/gapid/models/Memory.java
@@ -25,6 +25,7 @@ import static com.google.gapid.util.Ranges.relative;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.google.common.primitives.UnsignedLongs;
 import com.google.common.util.concurrent.Futures;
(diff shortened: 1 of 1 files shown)


chrisforbes:
Haven't looked at this in detail, but I have verified that it fixes the case in the bug.

gapic/src/main/com/google/gapid/models/Memory.java
```diff
@@ -641,38 +644,21 @@ private boolean mayHaveChildren() {
...
-
-      // Add the remaining unappended nodes to the returning result.
+      // Append pointed nodes to corresponding pointer field if possible.
+      Set<StructNode> appended = Sets.newHashSet();
       for (StructNode node : nodes.values()) {
-        simplified.add(node);
+        appendPointedNodes(node, nodes, appended);
       }
-      return simplified;
+      return nodes.values().stream().filter(n -> !appended.contains(n)).collect(Collectors.toList());
```
chrisforbes: I think this is a pretty big hammer... but OK

pmuetschard: The only real alternative would be to make a copy first and have `appendPointedNodes` remove from the copy.

PR #186, opened 2020-03-30
Update the validation page and make the site work on mobile/small windows.

pmuetschard:
- Updates the device validation page with a bit more info.
- Makes the rest of the site work on mobile/small windows: makes the sidebar "floating" and hidden by default on small screens.

### DIFF
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -19,6 +19,9 @@ toc:
         url: docs/gpu-counters/qualcomm
   - title: System requirements
     url: docs/requirements
+    subnav:
+      - title: Device Validation
+        url: validation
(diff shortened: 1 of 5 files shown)


pmuetschard:
+CC @jakon2048

hevrard:
@pmuetschard maybe it's a good idea to use this PR to also update this link: https://github.com/google/agi/blob/621013ce21b2b3fa13e8936172c7a2dd67c0bd81/gapic/src/main/com/google/gapid/views/TracerDialog.java#L287

pmuetschard:
@hevrard, that link has already been updated in #178.

PR #188, opened 2020-03-31
Show command buffer vulkan handle for subcommand node.

stellama0208:
- Bug: http://b/143734139.

### DIFF
--- a/gapis/api/vulkan/vulkan.go
+++ b/gapis/api/vulkan/vulkan.go
@@ -203,6 +203,7 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
 
 			switch args := GetCommandArgs(ctx, cb.CommandReferences().Get(uint32(i)), st).(type) {
 			case VkCmdExecuteCommandsArgsʳ:
+				d.SubcommandNames.SetValue(nv, "VkCmdExecuteCommandsArgs: ")
 				for j := uint64(0); j < uint64(args.CommandBuffers().Len()); j++ {
 					cbo := st.CommandBuffers().Get(args.CommandBuffers().Get(uint32(j)))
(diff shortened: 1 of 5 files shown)


gapis/api/vulkan/vulkan.go
```diff
@@ -203,6 +203,7 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
 
 			switch args := GetCommandArgs(ctx, cb.CommandReferences().Get(uint32(i)), st).(type) {
 			case VkCmdExecuteCommandsArgsʳ:
+				d.SubcommandNames.SetValue(nv, "VkCmdExecuteCommandsArgs: ")
```
chrisforbes: I like all of this change except this bit here -- can we instead undo whatever is clobbering the original label of a vkCmdExecuteCommands call?

yalcinmelihyasin: Does original label gives any information about it belongs to a vkCmdExecuteCommands?

chrisforbes: If we undo all the label mangling for this command then we should get a label + params etc much like any other command within a command buffer. Elsewhere we're throwing that away.

stellama0208: Thanks for bringing up this naming issue. The bug is now fixed in a new commit.
Since this is a generated subCommand node, the original label is a default group name which is "SubGroup". 
To get the information we are really interested in -- "vkCmdExecuteCommands", at last I set the group name to an empty string here. Because In the UI, when a command node can be interpreted either as a real command or a group node, it has a logic to allow the group name to override. So that only if we set the group name to an empty string will the UI show the original command.

stellama0208:
The comments above are addressed, feel free to take another look!

PR #191, opened 2020-03-31
Framebuffer Attachment Server Changes

hliatis:
Bug: b/152519982


gapis/api/service.proto
```diff
@@ -89,6 +89,22 @@ enum FramebufferAttachment {
   Color3 = 5;
 }
 
+enum FramebufferAttachmentType {
+  DepthAttachment = 0;
+  ColorAttachment = 1;
+}
```
pmuetschard: This belongs into `api/service.proto`

gapis/api/service.proto
```diff
@@ -89,6 +89,22 @@ enum FramebufferAttachment {
...
+
+message FramebufferAttachments {
+  repeated FramebufferAttachmentVulkan attachments = 1;
+}
+
+message FramebufferAttachmentVulkan {
```
pmuetschard: no Vulkan.

gapis/service/path/path.proto
```diff
@@ -218,6 +219,73 @@ message FramebufferObservation {
...
+  // than this.
+  uint32 max_width = 1;
+  // The desired minimum height of the image. The returned image may be larger
+  // than this.
+  uint32 max_height = 2;
+  // The draw mode to use when rendering.
+  DrawMode draw_mode = 3;
+}
+
+message ReplaySettings {
```
pmuetschard: Combine this with RenderSettings and drop the device.

gapis/api/service.proto
hliatis: This is api/service.proto

gapis/api/service.proto
hliatis: This is partially why it's WIP-do-not-review. This was a temp name

gapis/service/path/path.go
```diff
@@ -630,6 +636,10 @@ func (n *Command) ResourcesAfter(ids []*ID) *MultiResourceData {
 	}
 }
 
+func (n *Command) FramebufferAttachmentsAfter(ids []*ID) *FramebufferAttachments {
```
pmuetschard, line 639: what are the ids for?

hliatis: Holdover from something I was trying a while back. To be honest, this function isn't necessary, and I am thinking of deleting it.

gapis/api/service.proto
```diff
@@ -80,13 +80,9 @@ enum ResourceType {
...
-enum FramebufferAttachment {
-  Depth = 0;
-  Stencil = 1;
-  Color0 = 2;
-  Color1 = 3;
-  Color2 = 4;
-  Color3 = 5;
+enum FramebufferAttachmentType {
+  DepthAttachment = 0;
+  ColorAttachment = 1;
```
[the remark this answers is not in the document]
hliatis: Why not just `Color` and `Depth` like in the old enum? I only end them with `Attachment` because they were colliding with the now deleted old enum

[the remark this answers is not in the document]
hliatis: Also, `Attachment` fits with the fields in the subpass description struct

test/integration/service/service_test.go
```diff
@@ -161,24 +163,6 @@ func TestGetDevicesForReplay(t *testing.T) {
 	assert.For(ctx, "got").ThatSlice(got).IsNotEmpty()
 }
 
-func TestGetFramebufferAttachment(t *testing.T) {
```
chrisforbes, line 164: We're losing test coverage here -- should we have a test that covers the new mechanism?

gapis/resolve/framebuffer_changes.go
```diff
@@ -71,27 +72,45 @@ func (r *FramebufferChangesResolvable) Resolve(ctx context.Context) (interface{}
...
+		attachments: make([]framebufferAttachmentChanges, 0),
 	}
 
 	postCmdAndSubCmd := func(s *api.GlobalState, subcommandIndex api.SubCmdIdx, cmd api.Cmd) {
 		api := cmd.API()
+		count, _ := api.GetFramebufferAttachmentCount(ctx, s)
 		idx := append([]uint64(nil), subcommandIndex...)
-		for _, att := range allFramebufferAttachments {
+		for i := uint32(0); i < count; i++ {
+			//att := allFramebufferAttachments[i]
```
chrisforbes, line 83: Don't need this now?

test/integration/service/service_test.go
hliatis: I went ahead and added a test to TestGet() in the file

gapis/service/path/path.go
```diff
@@ -72,6 +72,8 @@ func (n *Device) Path() *Any                    { return &Any{Path: &Any_Device{
 func (n *DeviceTraceConfiguration) Path() *Any  { return &Any{Path: &Any_TraceConfig{n}} }
 func (n *Events) Path() *Any                    { return &Any{Path: &Any_Events{n}} }
 func (n *FramebufferObservation) Path() *Any    { return &Any{Path: &Any_FBO{n}} }
+func (n *FramebufferAttachments) Path() *Any    { return &Any{Path: &Any_FramebufferAttachments{n}} }
```
pmuetschard, line 75: also add `Format` functions for these paths.

gapic/src/main/com/google/gapid/views/FramebufferView.java
```diff
@@ -106,6 +121,20 @@ public FramebufferView(Composite parent, Models models, Widgets widgets) {
...
+      models.settings.preferences().getDisableReplayOptimization());
+
+    RENDER_OVERLAY = Paths.renderSettings(MAX_SIZE, MAX_SIZE, Path.DrawMode.WIREFRAME_OVERLAY, 
+      models.settings.preferences().getDisableReplayOptimization());
+    
+    RENDER_WIREFRAME = Paths.renderSettings(MAX_SIZE, MAX_SIZE, Path.DrawMode.WIREFRAME_ALL, 
+      models.settings.preferences().getDisableReplayOptimization());
+    
+    RENDER_OVERDRAW = Paths.renderSettings(MAX_SIZE, MAX_SIZE, Path.DrawMode.OVERDRAW, 
+      models.settings.preferences().getDisableReplayOptimization());
```
pmuetschard, line 134: The settings can change, and if they do, future requests should honor the change. We now need to create them each time we send a request.
I'd suggest to create an enum with a `Path.RenderSettings getRenderSettings(Settings settings)` method.

hliatis: Should we consider doing so for UsageHints as well? I am working on the UI strip, and I needed to add a new one with a different size, but also shouldn't I be using UsageHints.preview = true for it?

gapis/api/api.go
```diff
@@ -50,7 +50,9 @@ type API interface {
 		after []uint64,
 		state *GlobalState,
 		thread uint64,
-		attachment FramebufferAttachment) (info FramebufferAttachmentInfo, err error)
+		attachment uint32) (info FramebufferAttachmentInfo, err error)
+
+	GetFramebufferAttachmentCount(ctx context.Context, state *GlobalState) (uint32, error)
```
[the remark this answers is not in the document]
hliatis: Sure, but I would have to add any errors to the api.FramebufferAttachmentInfo struct, would I not?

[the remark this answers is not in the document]
pmuetschard: Taking a quick glance at the code I see two error types: 1) errors that happen no matter what the attachment is -> can return them in the list call, 2) unsupported attachment or attachment doesn't exist -> well, the list call just wont return these.

Do you see a specific error per attachment that you are concerned about keeping?

gapic/src/main/com/google/gapid/views/FramebufferView.java
pmuetschard: No, the UsageHints are constant. We use preview for thumbnails. The idea behind the usage hints is to allow the server to prioritize replays. If `primary` is true, it's meant to say that the user is blocked on the data, so it should run first. This should be used when requesting the FB contents for the main view, but not for previews/thumbnails. For the attachment selector, you should use preview.

gapis/api/api.go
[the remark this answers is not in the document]
hliatis: The question was more that I had to attach the errors to individual api.FramebufferAttachmentInfo's, but I think I have it handled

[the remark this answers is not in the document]
pmuetschard: If you combine the implementation as well and get rid of the "give me attachment x" function, you'll find that you no longer need the per-attachment error. All the per-attachment errors I see are of the kind "attachment x does not exist". A function that returns all attachments does not need that error. Please simplify the implementation by getting rid of the "get attachment x" functionality and just replace it with  "get all attachments" instead.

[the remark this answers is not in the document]
hliatis: I might still need to do it for the Unknown Format error

hevrard:
@hliatis this broke linux-test (b/154592150), please make sure to have all CI runs green before merging.

PR #192, opened 2020-04-01
Add DeviceDialog to select replay device

hevrard:
When zero or more than one replay devices are select, ask the user to
select a replay device via this dialog.


hevrard:
@pmuetschard @chrisforbes this is a first version, with the following features:
- dialog is shown only when there is zero or more than one replay device available
- if the dialog is already shown a a replay device refresh leads to a single device, we keep the dialog open, such that it does not auto-magically disappear
- it mirrors the tracer dialog style of device selection+refresh+validation, with replay devices instead of capture devices

This leads to a bit of duplicated code from TracerDialog, related to validation, which is a candidate for refactoring. But before this, can you do a first review and let me know whether this looks decent?

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,303 @@
...
+    synchronized (mutex) {
+      if (!dialogExists) {
+        dialog = new SelectReplayDeviceDialog(shell, models, widgets);
+        dialogExists = true;
+      }
+    }
+
+    if (dialog != null) {
+      if (models.devices.getReplayDevices() != null
+          && models.devices.getReplayDevices().size() == 1) {
```
chrisforbes, line 88: The device must also be successfully validated.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    protected void replayDevicesLoaded() {
+
+      noCompatibleDeviceFound.setVisible(
+          models.devices.getReplayDevices() == null || models.devices.getReplayDevices().isEmpty());
+      noCompatibleDeviceFound.requestLayout();
+
+      deviceCombo.setInput(models.devices.getReplayDevices());
+      deviceCombo.setSelection(null);
```
chrisforbes, line 231: Set to the first listed device

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,303 @@
...
+
+    protected void setValidationStatus(DeviceValidationResult result) {
+      if (result.skipped) {
+        validationStatusLoader.updateStatus(true);
+        validationStatusLoader.stopLoading();
+        validationStatusText.setText("Validation skipped.");
+        validationPassed = true;
+      } else {
+        validationStatusLoader.updateStatus(result.passed);
+        validationStatusText.setText("Validation "
```
chrisforbes, line 294: What happens if we switch devices during validation?

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,303 @@
...
+            + (result.passed ? "Passed." : "Failed. " + Messages.VALIDATION_FAILED_LANDING_PAGE));
+        validationStatusLoader.stopLoading();
+        validationPassed = result.passed;
```
chrisforbes, line 297: OK button enabled state needs to update when validation succeeds

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: done

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: done

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: Apparently, it just starts validation for the newly selected device, and all widgets upload. I'm still not very clear on how the UI thread & RPCs may intertwine, do you think there is a risk of having a stale status here? (If so, I guess there is the same issue in the selection of capture device, from where I got that logic).

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: done

hevrard:
@pmuetschard @chrisforbes I fixed a few things and removed the previous ReplayDeviceSelector dropdown. A few question remains:

1. Do we need to make showSelectReplayDeviceDialog() thread-safe? I'm not 100% clear on how UI thread(s) are orchestrated, but I didn't see much thread-safety checks in other parts of the UI code?

2. Right now, one can close the new dialog, and that leads to a GraphicsTraceView with no replay device selected. I first though we should not enable the replay device dialog to be closed, but that leads to a dead-end if no valid replay device is available. So I think upon closure of this window, we should go back to the welcome screen (rather than showing a graphics trace view with no replay devices), how does that sound?

hevrard:
You may notice that the "Capture device" field does not have a device "friendly name", this is because the device info metadata stored in the capture does not set this device.instance.name field properly. The fix is  https://github.com/google/agi/pull/197

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,322 @@
...
+
+  private DeviceDialog() {
+  }
+
+  public static void showSelectReplayDeviceDialog(Shell shell, Models models, Widgets widgets) {
+    SelectReplayDeviceDialog dialog = null;
+
+    // This method is called every time the list replay devices are refreshed,
+    // and the dialog it opens may itself refresh the list of replay device.
+    // Thus, make sure to have only one dialog open at a time.
```
pmuetschard, line 82: This sounds like the controller (controller as in MVC) is in the wrong place. The way the UI has been written is that the views contain their own controller that reacts to the model events, since the controller is usually tiny. So to fix this, I think the dialog should just mirror the way this is done in all the other views:

in `onCaptureLoaded` and `onReplayDevicesLoaded`:
```
if (capture and replay devices are loaded) {
  update our UI (device table)
  if (!devices.hasReplayDevice()) {
    if (there's only one compatible device) {
      select it
    } else {
      make ourselves visible
    }
  }
}
```

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,322 @@
...
+        device = models.devices.getReplayDevices().get(0);
+        DeviceValidationResult result = models.devices.getValidationStatus(
+            new DeviceCaptureInfo(Paths.device(device.getID()), device, null, null));
+        skipDialog = result.passed || result.skipped;
+      }
+
+      if (skipDialog) {
+        models.devices.selectReplayDevice(device);
+      } else {
+        dialog.setBlockOnOpen(true);
```
pmuetschard, line 106: not necessary

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,322 @@
...
+    protected Control createDialogArea(Composite parent) {
+      Composite composite = (Composite) super.createDialogArea(parent);
+
+      if (!models.capture.isGraphics()) {
+        createLabel(composite,
+            "Error: no graphics capture loaded?"
+                + " This dialog should not be reachable without a graphics capture loaded.")
+                    .setForeground(theme.missingInput());
+        return composite;
+      }
```
pmuetschard, line 158: If you want this check, just throw an exception, since this would clearly be a bug.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,322 @@
...
+
+      // Recap capture info
+      createLabel(composite, "Capture name: " + models.capture.getName());
+      createLabel(composite,
+          "Capture device: " + Devices.getLabel(models.capture.getData().capture.getDevice()));
+
+      // Warning when no compatible device found
+      noCompatibleDeviceFound = createLabel(composite, Messages.SELECT_DEVICE_NO_COMPATIBLE_FOUND);
+      noCompatibleDeviceFound.setForeground(theme.missingInput());
```
pmuetschard, line 167: Create your own color in the theme.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,322 @@
...
+
+      // Mirror the device combo from TracerDialog
+      Group mainGroup =
```
pmuetschard, line 170: I don't think we need the group here.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,322 @@
...
+        }
+      });
+
+      replayDevicesLoaded();
+      return composite;
+    }
+
+    @Override
+    protected void createButtonsForButtonBar(Composite parent) {
+      openTrace = createButton(parent, IDialogConstants.OK_ID, IDialogConstants.OK_LABEL, true);
```
pmuetschard, line 224: You don't need to store the button in an instance field. You can always get it back via `getButton(IDialogConstants.OK_ID)`.

gapic/src/main/com/google/gapid/util/Messages.java
```diff
@@ -87,4 +87,10 @@
   public static final String KEYBOARD_MOUSE_HELP_TITLE = "Keyboard/Mouse Shortcut Help";
   public static final String PROFILE_NO_SLICES =
       "GPU Profiling is not supported on this device or for this capture";
+  public static final String SELECT_DEVICE_TITLE =
+      "Select Replay Device";
+  public static final String SELECT_DEVICE_NO_COMPATIBLE_FOUND =
+      "No compatible replay device found. Plug a compatible device and refresh the device list.";
```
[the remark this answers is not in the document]
hevrard: done

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,322 @@
...
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Link;
+import org.eclipse.swt.widgets.Shell;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
+public class DeviceDialog {
+  protected static final Logger LOG = Logger.getLogger(TracerDialog.class.getName());
```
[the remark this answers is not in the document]
hevrard: done

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: OK I wasn't sure, this check is arguable, I removed it.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: done

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: I'd like to keep it, in order to have the same look and feel between tracing and replay device selection.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: Fixed

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: Does this means we create the dialog once, and then it just reacts to events? If so, when/where should this dialog be created? Maybe associate with GraphicsTraceView?

What you describe applies to all views, but I don't think it applies to dialogs, which are typically summoned from another view by some static showFoobarDialog() methods, which is the way I tried to follow here.

hevrard:
I've updated the controller logic, which feels much more natural now.

PTAL

gapic/src/main/com/google/gapid/GraphicsTraceView.java
```diff
@@ -103,6 +103,13 @@ public void store(TabArea.FolderInfo[] folders) {
     addListener(SWT.Dispose, e -> {
       models.follower.removeListener(this);
     });
+
+//    models.devices.addListener(new Devices.Listener() {
+//      @Override
+//      public void onReplayDevicesLoaded() {
+//        DeviceDialog.showSelectReplayDeviceDialog(getShell(), models, widgets);
+//      }
+//    });
```
stellama0208, line 112: I suppose this part is not intended to be here?

gapic/src/main/com/google/gapid/GraphicsTraceView.java
```diff
@@ -82,8 +83,7 @@ public GraphicsTraceView(Composite parent, Models models, Widgets widgets) {
 
     setLayout(new GridLayout(1, false));
 
-    new ReplayDeviceSelector(this, models)
-        .setLayoutData(new GridData(SWT.FILL, SWT.TOP, true, false));
+    deviceDialog = new DeviceDialog(this, getShell(), models, widgets);
```
stellama0208, line 86: This line will leave a blank space above the main tabs, because there's a concrete composite `deviceDialog` created under `this` composite -- which is `graphicsTraceView `. 
Had some test but haven't found a very good way to avoid it. Change this line to `deviceDialog = new DeviceDialog(tabs, getShell(), models, widgets);` and move it after `tabs` creation can solve the blank area issue, but there might be better solution.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: fixed.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,322 @@
...
+
+  private static Object mutex = new Object();
+  private static boolean dialogExists = false;
```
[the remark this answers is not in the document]
hevrard: removed

gapic/src/main/com/google/gapid/GraphicsTraceView.java
hevrard: Indeed! removed.

gapic/src/main/com/google/gapid/GraphicsTraceView.java
hevrard: I've changed for DeviceDialog to not extend Composite anymore.

We somehow need a widget to call `scheduleIfNotDisposed(someWidget, ...)`, I now use DeviceDialog parent Composite (in practice, GraphicsTraceView) as a widget here. The idea is that as long as the GraphicsTraceView is up, we can open the dialog (if need be).

I'm not sure how clean that is, but it seems to work at least.

@pmuetschard what do you think?

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,334 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
+/**
+ * View responsible to show a replay device selection dialog when need be.
+ */
+public class DeviceDialog implements Devices.Listener, Capture.Listener {
+  protected static final Logger LOG = Logger.getLogger(DeviceDialog.class.getName());
+
+  private final Shell shell;
```
pmuetschard, line 77: No need for this, just use `parent.getShell()` to get the shell.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -0,0 +1,334 @@
...
+        dialog = new SelectReplayDeviceDialog(shell, models, widgets);
+        scheduleIfNotDisposed(parent, () -> dialog.open());
+      }
+    }
+  }
+
+  /**
+   * Dialog to select a replay device.
+   */
+  static class SelectReplayDeviceDialog extends DialogBase {
```
pmuetschard, line 148: private.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: done

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: done

PR #194, opened 2020-04-01
Make validation more robust on Adreno.

lpy:
The average of 'Textures / Fragment' should be 1 for the vulkan sample while
each 50 ms sample might vary a bit larger than 0.1 from 1.

Minor: Move some code around to avoid duplication.

Bug: b/151801290

### DIFF
--- a/gapis/trace/android/validate/validate.go
+++ b/gapis/trace/android/validate/validate.go
@@ -79,62 +79,62 @@ func IsNumber(column *perfetto_service.QueryResult_ColumnValues, columnType perf
 	return true
 }
 
-// CheckLargerThanZero is a checker that checks that the values are all greater than zero.
-func CheckLargerThanZero(column *perfetto_service.QueryResult_ColumnValues, columnType perfetto_service.QueryResult_ColumnDesc_Type) bool {
-	longValues := column.GetLongValues()
-	doubleValues := column.GetDoubleValues()
-	for i := 0; i < sampleCounter; i++ {
-		if columnType == perfetto_service.QueryResult_ColumnDesc_LONG {
-			if longValues[i] <= 0 {
-				return false
-			}
-		} else if columnType == perfetto_service.QueryResult_ColumnDesc_DOUBLE {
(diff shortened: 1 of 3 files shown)


gapis/trace/android/validate/validate.go
```diff
@@ -117,26 +98,54 @@ func CheckEqualTo(num float64) Checker {
...
 	}
 }
 
+// CheckLargerThanZero is a checker that checks that the values are all greater than zero.
+func CheckLargerThanZero() Checker {
+	return ForeachValue(func(v float64) bool {
+		if v <= 0.0 {
+			return false
+		}
+		return true
```
pmuetschard, line 124: `return v > 0.0`

gapis/trace/android/validate/validate.go
```diff
@@ -117,26 +98,54 @@ func CheckEqualTo(num float64) Checker {
...
+	})
+}
+
+// CheckEqualTo returns a checker that checks that all returned value equal the given value.
+func CheckEqualTo(num float64) Checker {
+	return ForeachValue(func (v float64) bool {
+		if v != num {
+			return false
+		}
+		return true
```
pmuetschard, line 134: `return v == num`

gapis/trace/android/validate/validate.go
```diff
@@ -117,26 +98,54 @@ func CheckEqualTo(num float64) Checker {
...
+}
+
+// CheckAverageApproximateTo checks whether the average of the values is within
+// a margin of the given value.
+func CheckAverageApproximateTo(num, err float64) Checker {
+	return Average(func(v float64) bool {
+		if math.Abs(num - v) > err {
+			return false
+		}
+		return true
```
pmuetschard, line 145: err -> margin (err should be used only for errors) and
`return mat.Abs(num - v) <= margin`

lpy:
thanks for the feedback, ptal

hevrard:
(just need to fix presubmit)

PR #197, opened 2020-04-06
Fix device instance name on Android

hevrard:
Return the hardware name for the instance name. This mirrors:
https://github.com/google/agi/blob/77648d1557836fcf549c4238e65fb1e35c1d296f/core/os/android/adb/device.go#L245

This also removes the "serial" field, as reading serial number via
ro.serialno property is prevented by SELinux, see:
https://cs.android.com/android/platform/superproject/+/master:system/sepolicy/public/domain.te;l=541;drc=19a5cc2bab59a4eea424ab38a6f1aa67b65860f1

Bug: b/153327038

### DIFF
--- a/core/os/device/deviceinfo/cc/android/query.cpp
+++ b/core/os/device/deviceinfo/cc/android/query.cpp
@@ -136,7 +136,6 @@ struct Context {
   EGLContext mContext;
   int mNumCores;
   std::string mHost;
-  std::string mSerial;
   std::string mHardware;
   std::string mOSName;
   std::string mOSBuild;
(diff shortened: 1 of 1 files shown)


core/os/device/deviceinfo/cc/android/query.cpp
```diff
@@ -501,7 +500,7 @@ const char* gpuName() { return ""; }
 
 const char* gpuVendor() { return ""; }
 
-const char* instanceName() { return gContext.mSerial.c_str(); }
+const char* instanceName() { return gContext.mHardware.c_str(); }
```
pmuetschard, line 503: We already return the hardware name below. Why would we have two functions returning the same thing? Can you please fix what calls these functions instead?

hevrard: It looks like we have a duplicate in device.instance: device.instance.name == device.instance.configuration.hardware.name , at least in android. See:

https://github.com/google/agi/blob/77648d1557836fcf549c4238e65fb1e35c1d296f/core/os/android/adb/device.go#L245

Should we get rid of the device.instance.name field entirely?

pmuetschard: Ok, sorry, I was confusing things. It is fine to return the hardware name here.

PR #199, opened 2020-04-07
Framelifecycle polishes (Updated Perfetto)

silence-do-good:
Now that the frame events are separated from the gpu_slice table, this PR addresses that change and the following:
- Show statistics of all the frames/buffers that got composited into a displayed frame, when it is selected
- Highlight(draw in color) all related events of the frame when an instant event is selected, greying out other events
- Highlight(draw in color) all related events of all the frames within a displayed frame slice when it is selected, greying out other events
- Minor: Make sure that displayed frame track always goes on top of all other tracks under the SF events group
- Minor: Remove trackId from multiselection view and show layer name instead

Bug: 152533807, 150172851

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
@@ -400,14 +400,14 @@ public class FrameEventsTrack extends Track.WithQueryEngine<FrameEventsTrack.Dat
   public static class FrameSelection {
     public static FrameSelection EMPTY = new FrameSelection();
     // key = concat(layerName, '_', frameNumber)
-    private List<String> keys;
+    private Set<String> keys;
 
     public FrameSelection() {
-      keys = Lists.newArrayList();
+      keys = Sets.newHashSet();
     }
 
     public FrameSelection(long[] f, String[] l) {
-      keys = Lists.newArrayList();
+      keys = Sets.newHashSet();
       for (int i = 0; i < l.length; i++) {
         keys.add(l[i] + "_" + f[i]);
       }
--- a/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
@@ -25,22 +25,26 @@ import static com.google.gapid.perfetto.models.QueryEngine.expectOneRow;
 import static com.google.gapid.util.MoreFutures.transform;
 import static com.google.gapid.util.MoreFutures.transformAsync;
 import static java.lang.String.format;
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.common.primitives.Longs;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.gapid.perfetto.TimeSpan;
 import com.google.gapid.perfetto.views.FrameEventsMultiSelectionView;
 import com.google.gapid.perfetto.views.FrameEventsSelectionView;
 import com.google.gapid.perfetto.views.State;
 
-import java.util.Arrays;
 import org.eclipse.swt.widgets.Composite;
 
+import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.function.Consumer;
@@ -51,11 +55,12 @@ import java.util.function.Consumer;
 // TODO: dedupe code with SliceTrack.
 public class FrameEventsTrack extends Track.WithQueryEngine<FrameEventsTrack.Data>{
   private static final String BASE_COLUMNS =
-      "id, ts, dur, category, name, depth, stack_id, parent_stack_id, arg_set_id";
+      "id, ts, dur, name, depth, stack_id, parent_stack_id, arg_set_id, " +
+      "frame_numbers, layer_names";
   private static final String SLICES_VIEW =
-      "select " + BASE_COLUMNS + " from gpu_slice where track_id = %d";
+      "select " + BASE_COLUMNS + " from frame_slice where track_id = %d";
   private static final String SLICE_SQL =
-      "select " + BASE_COLUMNS + " from gpu_slice where id = %d";
+      "select " + BASE_COLUMNS + " from frame_slice where id = %d";
   private static final String SLICES_SQL =
        "select " + BASE_COLUMNS + " from %s " +
        "where ts >= %d - dur and ts <= %d order by ts";
@@ -68,8 +73,15 @@ public class FrameEventsTrack extends Track.WithQueryEngine<FrameEventsTrack.Dat
       "where ts < %d and ts + dur >= %d and depth >= %d and depth <= %d";
   private static final String RANGE_FOR_IDS_SQL =
       "select " + BASE_COLUMNS + " from %s where id in (%s)";
+  private static final String STAT_TABLE_SQL =
+      "select frame_numbers, layer_names, queue_to_acquire_time, " +
+      "acquire_to_latch_time, latch_to_present_time " +
+      "from frame_slice left join frame_stats " +
+      "on frame_slice.id = frame_stats.slice_id " +
+      "where frame_stats.slice_id = %d";
 
   private static final long SIGNAL_MARGIN_NS = 10000;
+  private static final long FRAMELIFECYCLE_QUANTIZE_CUTOFF = MICROSECONDS.toNanos(500);
 
   private final long trackId;
 
@@ -98,26 +110,30 @@ public class FrameEventsTrack extends Track.WithQueryEngine<FrameEventsTrack.Dat
 
   @Override
   public ListenableFuture<Data> computeData(DataRequest req) {
-    Window window = Window.compute(req, 5);
+    Window window = Window.compute(req, 5, FRAMELIFECYCLE_QUANTIZE_CUTOFF);
     return transformAsync(window.update(qe, tableName("window")),
         $ -> window.quantized ? computeSummary(req, window) : computeSlices(req));
   }
 
   private ListenableFuture<Data> computeSlices(DataRequest req) {
     return transformAsync(qe.query(slicesSql(req)), res ->
-    transform(qe.getAllArgs(res.stream().mapToLong(r -> r.getLong(8))), args -> {
+    transform(qe.getAllArgs(res.stream().mapToLong(r -> r.getLong(7))), args -> {
       int rows = res.getNumRows();
-      Data data = new Data(req, new long[rows], new long[rows], new long[rows], new int[rows],
-          new String[rows], new String[rows], new ArgSet[rows]);
+      Data data = new Data(req, new long[rows], new long[rows], new long[rows],
+          new String[rows], new long[rows][], new String[rows][], new ArgSet[rows]);
       res.forEachRow((i, row) -> {
         long start = row.getLong(1);
         data.ids[i] = row.getLong(0);
         data.starts[i] = start;
         data.ends[i] = start + row.getLong(2);
-        data.categories[i] = row.getString(3);
-        data.titles[i] = row.getString(4);
-        data.depths[i] = row.getInt(5);
-        data.args[i] = args.getOrDefault(row.getLong(8), ArgSet.EMPTY);
+        data.titles[i] = row.getString(3);
+        data.args[i] = args.getOrDefault(row.getLong(7), ArgSet.EMPTY);
+        long[] frameNumbers = Arrays.stream(row.getString(8).split(", "))
+            .mapToLong(Long::parseLong)
+            .toArray();
+        String[] layerNames = row.getString(9).split(", ");
+        data.frameNumbers[i] = frameNumbers;
+        data.layerNames[i] = layerNames;
       });
       return data;
     }));
@@ -147,11 +163,31 @@ public class FrameEventsTrack extends Track.WithQueryEngine<FrameEventsTrack.Dat
 
   public ListenableFuture<Slice> getSlice(long id) {
     return transformAsync(expectOneRow(qe.query(sliceSql(id))), r ->
-        transform(qe.getArgs(r.getLong(8)), args -> buildSlice(r, args)));
+        transformAsync(qe.getArgs(r.getLong(7)), args ->
+        transform(getStats(id, r.getLong(2)), stats -> buildSlice(r, args, stats))));
   }
 
-  protected Slice buildSlice(QueryEngine.Row row, ArgSet args) {
-    return new Slice(row, args, trackId);
+  private ListenableFuture<Map<String, FrameStats>> getStats(long id, long dur) {
+    if (dur == 0) { // No stats for instant events
+      return Futures.immediateFuture(null);
+    }
+    return transform(qe.query(statSql(id)), result -> {
+      Map<String, FrameStats> stats = Maps.newHashMap();
+      result.forEachRow((i, row) -> {
+        stats.put(row.getString(1).split(", ")[i],
+            new FrameStats(Long.parseLong(row.getString(0).split(", ")[i]),
+                row.getLong(2), row.getLong(3), row.getLong(4)));
+      });
+      return stats;
+    });
+  }
+
+  private static String statSql(long sliceId) {
+    return format(STAT_TABLE_SQL, sliceId);
+  }
+
+  protected Slice buildSlice(QueryEngine.Row row, ArgSet args, Map<String, FrameStats> frameStats) {
+    return new Slice(row, args, frameStats);
   }
 
   private static String sliceSql(long id) {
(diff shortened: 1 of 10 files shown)


gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -333,6 +388,62 @@ public void getRange(Consumer<TimeSpan> span) {
...
+    public final List<String> layerNames;
+    public static FrameSelection EMPTY = new FrameSelection();
+
+    public FrameSelection() {
+      frameNumbers = Lists.newArrayList();
+      layerNames = Lists.newArrayList();
+    }
+
+    public FrameSelection(long[] frameNumbers, String[] layerNames) {
+      this.frameNumbers = Lists.newArrayList(Longs.asList(frameNumbers));
```
rschiu, line 410: Just Lists.newArrayList?

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -333,6 +388,62 @@ public void getRange(Consumer<TimeSpan> span) {
...
+      return false;
+    }
+
+    public void combine(FrameSelection other) {
+      frameNumbers.addAll(other.frameNumbers);
+      layerNames.addAll(other.layerNames);
+    }
+
+    public boolean isEmpty() {
+      return frameNumbers.size() == 0 ? true : false;
```
rschiu, line 430: return frameNumbers.isEmpty();

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -333,6 +388,62 @@ public void getRange(Consumer<TimeSpan> span) {
...
+    }
+
+    public FrameSelection(long[] frameNumbers, String[] layerNames) {
+      this.frameNumbers = Lists.newArrayList(Longs.asList(frameNumbers));
+      this.layerNames = Lists.newArrayList(layerNames);
+    }
+
+    public boolean contains(long[] f, String[] l) {
+      for (int i = 0; i < f.length; i++) {
+        int idx = frameNumbers.indexOf(f[i]);
```
rschiu, line 416: This doesn't work if FrameSelection has multiple layers with the same frame number.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSummaryPanel.java
```diff
@@ -403,6 +408,16 @@ public void computeSelection(CombiningBuilder builder, Area area, TimeSpan ts) {
     }
   }
 
+  private static FrameSelection getSelectedFrames(State state) {
+    Selection selection = state.getSelection(Selection.Kind.FrameEvents);
+    if (selection instanceof FrameEventsTrack.Slice) {
+      return ((FrameEventsTrack.Slice) selection).getSelection();
+    } else if (selection instanceof FrameEventsTrack.Slices) {
```
rschiu, line 415: Can you combine FrameEventsTrack.Slice with FrameEventsTrack.Slices?  Stella did something similar.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSummaryPanel.java
```diff
@@ -323,14 +329,13 @@ private Hover sliceHover(FrameEventsTrack.Data data, Fonts.TextMeasurer m, doubl
         endts = ts + 12;
         ts -= 12;
       }
-      if (data.depths[i] == depth && x >= ts && x<= endts) {
+      if (x >= ts && x<= endts) {
```
rschiu, line 332: Same question with comment above.  Is this change necessary?

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -160,7 +196,7 @@ private static String sliceSql(long id) {
 
   public ListenableFuture<List<Slice>> getSlices(TimeSpan ts, int minDepth, int maxDepth) {
     return transform(qe.query(sliceRangeSql(ts, minDepth, maxDepth)),
-        res -> res.list(($, row) -> buildSlice(row, ArgSet.EMPTY)));
+        res -> res.list(($, row) -> buildSlice(row, ArgSet.EMPTY, null)));
```
stellama0208, line 199: Nit: personal opinion but I think an empty map might be better. The same for null below.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -205,14 +241,14 @@ public Data(DataRequest request, long bucketSize, String[] concatedIds, long[] n
       this.ids = null;
       this.starts = null;
       this.ends = null;
-      this.depths = null;
```
stellama0208, line 208: The depth is still queried as one of the base columns in the SQL query, and we also still keep the getSlices method inside this file using depth as parameter.. Do we want to remove that as well?

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -333,6 +388,62 @@ public void getRange(Consumer<TimeSpan> span) {
...
+      FrameSelection selection = new FrameSelection();
+      slices.forEach(s -> selection.combine(s.getSelection()));
+      return selection;
+    }
+  }
+
+  public static class FrameSelection {
+    public final List<Long> frameNumbers;
+    public final List<String> layerNames;
+    public static FrameSelection EMPTY = new FrameSelection();
```
stellama0208, line 402: nit: would be better to put it at the first line of the class.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -291,6 +343,9 @@ public void getRange(Consumer<TimeSpan> span) {
       }
     }
 
+    public FrameSelection getSelection() {
+      return new FrameSelection(frameNumbers, layerNames);
```
stellama0208, line 347: This function will be queried every time rendering the track, so it would be good to keep it efficient. It seems that frameNumbers and layerNames are fixed at slice creation time, maybe we can save a fixed FrameSelection at slice creation time as well? So as to avoid creating a new object every time. Or maybe I'm missing something here.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSelectionView.java
```diff
@@ -56,6 +59,34 @@ public FrameEventsSelectionView(Composite parent, State state, FrameEventsTrack.
...
+            withSpans(new GridData(), 2, 1));
+
+        createLabel(stats, "Queue to Acquire: ");
+        createLabel(stats, timeToString(v.queueToAcquireTime));
+
+        createLabel(stats, "Acquire to Latch: ");
+        createLabel(stats, timeToString(v.acquireToLatchTime));
+
+        createLabel(stats, "Latch to Present: ");
+        createLabel(stats, timeToString(v.latchToPresentTime));
```
stellama0208, line 80: Can we show frameNumbers as well here? I think we should show users a little hint why something got grouped together by color. 
From the code I see they are connected by frameNumbers and layerNames, but these things are not shown in UI.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -333,6 +388,62 @@ public void getRange(Consumer<TimeSpan> span) {
...
+      layerNames = Lists.newArrayList();
+    }
+
+    public FrameSelection(long[] frameNumbers, String[] layerNames) {
+      this.frameNumbers = Lists.newArrayList(Longs.asList(frameNumbers));
+      this.layerNames = Lists.newArrayList(layerNames);
+    }
+
+    public boolean contains(long[] f, String[] l) {
+      for (int i = 0; i < f.length; i++) {
```
stellama0208, line 415: Just a note that this function will also be queried every time rendering the track, and will be called for every visible slices. Is there any better key to identify the selection, and make this function run faster? If not then it's fine to keep it this way.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSummaryPanel.java
```diff
@@ -176,10 +176,16 @@ public void renderSlices(RenderContext ctx, FrameEventsTrack.Data data) {
...
+        ctx.setBackgroundColor(color.disabled);
+        if (selectedFrames.contains(data.frameNumbers[i], data.layerNames[i])) {
+          color.applyBase(ctx);
+        }
+      }
 
       if (tEnd - tStart > 3 ) {
         double rectWidth = Math.max(1, state.timeToPx(tEnd) - rectStart);
-        double y = depth * SLICE_HEIGHT;
+        double y = 0;
```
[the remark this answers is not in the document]
silence-do-good: Removed the depth from the query altogether because its just not being used anywhere. And currently, there is no plan of using it in the foreseeable future as well.  Even if we change the instant events to slices, the depth will always be 0. We can always add it back in case it is needed, but its just a waste for now, with so many select queries being sent.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
silence-do-good: The lines below this does that check

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
silence-do-good: My bad. I removed the depth from the query as well, but I think that got lost while cleaning up. I'll remove the depth from the queries altogether.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSummaryPanel.java
silence-do-good: I'll look into what Stella did and make changes if its suitable here.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
silence-do-good: This function will be called only when the event is selected. Storing the FrameSelection within the Slice would be a waste because the slices are created anew for every redraw.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
rschiu: I am thinking of the scenario where the layer name comes at a later index.  indexOf only returns the first one, the function will return false instead of true.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
stellama0208: Hmmm from my understanding, if following the chain of FrameEventSummaryPanel.java -> renderTrack -> renderSlices -> getSelectedFrames -> getSelection -> new FrameSelection, it's called at every track rendering time?

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
silence-do-good: Just checked again, depth is being used in the span_join operator and removing it from the query is going to change a lot of things. I suggest just removing depths from the data for now, since we are not using it.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSummaryPanel.java
stellama0208: To avoid type checking here, it will require `FrameEventsTrack.Slice` and `FrameEventsTrack.Slices` both implement `Selection` framework, and have them both saved under a new category `Kind.FurtherFrame`.
I can see the difficulty of letting them implementing Selection framework here, because the class's contains signature is `contains(long[] f, String[] l)` rather than `contains(long id)`, if we really can't find a long-typed id as a key, I'm fine with it checked in as this for now.
There are now two outliners doing type checking, one her and one in VulkanEvent track. We'd better think about how to add some flexibility into the selection framework to make these two parts more elegant.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
silence-do-good: Leaving it like this for now, based on our discussion.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSelectionView.java
silence-do-good: Done.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
silence-do-good: Got it. Done.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
silence-do-good: We check for null in the selection view and its kinda easier this way. There is a solid difference between an instant event and a displayed frame slice.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSummaryPanel.java
silence-do-good: Discussed in our meeting about this and keeping it as is.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
silence-do-good: Fixed it. Done.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSummaryPanel.java
silence-do-good: As Stella pointed out, grouping them together is going to involve a lot of changes to the Selection framework and I'd like to avoid mixing those things up in the change. Stella and I will look into refactoring the Selection framework further, if we stumble upon this usecase again. Right now, only the Vulkan events and Frame events require this behavior.

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Data about a GPU in the trace.
+ */
+public class FrameInfo {
+  public static final FrameInfo NONE = new FrameInfo(Collections.emptyList());
+
+  private static final String FRAME_SLICE_QUERY =
+      "select t.id, t.name, t.scope, max(depth) + 1 " +
+      "from gpu_track t join frame_slice s on (t.id = s.track_id) " +
```
pmuetschard, line 36: This doesn't look right. Shouldn't this be limited to a different table than gpu_track?

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
```diff
@@ -0,0 +1,103 @@
...
+        transform(qe.queries(FRAME_SLICE_QUERY), frame -> {
+            List<Buffer> buffers = Lists.newArrayList();
+            frame.forEachRow(($, r) -> {
+              buffers.add(new Buffer(r));
+            });
+
+            // Sort buffers by name, the query is sorted by track id for the queues.
+            buffers.sort((b1, b2) -> {
+              // Displayed Frame track should always be at top
+              if (b1.name.equals("Displayed Frame")) {
```
pmuetschard, line 73: Yikes, sounds like the buffer should have some type enum. At least make this a constant.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -180,6 +167,20 @@ private Tracks() {
...
+      String parent = "sf_events";
+      for (FrameInfo.Buffer buffer : data.getFrame().buffers()) {
+        FrameEventsTrack track = FrameEventsTrack.forBuffer(data.qe, buffer);
+        data.tracks.addTrack(parent, track.getId(), buffer.getDisplay(),
+            single(state -> new FrameEventsSummaryPanel(state, buffer, track), true, false));
+      }
+    }
+    return data;
+  }
+
```
pmuetschard, line 183: These tracks should still be in the gpu group, no?

silence-do-good: Moved them out of that group since the tables are now different as well. They will be just below the GPU for now.

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
silence-do-good: Good catch actually. I think moving this out of gpu_track is going to involve a bit of changes in perfetto judging by the places its being touched. I'll create a follow up bug and once that is done, I'll patch this separately? This change has some important UI changes that I want to get in first.

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
silence-do-good: Yeah. There are some potential changes coming along the way to the Displayed Frame track.  I'll refactor this logic depending on the decisions we make. For now, I've changed the literal into a constant.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -333,6 +391,59 @@ public void getRange(Consumer<TimeSpan> span) {
...
+    public FrameSelection(long[] f, String[] l) {
+      keys = Lists.newArrayList();
+      for (int i = 0; i < l.length; i++) {
+        keys.add(l[i] + "_" + f[i]);
+      }
+    }
+
+    public boolean contains(long[] f, String[] l) {
+      for (int i = 0; i < f.length; i++) {
+        if (keys.contains(l[i] + "_" + f[i])) {
```
stellama0208, line 418: It looks like we can change `List<String> keys` to `Set<String> keys` to improve efficiency this line. Or maybe I'm missing sth and there's a reason using list here?

silence-do-good: Done.

PR #202, opened 2020-04-07
Some simple optimizations

chrisforbes:
- Events can be built without mutating. The old GLES api definition had some stateful command flags, but Vulkan does not. The other user was extensions, of which we have none after removing the GVR & GLES support.

- The whole extension mechanism is now unused. Remove the remnants.

- As soon as we start building a batch, kick off precaching of the initial payload. This allows for better scheduling.


gapis/replay/batch.go
```diff
@@ -269,6 +270,27 @@ func (m *manager) execute(
 	}
 	ctx = log.V{"replay target ABI": replayABI}.Bind(ctx)
 
+	// We may or may not actually need the initial payload, but kick it off now
```
pmuetschard, line 273: I think we should skip this if `forceNonSplitReplay` is `true` or `generator` doesn't implement `SplitGenerator`, since then we know we won't need the initial payload.

chrisforbes: That's the nastiest bit of this whole business-- doing as you suggest actually defeats the optimization.

In the "load a trace with the GUI and wait for idle" scenario, the first replay we kick off does not need the initial payload, but we know *some* future replay against the same device certainly will.

We should probably rearrange the replay executor in a larger way so this looks natural. As soon as we've decided to run replays on a particular device, we want to, all in parallel:
- Build all the initial payload segments we're likely to need
- Connect to the GAPIR

chrisforbes: The other reasonable approach here is to completely extract the generate & build activities from the ReplayExecutor. They are slightly device specific, but there's no reason not to overlap them with execution.

chrisforbes: With things how they are currently though, kicking this work at this point is a win whether we use it or not. As mentioned in the commit message:
- If we use it, we managed to overlap it with Connect, which is slow.
- If we don't use it, some later replay will, and we managed to completely hide it in our own execution time.

pmuetschard: is this because the first replay we run is the profiling replay, where we force non-split? We might have to re-think how we prioritize things on start-up.

Either way, i'm fine with this as is.

chrisforbes: Yes -- profiling replay currently forces non-split.

PR #204, opened 2020-04-08
Add Swarming test results as nightly artifacts

hevrard:


yalcinmelihyasin:
A general question: would it be possible/feasible to convert this to Python?

hevrard:
> A general question: would it be possible/feasible to convert this to Python?

Good question! I actually refrained from using Python because we have no Python code in AGI (apart for a small MacOS-build-specific config), and I don't want to add another dependency.

Hence the bash approach, although to be honest I did not think it would turn out to be so convoluted.
I'm keen on getting this in as-is, and eventually re-visit later if need be.

PR #206, opened 2020-04-09
Added how to use dev drivers

jakon2048:
Added how to prepare application for dev drivers in getting started pages

### DIFF
--- a/tutorial.md
+++ b/tutorial.md
@@ -20,6 +20,13 @@ For the duration of the developer preview, instructions on how to set up the bet
   ```
 [debuggable attribute]: https://developer.android.com/guide/topics/manifest/application-element#debug
 
+* To utilize the beta GPU drivers, the following piece of metadata must be included in the `<application>` tag of the APK manifest:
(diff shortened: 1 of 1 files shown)


tutorial.md
```diff
@@ -20,6 +20,13 @@ For the duration of the developer preview, instructions on how to set up the bet
   ```
 [debuggable attribute]: https://developer.android.com/guide/topics/manifest/application-element#debug
 
+* To utilize the beta GPU drivers, the following piece of metadata must be included in the `<application>` tag of the APK manifest:
```
hevrard, line 23: For the sake of consistent naming: I can't remember reading "beta" anywhere before, is this how we call those drivers? I think we should agree/clarify on a proper naming (e.g. "developer/prerelease/beta/...") and stick with it all over the place.

jakon2048: let's call them beta developer drivers externally. "developer drivers" don't mean anything to outsiders, but if we add a beta, I think they will get the idea

PR #207, opened 2020-04-09
Implement draw command grouping.

stellama0208:
- Bug: b/143731054.

### DIFF
--- a/gapis/api/vulkan/vulkan.go
+++ b/gapis/api/vulkan/vulkan.go
@@ -17,6 +17,7 @@ package vulkan
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/google/gapid/core/app/status"
 	"github.com/google/gapid/core/log"
@@ -123,6 +124,7 @@ type MarkerType int
 const (
 	DebugMarker = iota
 	RenderPassMarker
+	DrawGroupMarker
 )
 
 type markerInfo struct {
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/vulkan.go
```diff
@@ -291,7 +307,41 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
 			case VkCmdDebugMarkerEndEXTArgsʳ:
 				popMarker(DebugMarker, uint64(i), nCommands)
 			}
+
+			// Handle draw commands grouping.
+			cmdName := ""
+			if generatingId != nil {
```
chrisforbes, line 313: When this fails (recording was prior to MEC start) we're going to have a problem.

gapis/api/vulkan/vulkan.go
```diff
@@ -291,7 +307,41 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
...
+				groupName := cmdName
+				if strings.HasPrefix(groupName, "vkCmd") { // Remove "vkCmd".
+					groupName = groupName[5:len(groupName)]
+				}
+				popMarkerWithNewGroupName(DrawGroupMarker, uint64(i), groupName)
+				canStartDrawGrouping = true
+			} else if !isStateSettingCmd && !isDrawCmd && !canStartDrawGrouping {
+				// Handle an edge case where a group of state setting commands are
+				// followed by something other than a drawing command.
+				groupName := "State Setting Group"
```
chrisforbes, line 339: If we default the group name to this rather than "Draw Group" then maybe this can be simplified a bit ["falling off the end" and just calling the normal popMarker will do the correct thing then?]

gapis/api/vulkan/vulkan.go
```diff
@@ -291,7 +307,41 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
 			case VkCmdDebugMarkerEndEXTArgsʳ:
 				popMarker(DebugMarker, uint64(i), nCommands)
 			}
+
+			// Handle draw commands grouping.
+			cmdName := ""
+			if generatingId != nil {
+				cmdName = generatingId.CmdName()
+			}
+			isDrawCmd := strings.HasPrefix(cmdName, "vkCmdDraw") || strings.HasPrefix(cmdName, "vkCmdDispatch")
```
chrisforbes, line 316: We should use flags to get these.

chrisforbes: Alternatively, we could fold this logic into the switch above and then we don't care so much about names at all -- we'll look at the command reference types instead

gapis/api/vulkan/vulkan.go
stellama0208: Thanks for pointing it out! The command name is no longer referenced from the generatingId after the new cimmits.

gapis/api/vulkan/vulkan.go
stellama0208: After discussing with Chris offline, we decided the judging logic for isDrawCmd value can be kept the same. 
1. We abandoned the use of api.cmd for safety against MEC case, so no flags here.
2. The switch option will result in enlisting a whole bunch of types, like all the "vkCmdSet*" commands.

stellama0208:
The comments above are addressed in new commits, PTAL!

PR #208, opened 2020-04-10
Cleanup of device info

pmuetschard:
Removes the GLES info from device info as we don't need it and cleans up the code.


core/os/device/deviceinfo/cc/cpu.cpp
```diff
@@ -18,55 +18,81 @@
...
   }
-  return "";
-}
 
-const char* cpuVendor() {
-  static union {
-    uint32_t reg[3];
-    char str[13];
-  };
+  str[12] = 0;
```
hevrard, line 38: Shouldn't this be `str[48] = 0;` ?

core/os/device/deviceinfo/cc/query.cpp
```diff
@@ -72,36 +72,35 @@ void deviceInstanceID(device::Instance* instance) {
...
-  const char* backupName = "";
-  if (query::hasGLorGLES()) {
-    // Instance.Configuration.Drivers.OpenGLDriver
-    auto opengl_driver = new OpenGLDriver();
-    query::glDriver(opengl_driver);
-    drivers->set_allocated_opengl(opengl_driver);
-    backupVendor = opengl_driver->vendor().c_str();
-    backupName = opengl_driver->renderer().c_str();
-  }
+  const char* gpuVendor = "";  // TODO: we no longer have a vlue for this
```
hevrard, line 102: typo: vlue -> value

Is this something we should get via Vulkan now? (doesn't have to be in this PR)

core/os/device/deviceinfo/cc/cpu.cpp
chrisforbes: No -- this is null termination for the old leaf 0 ("GenuineIntel" etc), which is exactly 12 characters long and does not include its own terminator

pmuetschard: Chris is correct, this is due to reusing the same buffer. If the vendor below uses exactly 12 bytes (without a terminating null), we wouldn't want it to also use the bytes from the device name.

I've added a comment.

core/os/device/deviceinfo/cc/query.cpp
pmuetschard: I've pushed another commit with a tiny PCI ID database to lookup the vendor name.

PR #211, opened 2020-04-14
Virt Swapchain: improvements, esp. for standalone

paulthomson:
Some improvements to the Virtual Swapchain layer, especially for standalone use (i.e. outside replay).

* Support Android properties instead of environment variables on Android.
* Add `VIRTUAL_SWAPCHAIN_SURFACE_EXTENT` parameter to set the default surface extent (width and height); without this, apps are likely to request a swapchain with `UINT_MAX` width and height, which will fail.
* `vkQueuePresentKHR`: set `pPresentInfo->pResults` if needed; without this, apps may think `vkQueuePresentKHR` always fails.
* Add some logging on failures; uses `__android_log_print` on Android, and `std::cerr` otherwise.

### DIFF
--- a/core/vulkan/vk_virtual_swapchain/cc/swapchain.cpp
+++ b/core/vulkan/vk_virtual_swapchain/cc/swapchain.cpp
@@ -408,10 +408,9 @@ vkQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR* pPresentInfo) {
         nullptr                                            // pSemaphores
     };
 
-    res = GetGlobalContext().GetQueueData(queue)->vkQueueSubmit(
-        queue, 1, &submitInfo, swp->GetFence(image_index));
+    res = EXPECT_SUCCESS(GetGlobalContext().GetQueueData(queue)->vkQueueSubmit(
+        queue, 1, &submitInfo, swp->GetFence(image_index)));
 
-    LOG_IF_FAILED(res, "vkQueuePresentKHR: vkQueueSubmit");
     if (res != VK_SUCCESS) {
       break;
     }
--- a/core/vulkan/vk_virtual_swapchain/cc/swapchain.cpp
+++ b/core/vulkan/vk_virtual_swapchain/cc/swapchain.cpp
@@ -15,11 +15,51 @@
  */
 
 #include "swapchain.h"
+
 #include <cassert>
+#include <sstream>
 #include <vector>
+
 #include "virtual_swapchain.h"
 
 namespace swapchain {
+
+// Used to set the value of VkSurfaceCapabilitiesKHR->currentExtent
+// returned from vkGetPhysicalDeviceSurfaceCapabilitiesKHR.
+// E.g. VIRTUAL_SWAPCHAIN_SURFACE_EXTENT="1960 1080"
+// If unset then the current extent will be the "special value"
+// {0xFFFFFFFF, 0xFFFFFFFF}, which some apps don't handle well.
+// I.e. they will try to create a swapchain with this maximum extent size and we
+// will then fail to create a buffer of this size.
+const char* kOverrideSurfaceExtentEnv = "VIRTUAL_SWAPCHAIN_SURFACE_EXTENT";
+// Android property names must be under 32 characters in Android N and below.
+const char* kOverrideSurfaceExtentAndroidProp = "debug.vsc.surface_extent";
+
+namespace {
+
+void OverrideCurrentExtentIfNecessary(VkExtent2D* current_extent) {
+  std::string overridden_extent;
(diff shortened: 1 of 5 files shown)


core/vulkan/vk_virtual_swapchain/cc/virtual_swapchain.cpp
```diff
@@ -353,27 +381,35 @@ void VirtualSwapchain::CopyThreadFunc() {
       }
     }
 
-    VkResult ret = functions_->vkWaitForFences(
+    VkResult res = functions_->vkWaitForFences(
         device_, 1, &image_data_[pending_image].fence_, false, UINT64_MAX);
-    (void)ret;  // TODO: Check this?
-    functions_->vkResetFences(device_, 1, &image_data_[pending_image].fence_);
+    LOG_IF_FAILED(res, "CopyThreadFunc: vkWaitForFences");
```
AWoloszyn, line 386: In general I would prefer an idiom more like 
```
VkResult res = EXPECT_SUCCESS(functions_->vkWaitForFences(
        device_, 1, &image_data_[pending_image].fence_, false, UINT64_MAX);

#define EXPECT_SUCCESS(...) \
    ({ auto r = __VA_ARGS__; if (r != 0) { printf(#__VA_ARGS__);}  r;})
```
Or if you want it to work in MSVC as well, wrap that in a lambda

This will output the exact code that failed, (You can also stick in a line/file if you want)

paulthomson: OK thanks. I have done something like this now. The warnings look like this:

```
VirtualSwapchainLayer: core/vulkan/vk_virtual_swapchain/cc/virtual_swapchain.cpp:225: functions_->vkAllocateMemory( device_, &image_memory_info, pAllocator, &image_data.image_memory_) RETURNED: -2
```

core/vulkan/vk_virtual_swapchain/cc/swapchain.cpp
```diff
@@ -15,12 +15,52 @@
...
 
+// Used to set the value of VkSurfaceCapabilitiesKHR->currentExtent
+// returned from vkGetPhysicalDeviceSurfaceCapabilitiesKHR.
+// E.g. VIRTUAL_SWAPCHAIN_SURFACE_EXTENT="1960 1080"
+// If unset then the current extent will be the "special value"
+// {0xFFFFFFFF, 0xFFFFFFFF}, which some apps don't handle well.
+// I.e. they will try to create a swapchain with this maximum extent size and we
+// will then fail to create a buffer of this size.
+const char* kOverrideSurfaceExtentEnv = "VIRTUAL_SWAPCHAIN_SURFACE_EXTENT";
+// Android property names must be under 32 characters in Android N and below.
```
hevrard, line 35: Do we care about Android N and below?

core/vulkan/vk_virtual_swapchain/cc/swapchain.cpp
```diff
@@ -355,13 +408,33 @@ vkQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR* pPresentInfo) {
...
-        queue, 1, &submitInfo, swp->GetFence(image_index));
-    res |= swp->PresentToSurface(queue, image_index);
+    res = EXPECT_SUCCESS(GetGlobalContext().GetQueueData(queue)->vkQueueSubmit(
+        queue, 1, &submitInfo, swp->GetFence(image_index)));
+
+    if (res != VK_SUCCESS) {
+      break;
+    }
+
+    res = swp->PresentToSurface(queue, image_index);
```
hevrard, line 418: Should EXPECT_SUCCESS be used here too?

core/vulkan/vk_virtual_swapchain/cc/swapchain.cpp
paulthomson: I vote yes because Vulkan exists on Android N (and this Vulkan layer can be used outside AGI) and the only requirement is we keep property names under 32 chars, which is not too bad.

hevrard: Sounds good!

core/vulkan/vk_virtual_swapchain/cc/swapchain.cpp
paulthomson: I decided to focus just on Vulkan calls, and just on those that I think should always succeed. Possibly this could fail and it would not be unexpected. Not sure. The EXPECT_SUCCESS macro will not do the break though, so we need that.

hevrard: Makes sense!

PR #215, opened 2020-04-16
Transform refactor

yalcinmelihyasin:


gapis/api/command_generator/command_generator.go
```diff
@@ -0,0 +1,25 @@
...
+package command_generator
+
+import (
+	"context"
+
+	"github.com/google/gapid/gapis/api"
+)
+
+type CommandGenerator interface {
+	GetNextCommand(ctx context.Context) api.Cmd
```
purvisa-at-google-com, line 24: When we add DCE, returning nil will no longer be enough to know we're at the end of the commands. We'll need another isAtEnd() function or something of that nature.

purvisa-at-google-com: Sorry this wasn't made super clear in the design doc!

gapis/api/command_generator/trace_reader.go
```diff
@@ -0,0 +1,43 @@
...
+
+package command_generator
+
+import (
+	"context"
+
+	"github.com/google/gapid/gapis/api"
+)
+
+type TraceReader struct {
```
purvisa-at-google-com, line 23: This isn't really reading a trace. Maybe a name like TrivialCommandGenerator or SliceCommandGenerator?

gapis/api/transform2/transform_chain.go
```diff
@@ -0,0 +1,120 @@
...
+	cmds := make([]api.Cmd, 0)
+	var err error
+	for _, transform := range chain.transforms {
+		cmds, err = transform.BeginTransform(ctx, cmds, chain.inputState)
+		if err != nil {
+			log.W(ctx, "Begin Transform Error [%v] : %v", transform, err)
+			// return nil, err
+		}
+
+		err = chain.mutateState(ctx, 0, cmds, chain.inputState)
```
purvisa-at-google-com, line 70: I know you haven't done the handleInitialState() / needsAccurateState() work yet, but this should probably be outside the loop until you handle that stuff. When that stuff is implemented, this will be inside the loop, but there will be something like "if transform.needsAccurateState() or isLastTransform() {" guarding it, and the state that is mutated will be selected based on the transform for this loop iteration.

gapis/api/transform2/transform_chain.go
```diff
@@ -0,0 +1,120 @@
...
+	var err error
+
+	for _, transform := range chain.transforms {
+		cmds, err = transform.EndTransform(ctx, cmds, chain.inputState)
+		if err != nil {
+			log.W(ctx, "End Transform Error [%v] : %v", transform, err)
+			// return nil, err
+		}
+
+		err = chain.mutateState(ctx, 0, cmds, chain.inputState)
```
purvisa-at-google-com, line 91: See comment about mutate in beginTransform()

gapis/api/transform2/transform_chain.go
```diff
@@ -0,0 +1,120 @@
...
+	var err error
+
+	for _, transform := range chain.transforms {
+		cmds, err = transform.TransformCommand(ctx, id, cmds, chain.inputState)
+		if err != nil {
+			log.W(ctx, "Error on Transform on cmd [%v:%v] with transform [%v] : %v", id, inputCmd, transform, err)
+			// return nil, err
+		}
+
+		err = chain.mutateState(ctx, id, cmds, chain.inputState)
```
purvisa-at-google-com, line 112: Same comment as beginTransform/endTransform

gapis/api/transform2/transformer.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"context"
+
+	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/api/command_generator"
+	"github.com/google/gapid/gapis/replay/builder"
+)
+
+type Transformer struct {
```
purvisa-at-google-com, line 25: Yeah, looking at the contents of this thing, it is going to evolve into the ControlFlowGenerator, as we had discussed in the video meetings!

gapis/api/transform2/transformer.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (transformer *Transformer) TransformAll(ctx context.Context) error {
+	transformChain := CreateTransformChain(transformer.state, transformer.transforms)
+
+	beginCmds, err := transformChain.BeginChain(ctx)
```
purvisa-at-google-com, line 58: This should happen automatically when getNextTransformedCommands() is called

gapis/api/transform2/transformer.go
```diff
@@ -0,0 +1,108 @@
...
+		if err != nil {
+			log.E(ctx, "Error on building transformed commands: [%v:%v] %v", currentCommandID, currentCommand, err)
+			//return err
+		}
+
+		currentCommand = transformer.commandGenerator.GetNextCommand(ctx)
+		currentCommandID++
+	}
+
+	endCmds, err := transformChain.EndChain(ctx)
```
purvisa-at-google-com, line 89: Similarly to BeginChain this should automatically happen.

gapis/api/transform2/transformer.go
```diff
@@ -0,0 +1,108 @@
...
+			transformer.builder.RevertCommand(err)
+			log.W(ctx, "Failed to write command [%v:%V] %v for replay: %v", id, i, cmd, err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (transformer *Transformer) TransformAll(ctx context.Context) error {
```
purvisa-at-google-com, line 55: I've resolved two of my own comments inside this function. As we talked about, I think the contents of this function needs to be rearranged into the transform chain, but for now, this is fine to bootstrap the rest of the system, and we can revisit this when the true ControlFlowGenerator gets added.

PR #216, opened 2020-04-16
Fix format definition of single precision float

chrisforbes:


chrisforbes:
This doesn't actually get used for much (these formats are special-cased everywhere), but is passed all over the place in RPCs and I'm sick of seeing nonsense in the dumps.

chrisforbes:
Fixed double as it was wrong too.

PR #218, opened 2020-04-16
Fix initial layout of profile tab, putting it at the top.

pmuetschard:
Note for users that have already opened AGI, this will have no effect. The Profile tab already had put itself at the top if it was added via the view menu, but if the tool is run for the first time, or if .agic's tab info is cleared, it was not put in the right place.

Bug: http://b/153899826


hevrard:
Linux build fails is a false alarm due to Swarming task expiring, see b/154297215

PR #225, opened 2020-04-17
Store cumulative nightly results, refactor Swarming result collection

hevrard:


kokoro/linux/build.sh
```diff
@@ -98,19 +98,30 @@ mkdir luci-py
 tar xzf luci-py.tar.gz --directory luci-py
 export LUCI_CLIENT_ROOT="$PWD/luci-py/client"
 
-# Prepare Swarming files
-# SWARMING_X20_TEST_DIR enables different Kokoro jobs to use different sets of tests
+
+# Initialize some environment variables, unless they have already been set
+# (e.g. by build-nightly.sh)
+if [ -z "${SWARMING_TIMESTAMP}" ] ; then
```
pau-baiget, line 104: don't you want to set this variable unconditionally so you have the most accurate date? Is there a concurrence problem you're trying to avoid?

test/swarming/collect.py
```diff
@@ -0,0 +1,144 @@
...
+    #       }
+    #     }
+    #   }
+    # }
+
+    if os.path.exists(args.results_json):
+        with open(args.results_json, 'r') as f:
+            results = json.load(f)
+    else:
+        print('Warning: results file "' + args.results_json + '" does not exists, it will be created')
```
pau-baiget, line 109: Nit: exists -> exist

test/swarming/collect.py
```diff
@@ -0,0 +1,144 @@
...
+        results[args.build_timestamp]['tests'][device] = {}
+    assert args.test_name not in results[args.build_timestamp]['tests'][device].keys()
+    results[args.build_timestamp]['tests'][device][args.test_name] = {
+        'task_id': task_id,
+        'status': status
+    }
+
+    #### Write back result file
+    os.makedirs(os.path.dirname(args.results_json), exist_ok=True)
+    with open(args.results_json, 'w') as f:
```
pau-baiget, line 130: You want to append to that file when the file existed, right? Wouldn't it be better to open it with "append"?
```
open(args.results_json, "a")...
```

pau-baiget: Ah sorry, the `results` variable already contains the previous results.

kokoro/linux/build.sh
hevrard: This is set to the timestamp at the beginning of the build in build-nightly.sh, such that this timestamp is close to the start-time of the nightly build, which makes it easier to correlate which build corresponds to which timestamp.

The code you point out sets the timestamp later in the build.sh script for PRs, as a result this timestamp is approx 1h30 behind the start of the build, because we build AGI in between.

I agree this is all a bit convoluted, I plan to clean-up in follow-up PRs.

test/swarming/collect.py
hevrard: Fixed.

PR #226, opened 2020-04-17
Make CPU model determination failure non-fatal.

pmuetschard:
- Always log errors if device info fails for some reason.
- Don't fail if reading ro.boot.hardware.platform fails.


chrisforbes:
Need this to unblock Note10 work.

PR #234, opened 2020-04-21
Fix linux-test/presubmit

hevrard:
The tests broke with #191

Bug: b/154592150


hevrard:
Build is failing because of a GitHub outage

hevrard:
Merging as the linux-test CI is green, only CI-linux fails and this is due to a github outage.

PR #238, opened 2020-04-22
Fix packed formats and add missing image formats

pau-baiget:
Packed formats are laid out backwards in Vulkan, see their definition at:
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#formats-definition

This PR fixes the existing Vulkan formats listed in AGI.

Bug: b/154018456


core/stream/fmts/abgr.go
```diff
@@ -17,6 +17,26 @@ package fmts
 import "github.com/google/gapid/core/stream"
 
 var (
+	ABGR_U4_NORM = &stream.Format{
```
AWoloszyn, line 20: Please add these formats to the format test package.
core/stream/fmts/fmts_test.go

PR #244, opened 2020-04-24
Adds a new SplitCapture API that allows splitting captures.

pmuetschard:
Along with a gapit command to exercise the API, this allows debugging
captures by chopping them into bits.

### DIFF
--- a/gapis/server/server.go
+++ b/gapis/server/server.go
@@ -39,6 +39,7 @@ import (
 	"github.com/google/gapid/core/app/status"
 	"github.com/google/gapid/core/event/task"
 	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/core/memory/arena"
 	"github.com/google/gapid/core/os/android/adb"
 	"github.com/google/gapid/core/os/device/bind"
 	"github.com/google/gapid/core/os/file"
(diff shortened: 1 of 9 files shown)


hevrard:
Sounds useful!

gapis/server/server.go
```diff
@@ -337,6 +338,35 @@ func (s *server) DCECapture(ctx context.Context, p *path.Capture, requested []*p
...
+	}
+
+	if from >= len(c.Commands) || to > len(c.Commands) || (to != 0 && to <= from) {
+		return nil, fmt.Errorf("Invalid range: [%d, %d) not in [%d, %d)", from, to, 0, len(c.Commands))
+	} else if to == 0 {
+		to = len(c.Commands)
+	}
+
+	name := fmt.Sprintf("%s [%d,%d)", c.Name(), from, to)
+	if gc, err := capture.NewGraphicsCapture(ctx, arena.New(), name, c.Header, c.InitialState, c.Commands[from:to]); err != nil {
```
AWoloszyn, line 363: I think this will generate you a VERY unstable capture. Since the state block is *not* in the correct state. (Commands may reference non-existent objects, or other horrible issues). So keep this in mind. (You may want to document this).

pau-baiget:
Is this required for b/141368458?

gapis/server/server.go
pmuetschard: Indeed. At first I just had it be a truncate kind of thing where only the end index could be given, but then it was simple/only made sense to also allow a start index. I was thinking that it might be useful to allow a "forward state" flag that will mutate the initial state to the given start command, but didn't need it myself, so didn't spend the time on it.

pmuetschard:
> Is this required for b/141368458?

I don't think this is required for anything. I just used this to debug an issue.

PR #245, opened 2020-04-24
Add a hidden flag to gapis to disable eager dep-graph resolving.

pmuetschard:
This helps in debugging problematic traces with gapit that would otherwise
crash gapis in the dependency graph resolve (e.g. panic in mutate) as they
can now be loaded.


pmuetschard:
> Should the commit title be "Add ... to enable ..." ?

No, since it's currently always enabled, the flag allows you to disable it.

PR #246, opened 2020-04-24
printf functionality in the API files.

pmuetschard:
This is meant to only be used for debugging.


gapil/bapi/bapi.proto
```diff
@@ -848,6 +855,7 @@ message Statement {
     uint64 slice_assign = 19;   // SliceAssign ID
     uint64 switch = 20;         // Switch ID
     uint64 write = 21;          // Write ID
+    uint64 print = 22;          // Print ID
```
pau-baiget, line 858: Why are you putting the ID at the end (thus breaking the nice alphabetic order, which is fine) but in the previous addition to the Node enum you prefer to insert the new print id in the middle of the list?

gapis/api/templates/api_spy.cpp.tmpl
```diff
@@ -468,4 +469,16 @@
...
     {{Template "C++.Statement.Default" $}}
   {{end}}
-{{end}}
\ No newline at end of file
+{{end}}
+
+
+{{/*
+-------------------------------------------------------------------------------
+  Emits the logic to execute a print statement
```
pau-baiget, line 477: I'm trying to understand this PR, and i'm probably failing sometimes. Is this comment accurate? Isn't this generating the print statement, rather than executing it (which happens in the mutate template below)?

gapis/api/templates/mutate.go.tmpl
```diff
@@ -27,6 +27,7 @@
     "strings"
 
     "github.com/google/gapid/core/event/task"
+    "github.com/google/gapid/core/log"
```
pau-baiget, line 30: Not sure why the linux build is complaining about this import

gapil/bapi/bapi.proto
pmuetschard: Done.

gapis/api/templates/api_spy.cpp.tmpl
pmuetschard: Correct. It doesn't execute it, it *emits* the code that will execute the print. This is a comment I copy-pasted and adjusted from other templates. For some reason we've used "logic to execute" rather than "code" in all the templates...

gapis/api/templates/mutate.go.tmpl
pmuetschard: Done.

PR #257, opened 2020-04-28
Refactor the Swarming bot scripts in Python

hevrard:
Also get rid of the need for env.sh files, which are superseded by
params.json files.


test/swarming/bot-harness.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python3
```
pau-baiget, line 1: Just checking that it's ok to have different Python version requirements in a single codebase

hevrard: Good point, but I am not aware of any other Python file in the codebase, except the config file for MacOS build (`kokoro/macos/dmg-settings.py`) ?

pau-baiget: Bazel depends on Python 2.7 IIRC

PR #261, opened 2020-04-29
Make state rebuilding more deterministic

hevrard:
This is a big but boring change to reduce the sources of non-determinism in the state rebuilder. It is only a first step, but it already massively reduces differences in the list of generated commands.

Make sure to review commit-by-commit, to have piecemeal diffs.

Tested on an MEC trace that generates 50k+ commands for state rebuilding, these changes brings the number of diff chunks (between the lists of state-rebuilding commands generated by several calls to `linearize_trace`) from ~2000 down to ~30.

Bug: b/155048196

### DIFF
--- a/gapis/api/vulkan/state_rebuilder.go
+++ b/gapis/api/vulkan/state_rebuilder.go
@@ -720,11 +720,19 @@ func (sb *stateBuilder) createDevice(d DeviceObjectʳ) {
 		queuePriorities[q.QueueFamilyIndex()][q.QueueIndex()] = q.Priority()
 	}
 	reorderedQueueCreates := map[uint32]VkDeviceQueueCreateInfo{}
-	i := uint32(0)
-	for k, v := range queueCreate {
-		v.SetPQueuePriorities(NewF32ᶜᵖ(sb.MustAllocReadData(queuePriorities[k]).Ptr()))
-		reorderedQueueCreates[i] = v
-		i++
+	{
+		keys := []uint32{}
+		for k := range queueCreate {
+			keys = append(keys, k)
+		}
+		sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
+		i := uint32(0)
+		for _, k := range keys {
+			v := queueCreate[k]
+			v.SetPQueuePriorities(NewF32ᶜᵖ(sb.MustAllocReadData(queuePriorities[k]).Ptr()))
+			reorderedQueueCreates[i] = v
+			i++
+		}
 	}
 
 	pNext := NewVoidᵖ(memory.Nullptr)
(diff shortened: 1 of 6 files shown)


gapis/api/vulkan/state_rebuilder.go
```diff
@@ -461,11 +461,13 @@ func (sb *stateBuilder) write(cmd api.Cmd) {
 
 func (sb *stateBuilder) createInstance(vk VkInstance, inst InstanceObjectʳ) {
 	enabledLayers := []Charᶜᵖ{}
-	for _, layer := range inst.EnabledLayers().All() {
+	for _, k := range inst.EnabledLayers().Keys() {
+		layer := inst.EnabledLayers().Get(k)
```
chrisforbes, line 465: Whoa -- we must preserve the original order here! Layer order matters.

gapis/api/vulkan/state_rebuilder.go
```diff
@@ -681,18 +683,21 @@ func (sb *stateBuilder) createSurface(s SurfaceObjectʳ) {
 
 func (sb *stateBuilder) createDevice(d DeviceObjectʳ) {
 	enabledLayers := []Charᶜᵖ{}
-	for _, layer := range d.EnabledLayers().All() {
+	for _, k := range d.EnabledLayers().Keys() {
```
chrisforbes, line 686: ditto

gapis/api/vulkan/state_rebuilder.go
```diff
@@ -681,18 +683,21 @@ func (sb *stateBuilder) createSurface(s SurfaceObjectʳ) {
...
 		enabledLayers = append(enabledLayers, NewCharᶜᵖ(sb.MustAllocReadData(layer).Ptr()))
 	}
 	enabledExtensions := []Charᶜᵖ{}
-	for _, ext := range d.EnabledExtensions().All() {
+	for _, k := range d.EnabledExtensions().Keys() {
+		ext := d.EnabledExtensions().Get(k)
 		enabledExtensions = append(enabledExtensions, NewCharᶜᵖ(sb.MustAllocReadData(ext).Ptr()))
 	}
 
 	queueCreate := map[uint32]VkDeviceQueueCreateInfo{}
```
chrisforbes, line 696: You're still iterating over this map on :724 below..

gapis/api/vulkan/state_rebuilder.go
hevrard: I think this change actually fixes it, as the EnabledLayers here is a map of index to layer names, and they are stored in order during capture:

https://github.com/google/agi/blob/master/gapis/api/vulkan/api/instance.api#L120-L134

So following the order of this map keys (as we do here with ranging on `Keys()`) should respect the original order.

gapis/api/vulkan/state_rebuilder.go
hevrard: Fixed.

gapis/api/vulkan/state_rebuilder.go
chrisforbes: :+1:

Issue #254, opened 2020-04-28
Build Documentation Incorrect


tobine:
The AGI Windows build instructions under [Install Bazel](https://github.com/google/agi/blob/master/BUILDING.md#install-bazel) state:

> Note: Installing bazel will also install MSYS into C:\tools\msys64 and Python into C:\tools\python27.

However, I successfully followed the instructions to run `choco install bazel` and bazel is installed without having msys64 or python27 installed in the c:\tools dir.

hevrard:
Note on this: now that we're using Bazel 2.0.0, we don't need Python2.7 as a dependency any more (it was required for the previous version of Bazel we used).

pau-baiget:
I've got a question regarding Python dependency. Installed Python 3.8 in my windows desktop and I was getting the following errors while building:

```bazel build pkg
Starting local Bazel server and connecting to it...
ERROR: C:/development/agi-2/agi/core/vulkan/vk_virtual_swapchain/cc/BUILD.bazel:75:1: in strip rule //core/vulkan/vk_virtual_swapchain/cc:libVkLayer_VirtualSwapchain_stripped.dll:
Traceback (most recent call last):
        File "C:/development/agi-2/agi/core/vulkan/vk_virtual_swapchain/cc/BUILD.bazel", line 75
                strip(name = 'libVkLayer_VirtualSwapchain_stripped.dll')
        File "C:/development/agi-2/agi/tools/build/rules/cc.bzl", line 62, in _strip_impl
                ctx.actions.run(<4 more arguments>)
        File "C:/development/agi-2/agi/tools/build/rules/cc.bzl", line 63, in ctx.actions.run
                cc_toolchain.strip_executable()
'string' object is not callable
ERROR: Analysis of target '//:pkg' failed; build aborted: Analysis of target '//core/vulkan/vk_virtual_swapchain/cc:libVkLayer_VirtualSwapchain_stripped.dll' failed; build aborted
INFO: Elapsed time: 5.722s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (84 packages loaded, 1829 targets configured)
    Fetching @androidndk; Restarting.
```

I fixed that with this patch that I apply locally:

```
diff --git a/gapidapk/android/apk/rules.bzl b/gapidapk/android/apk/rules.bzl
index 81e28920..5de1c921 100644
--- a/gapidapk/android/apk/rules.bzl
+++ b/gapidapk/android/apk/rules.bzl
[...]

PR #262, opened 2020-04-29
Remove Python dependency from Windows build

hevrard:


hevrard:
Related to #254 : this experiment proves that Python is indeed needed to build perfetto.

PR #266, opened 2020-05-02
Basic Framebuffer Attachment Picker

hliatis:
Bug: b/150649691

Attachment picker is collapsible.

Note: problems with depth attachments such as tone mapping
and resizing will be handled in the future.


gapic/src/main/com/google/gapid/views/FramebufferView.java
```diff
@@ -128,14 +139,37 @@ public FramebufferView(Composite parent, Models models, Widgets widgets) {
 
     setLayout(new GridLayout(2, false));
 
-    attachmentListener = new AttachmentListener(widgets.theme);
+    Composite header = createComposite(this, new GridLayout(2, false));
 
-    toolBar = createToolBar(widgets.theme);
+    picker = new AttachmentPicker(this);
```
stellama0208, line 144: Can we create the picker at the right side of the toolbar? So that we can avoid the movement and blankness of the toolbar after clicking show attachment.

gapic/src/main/com/google/gapid/views/FramebufferView.java
```diff
@@ -128,14 +139,37 @@ public FramebufferView(Composite parent, Models models, Widgets widgets) {
...
+    picker.addContentListener(SWT.MouseDown,
+      e -> picker.selectAttachment(picker.getItemAt(e.x)));
+
+    ToolBar toolBar = createToolBar(widgets.theme);
     imagePanel = new ImagePanel(this, View.Framebuffer, models.analytics, widgets, true);
 
+    GridData pickerGridData = new GridData(SWT.FILL, SWT.FILL, true, true, 2, 1);
```
stellama0208, line 151: Would it be better to make it `new GridData(SWT.FILL, SWT.FILL, true, false, 2, 1);`? It will avoid blank area between the picker strip and the framebuffer below.

hliatis: My worry with this is that now it's too small in some cases. I'm looking into having it resize or use a sashform.

gapic/src/main/com/google/gapid/views/FramebufferView.java
hliatis: We are basing this on UI mockups linked in go/gbuffer-strip. Direct link: https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZUyss3W9BIEB/files/MCEJu8Y2hyDScXfsMX7dIYFTKmMLUPLxbBY

We went with the top one because it was more discoverable, which was a problem with the original we wanted to address. In the bottom is where it is part of the toolbar like before.

However, you are suggesting to have it function like the top (not part of the toolbar, has a show/hide button), but look like the bottom in that it is to the side of the toolbar, correct? I can see the appeal to that. 

Anyone with second opinions?

gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -73,7 +73,9 @@
           .setCheckForUpdates(true))
       .setUi(SettingsProto.UI.newBuilder()
           .setPerfetto(SettingsProto.UI.Perfetto.newBuilder()
-              .setDrawerHeight(250)))
+              .setDrawerHeight(250))
+          .setFramebufferPicker(SettingsProto.UI.FramebufferPicker.newBuilder()
+              .setEnabled(true)))
```
stellama0208, line 78: Would it be better if we hide the feature for a while by setting the default toggle to false here? When we test the PR we can set it to true locally.

PR #269, opened 2020-05-06
Update to Java11 using Zulu JDK

hevrard:
This is an experiment, do NOT merge.


hevrard:
We may need to add a reference to the JDK license(s)

hevrard:
Updated to use the Zulu JDK from Bazel. We need to double-check that the Windows and macOS build artifacts are running fine.

Currently, this adds the whole JDK source code as part of the artifact. We can hopefully get rid of this soon, and just point to the Bazel mirror page.

PR #270, opened 2020-05-06
Introduce experimental flag.

lpy:
This patch introduces the experimental flag for all under development features.
Features under development are unstable and unshaped, and are changed a lot
frequently. However, they are also under team-fooding actively. To balance
this, we put such features behind the experimental flag. UI elements must also
have clear indication that the features are experimental.

This patch also puts frame lifecycle behind the experimental flag.

Bug: b/155889093

### DIFF
--- a/core/app/flags/experimental.go
+++ b/core/app/flags/experimental.go
@@ -1,0 +1,28 @@
+// Copyright (C) 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package flags
+
+import (
+	"flag"
+)
+
+// Experimental features are hidden behind the flags. All experimental feature flags must:
+//     1) be named --experimental-enable-<feature-name>
+//     2) be by default false/off
+//     3) be removed once the feature is no longer in experiment
+var (
+	EnableFrameLifecycle = flag.Bool("experimental-enable-frame-lifecycle", false, "Enable the experimental feature Frame Lifecycle.")
+	EnableVulkanTracing  = flag.Bool("experimental-enable-vulkan-tracing", false, "Enable the experimental feature Vulkan tracing.")
+)
--- a/gapic/src/main/com/google/gapid/server/GapisProcess.java
+++ b/gapic/src/main/com/google/gapid/server/GapisProcess.java
@@ -21,6 +21,7 @@ import static java.util.logging.Level.WARNING;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.gapid.models.Settings;
+import com.google.gapid.util.Experimental;
 import com.google.gapid.util.Flags;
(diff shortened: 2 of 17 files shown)


lpy:
ptal

gapis/trace/tracer/default_api_trace_options.go
```diff
@@ -22,7 +22,7 @@ import (
 func VulkanTraceOptions() *service.TraceTypeCapabilities {
 	return &service.TraceTypeCapabilities{
 		Type:                           service.TraceType_Graphics,
-		Api:                            "Vulkan",
+		Api:                            "Vulkan (experimental)",
```
pmuetschard, line 25: Does this work? I would expect this to break things in fun ways when the UI tries to take a trace and passes this string back to the server.

core/app/flags/experimental.go
```diff
@@ -0,0 +1,28 @@
...
+	"flag"
+)
+
+// Experimental features are hidden behind the flags. All experimental feature flags must:
+//     1) be named --experimental-enable-<feature-name>
+//     2) be by default false/off
+//     3) be removed once the feature is no longer in experiment
+var (
+	EnableFrameLifecycle      = flag.Bool("experimental-enable-frame-lifecycle", false, "Enable the experimental feature Frame Lifecycle.")
+	EnableVulkanFrameProfiler = flag.Bool("experimental-enable-vulkan-frame-profiler", false, "Enable the experimental feature Vulkan Frame Profiler.")
```
pmuetschard, line 27: I'd prefer to call this `--experimental-enable-vulkan-tracing`.

gapic/src/main/com/google/gapid/server/GapisProcess.java
```diff
@@ -96,6 +97,20 @@ protected Exception prepare(ProcessBuilder pb) throws GapiPaths.MissingToolsExce
...
+      args.add("--experimental-enable-vulkan-frame-profiler");
+    } else {
+      if (Experimental.enableFrameLifecycle.get()) {
+        args.add("--experimental-enable-frame-lifecycle");
+      }
+      if (Experimental.enableVulkanFrameProfiler.get()) {
+        args.add("--experimental-enable-vulkan-frame-profiler");
+      }
+    }
+
```
pmuetschard, line 113: Can you move this logic into your new `Experimental` class? i.e. have something like `Experimental.getGapisFlags(..)`. This would keep the logic a bit more contained.

gapis/trace/tracer/default_api_trace_options.go
lpy: works now, miss a string compare

lpy:
Thanks, ptal

gapis/trace/tracer/default_api_trace_options.go
pmuetschard: As the failure in the build shows, this is too far-reaching of a change. Instead of changing the API here, just change the label in the UI. In the `createApiDropDown(..)` function, simple change the Graphics case to: `return ttc.getApi() + " (Experimental)"`.

lpy:
Thanks, ptal!

gapic/src/main/com/google/gapid/views/SettingsDialog.java
```diff
@@ -79,6 +80,10 @@ protected void beforeAnalytics() {
             "Disable replay optimization",
             models.settings.preferences().getDisableReplayOptimization()),
             withSpans(new GridData(SWT.LEFT, SWT.TOP, false, false), 2, 1));
+        enableAllExperimentalFeatures = withLayoutData(createCheckbox(this,
+            "Enable all experimental features (unsupported, required restart)",
```
pmuetschard, line 84: Either "Enable all unsupported experimental features (requires restart)" or drop "unsupported". The way it's written now is a bit confusing. Also notice that it is "requires" not "required".

PR #272, opened 2020-05-07
Show submission node in command list for empty command buffer.

stellama0208:
- Make sync.Data.SubcommandGroups point to the the next utilizable
    subcommand index, so that we can identify an empty command buffer at
    command tree creation time.
 - Bug: b/143733663.


stellama0208:
> LGTM, could you add the relevant part of the good recap you wrote in b/143733663#comment4 in the commit message?

Thanks for suggestion. :) It's updated now!

PR #276, opened 2020-05-08
WIP:Hacking ANGLE frame capture support

tobine:
Not meant to be functional, just trying to capture all of the
modifications required to support frame capture with ANGLE over
OpenGL.

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -373,6 +373,35 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
 	return result, nil
 }
 
+// TODO: Currently using Android Q/10, API 29 as ANGLE support cut-off
+// TODO: If I knew squat about go then I would refactor this to be shared
+//  function w/ "SupportsPerfetto" and have const values defined somewhere
+//  such as "PERFETTO_SUPPORT_OS_VERSION = 28" and "ANGLE_SUPPORT_OS_VERSION = 29"
+//  then call that shared version with the version as input
+func (b *binding) SupportsAngle(ctx context.Context) bool {
+	os := b.Instance().GetConfiguration().GetOS()
+	return os.GetAPIVersion() >= 29
+}
+
+func (b *binding) QueryAnglePackageName(ctx context.Context) (string, error) {
+	if !b.SupportsAngle(ctx) {
+		return "", fmt.Errorf("ANGLE not supported on this device")
+	}
+	// ANGLE supported, so check for installed ANGLE package
+	//  Favor custom installed package first, followed by default system package
+	custom := "com.chromium.angle"
+	system := "com.google.android.angle"
+	// Check installed packages for ANGLE package
+	packages, _ := b.InstalledPackages(ctx)
+	if pkg := packages.FindByName(custom); pkg != nil {
+		return custom, nil
+	}
+	if pkg := packages.FindByName(system); pkg != nil {
+		return system, nil
+	}
+	return "", fmt.Errorf("No ANGLE packages installed on this device")
+}
+
 func extrasFlags(extras []android.ActionExtra) []string {
 	flags := []string{}
 	for _, e := range extras {
--- a/core/os/android/adb/device.go
+++ b/core/os/android/adb/device.go
@@ -217,6 +217,11 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
 		d.To.Configuration.PerfettoCapability = perfettoCapability
 	}
 
+	// Query device ANGLE support
+	if anglePackage, err := d.QueryAnglePackageName(ctx); err == nil {
+		d.To.Configuration.AnglePackage = anglePackage
+	}
+
 	// If the VkRenderStagesProducer layer exist, we assume the render stages producer is
 	// implemented in the layer.
 	for _, l := range d.To.Configuration.GetDrivers().GetVulkan().GetLayers() {
--- a/core/os/device/device.proto
+++ b/core/os/device/device.proto
@@ -173,6 +173,8 @@ message Configuration {
   Drivers drivers = 4;
   // Perfetto capability.
   PerfettoCapability perfetto_capability = 5;
+  // Angle package name
+  string angle_package = 6;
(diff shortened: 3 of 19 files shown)


core/os/android/adb/device.go
```diff
@@ -217,6 +217,11 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
 		d.To.Configuration.PerfettoCapability = perfettoCapability
 	}
 
+	// Query device ANGLE support
+	if anglePackage, err := d.QueryAnglePackageName(ctx); err == nil {
+		d.To.Configuration.AnglePackage = anglePackage
```
tobine, line 222: This is meant to set a variable based on the queried angle package name.
However, what I don't understand is what is the angle_package string that I put into device.proto used for? Do I need to fill that in here? If so, how and why?

core/os/device/device.proto
```diff
@@ -173,6 +173,8 @@ message Configuration {
   Drivers drivers = 4;
   // Perfetto capability.
   PerfettoCapability perfetto_capability = 5;
+  // Angle package.
+  string angle_package = 6;
```
tobine, line 177: Added this member to the proto buffer, but still not clear to me how it gets filled in or what I need to use it for.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -167,6 +167,8 @@ public static void showSaveTraceDialog(Shell shell, Models models) {
   }
 
   public static void showTracingDialog(Client client, Shell shell, Models models, Widgets widgets) {
+	// TODO: I believe somewhere around here is where we should output a message about which
+	//  ANGLE package is being used if we're in ANGLE trace mode
```
tobine, line 171: Not sure exactly where to output a message that tells the user "enabling angle package <pkg_name> to trace OpenGL app <app_name> with Vulkan over ANGLE" or something like that. Also need to have those two names available when I output the message.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -578,6 +581,7 @@ public String getText(Object element) {
             switch (ttc.getType()) {
               case Graphics: return ttc.getApi();
               case Perfetto: return "System Profile";
+			  case Angle: return "OpenGL on ANGLE";
```
tobine, line 584: Want a new pull-down option in "Type" for "OpenGL on ANGLE" but the trace itself is not really a new type. This selection really means "Do a Graphics Vulkan trace, but first run a couple of commands to make the OpenGL app run using ANGLE which will handle the Vulkan translation."

tobine:
This is more a set of hacks and ?s then an initial PR, but wanted to get it up here as a starting point. @silence-do-good when you're able please let me know if anything looks completely wrong. Still lots of work to go.

core/os/android/adb/device.go
silence-do-good: Yes. The AnglePackage set here can then be retrieved in the UI code so that you know which package is enabled for the selected device. This can then allow you to either show Angle trace type or not, based on the selected device's configuration.

gapic/src/main/com/google/gapid/views/TracerDialog.java
silence-do-good: We can start with having a new trace type and get the functional components working. Once the tracing is in a good state, we can think about refactoring this to be cleaner and integrated within the graphics trace type, if we want.

silence-do-good:
Just thinking out loud. One way you to split the work here:
1) Get the UI code in first, so you can see the trace type for supported devices. This can be done by filling in the config like you've done in the go code. And then fetching that in the UI code and displaying a new trace type based on that. To test this alone, when start trace is clicked, you can redirect it to go back to the replay profiler. This way, we can make sure that the prerequisites are tested and working.
2) Maybe after the UI, the functions required to set up ANGLE can be written and plugged into.

core/os/android/adb/commands.go
```diff
@@ -372,6 +372,35 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
...
 }
 
+// TODO: Currently using Android Q/10, API 29 as ANGLE support cut-off
+// TODO: If I knew squat about go then I would refactor this to be shared
+//  function w/ "SupportsPerfetto" and have const values defined somewhere
+//  such as "PERFETTO_SUPPORT_OS_VERSION = 28" and "ANGLE_SUPPORT_OS_VERSION = 29"
+//  then call that shared version with the version as input
+func (b *binding) SupportsAngle(ctx context.Context) bool {
+	os := b.Instance().GetConfiguration().GetOS()
+	return os.GetAPIVersion() >= 29
```
lpy, line 382: Sorry for jumping into a WIP patch. Just in case I forgot. Should we just use 30(R)? Supporting one version back while we are not really actively testing it may end up biting us. Besides all quarterly releases of Q are now closed :)

tobine: I'm still doing all of my testing on Q so I'll keep it Q for now. Can potentially bump it up if/when all ANGLE engineers have moved to R.

tobine:
Alright, I've got this compiling now and it appears that the current CI fail is just for code formatting.
However, there are still some holes I'm working to fill in today. Main gaps are:

- Need to set some state when ANGLE mode is selected and then only run ADB commands to enable ANGLE on startup and disable on shutdown when ANGLE mode is selected.
- Need to save any existing ANGLE env state and restore it instead of blasting it.
- When ANGLE mode selected, everything should proceed as if we're in Graphics (frame capture) mode with the additional effect that ANGLE env vars are set prior to running.

I also need to hide this behind experimental flag but will get these other issues fixed first.

Tagging this for another review in case people can give me some pointers and/or indicate current code that's not correct.

tobine:
Ok, pushed another update. This is now behind the flag "--experimental-enable-angle-tracing." I'm at least able to see the "OpenGL on ANGLE" tracing option when using that flag. Not sure anything is working correctly yet, though.

tobine:
Killing this PR in favor of updated and closer to final #320

PR #277, opened 2020-05-09
Clean up layer handling for profiling and find_issues

chrisforbes:
Is critical that we don't have validation layers enabled during
profiling; the layers do handle remapping which destroys our ability to
correlate data coming directly from the driver through perfetto with the
replayer's own observations of handles.

Is also important to not have validation layers doubly-enabled during
the find_issues replay, as they don't work when loaded twice.

The existing code tried to do this in the find_issues case but failed to
because of a quirk in the way strings are handled.

All known layers which we'd want to leave enabled are imposed from
the environment rather than being part of the vkCreateInstance call
(eg Optimus on desktop). Just remove all explicit layers, and readd the
one we want if any (renderstages layer if profiling on a device which
requires it; validation layer if finding issues).

This fixes various bad behavior in these replays when running Bender.

Bug: b/156104191

### DIFF
--- a/gapis/api/vulkan/find_issues.go
+++ b/gapis/api/vulkan/find_issues.go
@@ -17,6 +17,7 @@ package vulkan
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/gapir"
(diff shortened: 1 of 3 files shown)


gapis/api/vulkan/find_issues.go
```diff
@@ -142,7 +115,7 @@ func (t *findIssues) Transform(ctx context.Context, id api.CmdID, cmd api.Cmd, o
 		var debugReportExtNameData api.AllocResult
 		hasDebugReport := false
 		for _, e := range exts {
-			if e.String() == debugReportExtension {
+			if debugReportExtension == strings.TrimRight(string(memory.CharToBytes(e.StringSlice(ctx, s).MustRead(ctx, cmd, s, nil))), "\x00") {
```
hevrard, line 118: if there is no better way to access the string, maybe add a small comment to clarify why this has to be so convoluted?

chrisforbes: Done.

PR #279, opened 2020-05-11
Set draw command grouping nodes' index to that of the real draw command.

stellama0208:
- Bug: b/156294498.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/CommandTree.java
+++ b/gapic/src/main/com/google/gapid/views/CommandTree.java
@@ -341,7 +341,8 @@ public class CommandTree extends Composite
                 string, string.identifierStyle());
           }
         } else {
-          string.append(Formatter.firstIndex(data.getCommands()) + ": ", string.defaultStyle());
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/views/CommandTree.java
```diff
@@ -341,7 +341,8 @@ protected void load(CommandStream.Node node, Runnable callback) {
                 string, string.identifierStyle());
           }
         } else {
-          string.append(Formatter.firstIndex(data.getCommands()) + ": ", string.defaultStyle());
+          String index = shouldShowLastIndex(data) ? Formatter.lastIndex(data.getCommands()) : Formatter.firstIndex(data.getCommands());
```
chrisforbes, line 344: Can we use the ID of the `representation` instead here so we don't have to teach the frontend that draws are different?

chrisforbes: see CommandStream.java:375 or so

stellama0208:
Change the feature request to b/156548627.

PR #286, opened 2020-05-14
Handle Input Attachments in FramebufferView

hliatis:
Bug: b/154014399

### DIFF
--- a/gapis/api/vulkan/state.go
+++ b/gapis/api/vulkan/state.go
@@ -104,6 +104,34 @@ func (st *State) getSubmitAttachmentInfo(attachment uint32) (w, h uint32,
 	}
 
 	// Search for index in input attachments
+	for _, attIndex := range subpassDesc.InputAttachments().Keys() {
+		attRef := subpassDesc.InputAttachments().Get(attIndex)
+		if attachment == attRef.Attachment() {
+			if inputAttachment, ok := lastDrawInfo.Framebuffer().ImageAttachments().Lookup(attachment); ok {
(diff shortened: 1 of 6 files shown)


gapis/api/vulkan/state.go
```diff
@@ -104,6 +104,34 @@ func (st *State) getSubmitAttachmentInfo(attachment uint32) (w, h uint32,
 	}
 
 	// Search for index in input attachments
+	for _, attIndex := range subpassDesc.InputAttachments().Keys() {
+		attRef := subpassDesc.InputAttachments().Get(attIndex)
+		if attachment == attRef.Attachment() {
+			if inputtAttachment, ok := lastDrawInfo.Framebuffer().ImageAttachments().Lookup(attachment); ok {
+				inputImg := inputtAttachment.Image()
```
stellama0208, line 111: nit: double t.

gapis/resolve/framebuffer_attachment.go
```diff
@@ -68,6 +68,10 @@ func typeToString(t api.FramebufferAttachmentType) string {
 		return "Color"
 	case api.FramebufferAttachmentType_OutputDepth:
 		return "Depth"
```
stellama0208, line 70: nit: do we want to change these two to `OutputColor` and `OutputDepth` as well?

hliatis: I rarely hear people use the qualification of "output" for these attachments like they do with input attachments. I am open to changing this.

PR #287, opened 2020-05-15
Make sure to always kill the server on UI exit

hevrard:
This strengthen the logic around terminating child processes of the UI.

Bug: b/150773653

### DIFF
--- a/gapic/src/main/com/google/gapid/server/ChildProcess.java
+++ b/gapic/src/main/com/google/gapid/server/ChildProcess.java
@@ -29,6 +29,7 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 
 /**
@@ -36,6 +37,8 @@ import java.util.logging.Logger;
  */
 public abstract class ChildProcess<T> {
   private static final Logger LOG = Logger.getLogger(ChildProcess.class.getName());
+  // Number of seconds to wait for a child process to terminate before force-killing it.
+  private static final int terminateTimeoutSec = 2;
 
   protected final String name;
   protected final Settings settings;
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/server/ChildProcess.java
```diff
@@ -104,12 +105,24 @@ protected void runProcess(final SettableFuture<T> result, final ProcessBuilder p
...
+      killProcess();
     } finally {
       onExit(exitCode);
     }
   }
 
+  protected void killProcess() {
+    process.destroy();
+    try {
+      boolean destroyed = process.waitFor(500, TimeUnit.MILLISECONDS);
```
pmuetschard, line 117: This seems arbitrary and way to short.

hevrard: My stance is that there may be a lot of reasons why GAPIS does not terminate, and we need to be robust to this. So if GAPIS does not terminate within a certain duration after process.Destroy(), we force-kill it.

The duration is completely arbitrary, indeed. How long would you allow GAPIS to terminate properly? 2 seconds?

hevrard:
I agree that we should debug why GAPIS does not terminate. But also, there may be many reason why it does not terminate, including reasons behind our control. So I think we need to be robust to this, and if the UI started GAPIS, it should make sure to force-kill it on termination.

Does this change seems more acceptable under this point of view?

hevrard:
I've updated to wait a couple of seconds, and log an error when the child process does not terminate within this duration.

@pmuetschard what do you think of this?

hevrard:
After a chat offline with @pmuetschard, we came to the conclusion that it's better to not "hide" server shutdown issues, such that the team or users can see this and report it as an issue. The server should always manage to terminate upon a Shutdown(), if it hangs, then it's a bug, and we should see it, and our users should see it.

PR #290, opened 2020-05-16
Persist gpu counters selection result with search bar.

lpy:
Currently when search bar is used to filter gpu counters, all selection result
outside of the filter will be lost after clearing the search bar. This patch
makes sure selection result is persisted properly with search bar in used.

1) selection result filtered out by the search bar will be persisted;
2) none, default and all link will only take effect on the current search
result.

Bug: b/155357212

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
@@ -57,6 +57,8 @@ import com.google.protobuf.TextFormat.ParseException;
 import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.jface.viewers.ArrayContentProvider;
 import org.eclipse.jface.viewers.CheckboxTableViewer;
+import org.eclipse.jface.viewers.CheckStateChangedEvent;
+import org.eclipse.jface.viewers.ICheckStateListener;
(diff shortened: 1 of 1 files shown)


lpy:
> I was testing this out and found out one edge case.
> 1)Type "shad" in the filter and select all.
> 2)Clear the filter and type "shader".
> 3) Press backspace twice so that the filter term becomes "shad" again. The checks will now only be present for those that were part of "shader". The remaining ones will be unchecked.
> 
> This seems to be only a UI bug, in that, if you click okay, all the elements are properly checked.

Thanks, I forgot to resume checked elements when the filter query changed. PTAL.

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -862,13 +891,27 @@ protected Point getInitialSize() {
 
       @Override
       protected void okPressed() {
-        selectedIds = Arrays.stream(table.getCheckedElements())
+        // When clicking ok button while the filter is still in used, reset all filters and resume
+        // all checked elements before collect them.
+        selectedIds = Arrays.stream(checkedElements.toArray())
```
silence-do-good, line 896: Is the comment stale?

lpy:
Windows CI is broken, I am going to proceed to land this, feel free to revert this if it causes CI issue.

PR #292, opened 2020-05-19
Report replayer crash errors to the UI

hevrard:
This adds error propagation for replayer and replayer connection
errors. Also, fix a few typos.

Bug: b/156232273


gapis/resolve/report.go
```diff
@@ -107,6 +107,13 @@ func (r *ReportResolvable) Resolve(ctx context.Context) (interface{}, error) {
 		}
 	}
 
+	// Start with issues that are not specific to a command, like
```
pau-baiget, line 110: Is this "Start with" because you're going to do related changes in this specific area of code later? Otherwise this comment seems more PR-related.

pau-baiget: Ok, i guess it ties well with the next one "Gather report items.." ?

hevrard: Yes, it is "start with" in the sense that the piece of code continues with regular commands.

PR #293, opened 2020-05-19
Stop embedding a JRE in macOS builds

hevrard:
AdoptOpenJDK JRE does not work on macOS Catalina. We are currently
exploring using an other JDK. In the meantime, we stop embedding a JRE
in macOS builds, and we offer a script (on agi-dev-releases repo) to
let the user download a external JRE.

Bug: b/150268876


hevrard:
Related: https://github.com/google/agi-dev-releases/pull/1

PR #294, opened 2020-05-19
Remove remnant of GL error tracking from gapii

chrisforbes:


chrisforbes:
Will remove more in a subsequent change.

PR #296, opened 2020-05-19
Display durations of nodes in the command tree.

pmuetschard:


gapic/src/main/com/google/gapid/models/Profile.java
```diff
@@ -119,10 +129,32 @@ public int hashCode() {
...
       this.profile = profile;
+      this.durationByGroup = aggregateSliceTimeByGroup(profile);
+      this.groups = getSortedGroups(profile, durationByGroup.keySet());
+    }
+
+    private static Map<Integer, Long> aggregateSliceTimeByGroup(Service.ProfilingData profile) {
+      Set<Integer> groups = profile.getSlices().getGroupsList().stream()
+          .map(Service.ProfilingData.GpuSlices.Group::getId)
+          .collect(toSet());
+      return profile.getSlices().getSlicesList().stream()
```
chrisforbes, line 146: On Mali, this may add up to as much as twice the actual wall clock time due to the two parallel queues. Do we just not care for now?

hevrard:
Rebase on master to fix windows build.

pau-baiget:
@stellama0208 if you are comfortable with the PR as is, please consider merging it. Thanks@

PR #298, opened 2020-05-20
Remove caller machinery

chrisforbes:
This was necessary for EGL, where we could see for example eglInitialize
call eglGetError. Vulkan has no such thing.

### DIFF
--- a/gapis/capture/decoder.go
+++ b/gapis/capture/decoder.go
@@ -89,10 +89,12 @@ func (d *decoder) endGroupImpl(ctx context.Context, id uint64, terminated bool) 
 	case *cmdGroup:
 		obj.invoked = true
 		obj.cmd.SetTerminated(terminated)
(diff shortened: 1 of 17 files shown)


gapis/capture/decoder.go
```diff
@@ -89,10 +89,7 @@ func (d *decoder) endGroupImpl(ctx context.Context, id uint64, terminated bool)
 	case *cmdGroup:
 		obj.invoked = true
 		obj.cmd.SetTerminated(terminated)
-		id := d.builder.addCmd(ctx, obj.cmd)
-		for _, c := range obj.children {
-			c.SetCaller(id)
-		}
+		d.builder.addCmd(ctx, obj.cmd)
```
pmuetschard, line 92: Should we have a check here and fail the decode if there are children?

chrisforbes: Probably. I'll add it.

PR #300, opened 2020-05-21
Use --verbose_failures in all kokoro builds

chrisforbes:


chrisforbes:
@hevrard this is the logging change extracted from the windows fix.

hevrard:
@pmuetschard yes indeed, it is present in `.bazelrc`. I added it while debugging the windows build failure when I was doing various hacks to get the build.bat script as verbose as it could upon failures, and I didn't realize this flag was already in bazlerc. Sorry for the noise!

PR #301, opened 2020-05-21
Load all the decoded memory struct values in one grpc session.

stellama0208:
- Pre-resolve the struct value at backend, from path.MemoryAsType
 to memory_box.Value. In this way, we avoid client sending out the
 resolving request a sencond time for each struct.
 - Bug: b/155059249.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/MemoryView.java
+++ b/gapic/src/main/com/google/gapid/views/MemoryView.java
@@ -792,15 +792,12 @@ public class MemoryView extends Composite
         packColumns(treeViewer.getTree());
         return;
       }
-      // long start = System.currentTimeMillis();
 
       Rpc.listen(models.types.loadStructNodes(structs),
           new UiCallback<List<StructNode>, List<StructNode>>(this, LOG) {
         @Override
         protected List<StructNode> onRpcThread(Result<List<StructNode>> result)
             throws RpcException, ExecutionException {
-          // Snippet for time cost evaluation.
-          // System.out.println("loadStructNodes() method duration: " + (System.currentTimeMillis() - start) + "ms.");
           return StructNode.simplifyTrees(result.get());
         }
 
--- a/gapic/src/main/com/google/gapid/views/MemoryView.java
+++ b/gapic/src/main/com/google/gapid/views/MemoryView.java
@@ -792,12 +792,15 @@ public class MemoryView extends Composite
         packColumns(treeViewer.getTree());
         return;
       }
+      // long start = System.currentTimeMillis();
 
       Rpc.listen(models.types.loadStructNodes(structs),
           new UiCallback<List<StructNode>, List<StructNode>>(this, LOG) {
         @Override
         protected List<StructNode> onRpcThread(Result<List<StructNode>> result)
             throws RpcException, ExecutionException {
+          // Snippet for time cost evaluation.
+          // System.out.println("loadStructNodes() method duration: " + (System.currentTimeMillis() - start) + "ms.");
           return StructNode.simplifyTrees(result.get());
         }
 
--- a/gapis/resolve/memory.go
+++ b/gapis/resolve/memory.go
@@ -127,17 +127,7 @@ func Memory(ctx context.Context, p *path.Memory, rc *path.ResolveConfig) (*servi
 				if rng.Overlaps(r) {
 					interval.Merge(&reads, rng.Window(r).Span(), false)
 					if p.IncludeTypes {
-						pathMemoryAsType := &path.MemoryAsType{
-							Address: rng.Base,
-							Size:    rng.Size,
-							Pool:    p.Pool,
-							After:   p.After,
-							Type: &path.Type{
-								TypeIndex: id,
-								API:       &path.API{ID: path.NewID(apiId)},
-							},
-						}
-						value, err := MemoryAsType(ctx, pathMemoryAsType, rc)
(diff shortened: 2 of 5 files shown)


stellama0208:
Chris provided some suggestions offline: there're duplicated mutations in this implementation. Will have a new commit coming to address this issue.

stellama0208:
The duplicated mutations problem is solved in updated commit, ready for review! :)

gapis/resolve/memory.go
```diff
@@ -309,3 +320,73 @@ func MemoryAsType(ctx context.Context, p *path.MemoryAsType, rc *path.ResolveCon
 
 	return vals[0], nil
 }
+
+// Function memoryAsType resolves typed memory related raw data,
+// and return the decoded memory.
+// This is a private resolving function inside gapis,.
+func memoryAsType(ctx context.Context, s *api.GlobalState, rng memory.Range, poolId uint32, typeIndex uint64,
```
chrisforbes, line 327: What happens to the existing MemoryAsType function? Should it call this?

gapis/resolve/memory.go
```diff
@@ -309,3 +320,73 @@ func MemoryAsType(ctx context.Context, p *path.MemoryAsType, rc *path.ResolveCon
...
+	p path.Node, rc *path.ResolveConfig) (*memory_box.Value, error) {
+	// Check whether the requested pool was ever created.
+	pool, err := s.Memory.Get(memory.PoolID(poolId))
```
chrisforbes, line 330: Should we pass the pool from the outside rather than the id? We have it.

gapis/resolve/memory.go
stellama0208: done in the new commit.

gapis/resolve/memory.go
stellama0208: done in the new commit.

gapic/src/main/com/google/gapid/views/MemoryView.java
```diff
@@ -792,12 +792,15 @@ private void loadAndSetStructs(StructObservation[] structs) {
         packColumns(treeViewer.getTree());
         return;
       }
+      // long start = System.currentTimeMillis();
```
chrisforbes, line 795: Should probably drop this out of the final change.

stellama0208:
The above force push is only to rebase this PR on agi-tot. (to see whether the CI tests get fixed.)

gapis/resolve/memory.go
```diff
@@ -127,17 +127,7 @@ func Memory(ctx context.Context, p *path.Memory, rc *path.ResolveConfig) (*servi
...
-							Address: rng.Base,
-							Size:    rng.Size,
-							Pool:    p.Pool,
-							After:   p.After,
-							Type: &path.Type{
-								TypeIndex: id,
-								API:       &path.API{ID: path.NewID(apiId)},
-							},
-						}
-						value, err := MemoryAsType(ctx, pathMemoryAsType, rc)
```
purvisa-at-google-com: This commit is undoing a lot of the work of the previous commit. It's very hard to know what the cumulative effect of these 5 commits is going to be once they are squished. Is there anything we can do to make this clearer?

gapis/resolve/memory.go
```diff
@@ -339,3 +320,73 @@ func MemoryAsType(ctx context.Context, p *path.MemoryAsType, rc *path.ResolveCon
...
+	if sz == 0 {
+		sz = 0xFFFFFFFFFFFFFFFF
+	}
+
+	dec := s.MemoryDecoder(ctx, pool.Slice(memory.Range{
+		Base: rng.Base,
+		Size: sz,
+	}))
+
+	e, err := types.GetType(typeIndex)
```
purvisa-at-google-com, line 284: Can we change to longer names? It's not clear why the return of GetType is named "e". What does "e" stand for?

gapis/resolve/memory.go
```diff
@@ -256,69 +256,8 @@ func MemoryAsType(ctx context.Context, p *path.MemoryAsType, rc *path.ResolveCon
...
-
-	e, err := types.GetType(p.Type.TypeIndex)
-	if err != nil {
-		return nil, err
-	}
-
-	nElems := 1
-	elemSize := 0
-	isSlice := false
-	ty := e
```
purvisa-at-google-com: Why do we have this line? I prefer the name "ty" to "e", because "ty" probably means type. But we have the same value stored in two local variables with highly non descriptive names; and that just hurts readability.

gapis/resolve/memory.go
stellama0208: Fixed in new commit: Address comment #3.

gapis/resolve/memory.go
stellama0208: Fixed in new commit: Address comment #3.

gapis/resolve/memory.go
stellama0208: Hey Alan, I think the cumulative change given by github can make it clearer? (https://github.com/google/agi/pull/301/files)

Sorry for the deleting work in this commit, it was to remove the unnecessary path generation step, by directly transferring needed information.

purvisa-at-google-com: Hey, thanks Stella! Sorry, I was not aware of this github feature. It makes reviews like this much easier to read, thanks :)

stellama0208:
（Sorry, clicked both re-review by accident ;(

PR #309, opened 2020-05-28
Move command index to the bottom in command panel.

stellama0208:
- Give space to the more informative command name and parameters.
 - Bug: b/156548627.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/CommandStream.java
+++ b/gapic/src/main/com/google/gapid/models/CommandStream.java
@@ -41,6 +41,7 @@ import com.google.gapid.util.Loadable;
 import com.google.gapid.util.MoreFutures;
 import com.google.gapid.util.Paths;
 
+import com.google.gapid.views.Formatter;
 import org.eclipse.swt.widgets.Shell;
 
 import java.util.concurrent.ExecutionException;
@@ -374,6 +375,16 @@ public class CommandStream
     public CommandIndex getIndex() {
       return (data == null) ? null : CommandIndex.forNode(data.getRepresentation(),
           getPath(Path.CommandTreeNode.newBuilder()).build());
+    }
+
+    public String getIndexString() {
+      if (data == null) {
+        return "";
+      } else if (data.getGroup().isEmpty() && data.hasCommands()) {
+        return Formatter.lastIndex(data.getCommands());
+      } else {
+        return Formatter.firstIndex(data.getCommands());
+      }
     }
 
     public ListenableFuture<Node> load(Shell shell, Supplier<ListenableFuture<NodeData>> loader) {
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/models/CommandStream.java
```diff
@@ -376,6 +377,16 @@ public CommandIndex getIndex() {
           getPath(Path.CommandTreeNode.newBuilder()).build());
     }
 
+    public String getIndexString() {
+      if (data == null) {
+        return "";
```
pau-baiget, line 382: Is it safe to return an empty string? Shouldn't we return null here (similar to function above)?

gapic/src/main/com/google/gapid/views/CommandTree.java
```diff
@@ -96,9 +102,17 @@ public CommandTree(Composite parent, Models models, Widgets widgets) {
     SearchBox search = new SearchBox(this, false);
     loading = LoadablePanel.create(this, widgets, p -> new Tree(p, models, widgets));
     tree = loading.getContents();
+    commandIdx = createLabel(this, COMMAND_INDEX_DSCRP);
+    commandIdx.setToolTipText(COMMAND_INDEX_HOVER);
+    commandIdx.addListener(SWT.MouseDoubleClick, e -> {
+      if (commandIdx.getText().length() > COMMAND_INDEX_DSCRP.length()) {
+        widgets.copypaste.setContents(commandIdx.getText().substring(COMMAND_INDEX_DSCRP.length()));
```
pau-baiget, line 109: Is the substring necessary? It was created with COMMAND_INDEX_DSCRP just a few lines before.

gapic/src/main/com/google/gapid/models/CommandStream.java
stellama0208: yeah it would be safe, usually for string in java an empty one would be safer than a null.

gapic/src/main/com/google/gapid/views/CommandTree.java
stellama0208: Yes it would be necessary, it will cut out the description part in the label string.
e.g. for a label string of `Command index: 123.0.1`, after using substring to clip it, we only send `123.0.1` to user's clipboard, so it's more ready to be pasted in the goto-command dialog.

gapic/src/main/com/google/gapid/views/CommandTree.java
```diff
@@ -80,10 +83,13 @@
 public class CommandTree extends Composite
     implements Tab, Capture.Listener, CommandStream.Listener {
   protected static final Logger LOG = Logger.getLogger(CommandTree.class.getName());
+  private static final String COMMAND_INDEX_HOVER = "Double click to copy index. Use Ctrl+G to do command jumping.";
+  private static final String COMMAND_INDEX_DSCRP = "Command index: ";
```
hevrard, line 87: "Command jumping" may not be the most understandable to our users, maybe it would be more understandable with something like: "Use Ctrl+G to jump to a given command index." ?

what do you think?

stellama0208: Thanks! That sounds good, updated in a new commit. :)

Issue #60, opened 2020-02-18
Upgrade grpc to avoid gettid problems


mattyclarkson:
gRPC has provided a [PR](https://github.com/grpc/grpc/pull/18950) that fixes duplicate `gettid` declarations on `glibc-2.30+` distributions.

The PR landed on 2019-08-26 and the first release after that was [`v1.24.0`](https://github.com/grpc/grpc/releases/tag/v1.24.0) on 2019-09-25.

We should [roll up](https://github.com/google/agi/blob/ca386ff2eb30ecefe3f5138ebfec7e63ef1d7710/tools/build/workspace.bzl#L72) to at least `v1.24.0`. Ideally, newer.

hevrard:
Thanks @mattyclarkson, is this a blocking issue on your side? (just so that we can prioritize accordingly)

mattyclarkson:
I can't build AGI, so yes. I've tried upgrading on my side but stuck trying to resolve `@upd` and `@build_bazel_rules_apple`.

gRPC provides [`bazel/grpc_deps.bzl`](https://github.com/grpc/grpc/blob/master/bazel/grpc_deps.bzl) which finds all the necessary dependencies. Why do we not use that?

mattyclarkson:
OK, have worked around it by disabling the `gettid` delcaration. No rush.

We should upgrade the dependency though. I found it especially hard to upgrade as AGI dependencies don't seem to use the upstream project `get_deps` functions. It would make it much easier to upgrade if we changed to use those.

hevrard:
Thanks for the pointers, looking into using bazel/grpc_deps.bzl now.

Bazel is a (very) fast-moving target, we probably don't use the get_deps because they were not available when we created our own bazel rules.

mattyclarkson:
Makes sense. Thanks for the updates.

It seems strange there isn't a standard Bazel function where you declare dependencies and it automatically resolves them. Most modern build systems support that now. Seems that most projects manually write a `*_deps.bzl` which seems a bit boilerplate-y

hevrard:
@johnplate had a look and unfortunately, there seems to be breaking changes.

We plan to upgrade, but later this year.

johnplate:
Thank you @mattyclarkson for your report! I fully support your request to use at least gRPC 1.24 and grpc_deps.bzl. With our upcoming presentations at the GDC conference it would be too risky to change gRPC now, but I will take care of it soon after.

mattyclarkson:
[...]

PR #310, opened 2020-05-29
Update grpc to 1.29.1

hevrard:
Needed to fix gcc-9.2.1-31 compilation issue:

external/com_github_grpc_grpc/src/core/lib/iomgr/ev_epollex_linux.cc:1105:13: error: ambiguating new declaration of 'long int gettid()'

The fix on grpc side is: https://github.com/grpc/grpc/commit/de6255941a5e1c2fb2d50e57f84e38c09f45023d
Update to the latest grpc to get that fix in.


hevrard:
Fix #60

gapir/cc/grpc_replay_service.h
```diff
@@ -39,8 +37,8 @@ class ReplayResponse;
 namespace gapir {
 
 using ReplayGrpcStream =
-    grpc::ServerReaderWriter<replay_service::ReplayResponse,
-                             replay_service::ReplayRequest>;
+    grpc_impl::ServerReaderWriter<replay_service::ReplayResponse,
```
pau-baiget, line 40: Why is this needed? In the fix description you talk about upgrading the gRPC version but i don't see why you need to call the _impl directly here.

purvisa-at-google-com: Guess this goes for the changed include/declaration above too.

hevrard: This change is needed to be able to compile, the ServerReaderWriter template has been renamed in the new version.

Basically all the changes of this PR are there to be able to pass build with the new grpc version.

hevrard:
I'm not sure why the build fails on mac/win.

@pau-baiget could you try to build this PR on mac & windows?

WORKSPACE
```diff
@@ -33,5 +33,11 @@ load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
 go_rules_dependencies()
 go_register_toolchains()
 
+load("@com_github_grpc_grpc//bazel:grpc_deps.bzl", "grpc_deps", "grpc_test_only_deps")
```
pau-baiget, line 36: And why is this needed if it wasn't needed before? Just curiosity.

pau-baiget: Actually something more than curiosity: the build failures in Win/Mac are found in a dependency of grpc, probably loaded here.

tobine:
I verified that this fixes the build fail I was seeing on Linux

WORKSPACE
hevrard: I had to add these to resolve grpc dependency issues.

See: https://github.com/grpc/grpc/blob/560577371abff8f504091d906ddaf58126a07762/bazel/grpc_extra_deps.bzl#L10-L27

But I guess we should be able to remove the `grpc_test_only_deps`, can you try this?

silence-do-good: I checked it locally and it works. The build seems to finish successfully without the test_only_deps

silence-do-good: However, I have only a linux machine where I can build. For some reason, bazel doesn't run on my Mac, so I'm not able to verify on any other platform.

hevrard: @silence-do-good thanks for checking, I removed the `grpc_test_only_deps`

hevrard: @pau-baiget I also added the link above as a comment.

hevrard:
This is basically blocked on https://github.com/abseil/abseil-cpp/issues/326.

Recap:
- this grpc update is motivated to compile with a new version of gcc9
- the update transitively depends on abseil which has an open issue that prevents to build on macOS: https://github.com/abseil/abseil-cpp/issues/326
- to work around this, other projects patch abseil as part of their build (e.g. tensorflow, see https://github.com/tensorflow/tensorflow/commit/04e169ac74a81b00507f1bb571513a5c2f61547f); we don't want to go this way
- we decide to use gcc8 on all builds for now, see b/157873397

Closing, we may reuse this once GRPC depends on a version of Abseil that has a fix for the macOS builds.

pau-baiget:
Added a couple commits to patch the dependency (abseil-cpp) that's causing the build failure on macOS. The reason we're patching instead of forcing gcc-8 everywhere is that in Linux we can't enforce an older glibc (the problems happen from glibc 2.30). Once the https://github.com/abseil/abseil-cpp/issues/326 is fixed then we can remove the patch.

hevrard:
#319 is our fix for now. This PR here might come handy when we eventually want to upgrade grpc.

PR #312, opened 2020-06-01
Keep the same GCC version (8.x) everywhere

pau-baiget:
This PR forces GCC 8.x on Windows builds (Linux already uses gcc8 and MacOS uses clang via XCode). The dependencies' versions are chosen based on what pacman chose in my local checkout after executing. 

```
pacman -U mingw-w64-x86_64-gcc*-8.2.0-3-any.pkg.tar.xz
```

Bug: b/157873397

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -55,9 +55,9 @@ Using the msys64 shell (installed by choco at `C:\tools\msys64\mingw64`):
 1. Update MSYS with: `pacman -Syu`.
 2. If the update ends with “close the window and run it again”, close and reopen the window and repeat 1.
 3. Fetch required tools with: `pacman -S curl git zip unzip patch`
-4. Download gcc with: `curl -O http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-gcc-9.2.0-2-any.pkg.tar.xz`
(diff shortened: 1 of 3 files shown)


hevrard:
Build is failing on Windows.

I don't think you should update packages other than gcc, i.e. binutils etc should stay the same I think.

Could you try with changing only the gcc and gcc-libs packages version to 8.2.0-3?

(This seems to be the highest gcc 8 version listed in http://repo.msys2.org/mingw/x86_64/)

hevrard:
Also, I think we want to impose gcc8 on Linux, and the Build documentation should reflect this.

This can be done by something like (to test):
```
sudo apt-get install gcc-8 g++-8
export CC=gcc-8
export CXX=g++-8
```

BUILDING.md
```diff
@@ -210,6 +210,11 @@ cd <sdk-path>
 tools/bin/sdkmanager platform-tools
 ```
 
+### Install gcc (required version is 8)
+sudo apt-get install gcc-8 g++-8
+export CC=gcc-8
+export CXX=g++-8
```
hevrard, line 216: Surround this with triple backquotes. Have you checked that this works on glinux?

pau-baiget:
After https://github.com/google/agi/pull/319 we're keeping gcc 9.x in some platforms so we'll need to do some work to make sure that this version is everywhere. We'll open another PR when that's ready.

PR #314, opened 2020-06-01
Descriptor memory layout

chrisforbes:


kocdemir:
I don't think this needs merging anymore, closing.

PR #315, opened 2020-06-02
Add support for counter groups

rschiu:
Perfetto is updated with counter groups.
Change to UI to group counters if GpuCounterDescriptor has group numbers.

### DIFF
--- a/core/os/device/gpu_counter_descriptor.proto
+++ b/core/os/device/gpu_counter_descriptor.proto
@@ -29,6 +29,18 @@ option go_package = "github.com/google/gapid/core/os/device";
 // This message is sent by a GPU counter producer to specify the counters
 // available in the hardware.
 message GpuCounterDescriptor {
+  // Logical groups for a counter.  This is used in the UI to present the
+  // related counters together.
+  enum GpuCounterGroup {
+    UNCLASSIFIED = 0;
+    SYSTEM = 1;
+    VERTICES = 2;
+    FRAGMENTS = 3;
+    PRIMITIVES = 4;
+    MEMORY = 5;  // Includes counters relating to caching and bandwidth.
+    COMPUTE = 6;
+  };
+
   message GpuCounterSpec {
     uint32 counter_id = 1;
     string name = 2;
@@ -37,10 +49,11 @@ message GpuCounterDescriptor {
     oneof peak_value {
       int64 int_peak_value = 5;
       double double_peak_value = 6;
-    };
+    }
     repeated MeasureUnit numerator_units = 7;
     repeated MeasureUnit denominator_units = 8;
     bool select_by_default = 9;
+    repeated GpuCounterGroup groups = 10;
   }
   repeated GpuCounterSpec specs = 1;
 
@@ -50,7 +63,7 @@ message GpuCounterDescriptor {
   message GpuCounterBlock {
     // required. Unique ID for the counter group.
     uint32 block_id = 1;
-    // Number of counters supported by the block. No limit if unset.
+    // optional. Number of counters supported by the block. No limit if unset.
     uint32 block_capacity = 2;
     // optional. Name of block.
     string name = 3;
@@ -73,6 +86,7 @@ message GpuCounterDescriptor {
   // command buffer.
   bool supports_instrumented_sampling = 5;
 
+  // next id: 41
   enum MeasureUnit {
     NONE = 0;
 
--- a/gapic/src/main/com/google/gapid/models/Perfetto.java
+++ b/gapic/src/main/com/google/gapid/models/Perfetto.java
@@ -184,13 +184,14 @@ public class Perfetto extends ModelBase<Perfetto.Data, Path.Capture, Loadable.Me
     public final GpuInfo gpu;
     public final FrameInfo frame;
     public final ImmutableMap<Long, CounterInfo> counters;
+    public final ImmutableListMultimap<Long, Long> gpuCounterGroups;
     public final VSync vsync;
     public final TrackConfig tracks;
 
     public Data(QueryEngine queries, TimeSpan traceTime, CpuInfo cpu,
         ImmutableMap<Long, ProcessInfo> processes, ImmutableMap<Long, ThreadInfo> threads,
-        GpuInfo gpu, FrameInfo frame, ImmutableMap<Long, CounterInfo> counters, VSync vsync,
-        TrackConfig tracks) {
+        GpuInfo gpu, FrameInfo frame, ImmutableMap<Long, CounterInfo> counters,
+        ImmutableListMultimap<Long, Long> gpuCounterGroups, VSync vsync, TrackConfig tracks) {
       this.qe = queries;
       this.traceTime = traceTime;
       this.cpu = cpu;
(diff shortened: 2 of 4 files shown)


rschiu:
Pascal or Stella, could you take a look?  Thanks.

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
```diff
@@ -45,6 +46,9 @@
       "where (value + 1 > value or value - 1 < value) " +
       "group by ct.id";
 
+  private static final String LIST_COUNTER_GROUP_SQL =
+      "select group_id, track_id from gpu_counter_group";
+
```
pmuetschard, line 51: Why not join with the group table in the above list query and make the group_id a field of the `CounterInfo`?

gapic/src/main/com/google/gapid/models/Perfetto.java
```diff
@@ -95,8 +95,9 @@ public void onCaptureLoaded(Loadable.Message error) {
             transformAsync(withStatus("Querying GPU info...", queryGpu(data)), $3 ->
               transformAsync(withStatus("Querying Frame info...", queryFrame(data)), $4 ->
                 transformAsync(withStatus("Querying counters...", queryCounters(data)), $5 ->
-                  transform(withStatus("Enumerating tracks...", enumerateTracks(data)), $6 ->
-                    data.build()))))));
+                  transformAsync(withStatus("Querying counter groups...", queryCounterGroups(data)), $6 ->
```
pmuetschard, line 98: I don't think this needs it's extra section. Getting the group information should just be part of the counters.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -157,11 +174,38 @@ private Tracks() {
...
+          parent = "gpu_counters_group_" + group.name();
+          String name = GPU_COUNTER_GROUP_NAMES.get(group);
+          data.tracks.addLabelGroup("gpu_counters", parent, name,
+              group(state -> new TitlePanel(name), true));
+          for (Long counter_id : grouped_counters) {
+            CounterInfo counter = counters.get(counter_id);
+            if (counter == null) {
+              continue;
+            }
+            CounterTrack track = new CounterTrack(data.qe, counter);
```
pmuetschard, line 196: I just realized that the way it's written now, a counter can be in more than one group. You can only instantiate a `CounterTrack` for each counter once. If we want a counter to show up multiple times, we can have multiple `CounterPanel` instances, but they all need to share the same `CounterTrack` instance.

silence-do-good:
Please check the perfetto version once more, before merging this. I've merged the frame lifecycle changes which updated perfetto.

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
rschiu: A counter can be in multiple groups.  The counter group map make it easier to add the track in the correct order.  It's either we build it here or later in Tracks.enumerateGpu.
If we are building the table here, it wouldn't make sense to join with the list_sql query, would it?

gapic/src/main/com/google/gapid/models/Perfetto.java
rschiu: Done.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
rschiu: Done

rschiu:
Pascal, could you please take another look?  Thanks.

core/os/device/gpu_counter_descriptor.proto
```diff
@@ -37,10 +49,11 @@ message GpuCounterDescriptor {
     oneof peak_value {
       int64 int_peak_value = 5;
       double double_peak_value = 6;
-    };
+    }
```
stellama0208, line 52: Is there a reason we delete the semicolon here?

gapic/src/main/com/google/gapid/models/Perfetto.java
```diff
@@ -122,6 +122,10 @@ public void onCaptureLoaded(Loadable.Message error) {
     return CounterInfo.listCounters(data);
   }
 
+  private static ListenableFuture<Data.Builder> queryCounterGroups(Data.Builder data) {
+    return CounterInfo.listCounterGroups(data);
+  }
+
```
pmuetschard, line 128: This is no longer called, is it?

gapic/src/main/com/google/gapid/models/Perfetto.java
```diff
@@ -184,13 +188,14 @@ protected void fireLoadedEvent() {
     public final GpuInfo gpu;
     public final FrameInfo frame;
     public final ImmutableMap<Long, CounterInfo> counters;
+    public final ImmutableListMultimap<Long, Long> gpuCounterGroups;
```
pmuetschard, line 191: nit: Not a big fan of this in such a "global" place, I'd prefer some structure that combines this with the counter info. Mainly because it's not typed, just a map of long-to-longs..., but eh, am ok.

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
pmuetschard: Yeah, when I wrote this, I hadn't realized that counters can be in multiple groups.

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
```diff
@@ -169,10 +174,21 @@ public static Unit unitFromString(String unit) {
...
+        listCounterGroups(data),
+        $1 -> transform(
+          data.qe.query(LIST_SQL), res -> {
+            ImmutableMap.Builder<Long, CounterInfo> counters = ImmutableMap.builder();
+            res.forEachRow((i, r) -> counters.put(r.getLong(0), new CounterInfo(r)));
+            return data.setCounters(counters.build());
+          }));
+  }
+
+  public static ListenableFuture<Perfetto.Data.Builder> listCounterGroups(Perfetto.Data.Builder data) {
```
pmuetschard, line 187: This should be private as it's called by `listCounters`.

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
```diff
@@ -169,10 +174,21 @@ public static Unit unitFromString(String unit) {
...
-      res.forEachRow((i, r) -> counters.put(r.getLong(0), new CounterInfo(r)));
-      return data.setCounters(counters.build());
+    return transformAsync(
+        listCounterGroups(data),
+        $1 -> transform(
+          data.qe.query(LIST_SQL), res -> {
+            ImmutableMap.Builder<Long, CounterInfo> counters = ImmutableMap.builder();
+            res.forEachRow((i, r) -> counters.put(r.getLong(0), new CounterInfo(r)));
+            return data.setCounters(counters.build());
+          }));
```
pmuetschard, line 184:
```suggestion
    return transformAsync(listCounterGroups(data), $1 ->
      transform(data.qe.query(LIST_SQL), res -> {
        ImmutableMap.Builder<Long, CounterInfo> counters = ImmutableMap.builder();
        res.forEachRow((i, r) -> counters.put(r.getLong(0), new CounterInfo(r)));
        return data.setCounters(counters.build());
      }));
```

core/os/device/gpu_counter_descriptor.proto
rschiu: This is a copy of the file in Perfetto source code.  We shouldn't add a semicolon if it's not necessary.

PR #317, opened 2020-06-04
Frame lifecycle UI Updates

silence-do-good:
We have decided to revamp the frame lifecycle UI to be friendly towards users who don't have the knowledge of SurfaceFlinger's events. This new change brings in an ownership view where the old events are taken and are shown as slices of DISPLAY (Present to Present), APP (Dequeue to Queue), Wait for GPU (Queue to Acquire), Composition (Latch to Present). The old buffer events will still be made available so as to not block experienced users.

This change also removes the old summary view as it is no longer relevant and there isn't a design yet to show some kind of a summary.

Design doc: go/sf-profiling-design
    
 Bug: 155242423

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
@@ -94,8 +94,7 @@ public class FrameEventsTrack extends Track.WithQueryEngine<FrameEventsTrack.Dat
   @Override
   public ListenableFuture<Data> computeData(DataRequest req) {
     Window window = Window.compute(req, 5);
-    return transformAsync(window.update(qe, tableName("window")),
-        $ -> computeSlices(req));
+    return transformAsync(window.update(qe, tableName("window")), $ -> computeSlices(req));
   }
 
   private ListenableFuture<Data> computeSlices(DataRequest req) {
@@ -197,11 +196,6 @@ public class FrameEventsTrack extends Track.WithQueryEngine<FrameEventsTrack.Dat
 
     public Slices(QueryEngine.Row row, ArgSet args, String layerName, String eventName) {
       add(row.getLong(0), row.getLong(1), row.getLong(2), row.getString(3), args,
-          row.getLong(8), layerName, eventName, row.getLong(9), row.getLong(10), row.getLong(11));
-    }
-
-    public Slices(QueryEngine.Row row, String layerName, String eventName) {
-      add(row.getLong(0), row.getLong(1), row.getLong(2), row.getString(3), ArgSet.EMPTY,
           row.getLong(8), layerName, eventName, row.getLong(9), row.getLong(10), row.getLong(11));
     }
 
--- a/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
@@ -16,21 +16,16 @@
 
 package com.google.gapid.perfetto.models;
 
-import static com.google.gapid.perfetto.models.QueryEngine.createSpan;
-import static com.google.gapid.perfetto.models.QueryEngine.createView;
 import static com.google.gapid.perfetto.models.QueryEngine.createWindow;
 import static com.google.gapid.perfetto.models.QueryEngine.dropTable;
-import static com.google.gapid.perfetto.models.QueryEngine.dropView;
 import static com.google.gapid.perfetto.models.QueryEngine.expectOneRow;
 import static com.google.gapid.util.MoreFutures.transform;
 import static com.google.gapid.util.MoreFutures.transformAsync;
 import static java.lang.String.format;
-import static java.util.concurrent.TimeUnit.MICROSECONDS;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
-import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.gapid.perfetto.TimeSpan;
 import com.google.gapid.perfetto.views.FrameEventsSelectionView;
@@ -38,9 +33,7 @@ import com.google.gapid.perfetto.views.State;
 
 import org.eclipse.swt.widgets.Composite;
 
-import java.util.Arrays;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.function.Consumer;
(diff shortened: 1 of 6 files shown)


gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
```diff
@@ -31,70 +37,307 @@
...
-      "order by t.id";
-  private static final String DISPLAYED_FRAME_TRACK_NAME = "Displayed Frame";
+  private static final String LAYER_NAME_QUERY =
+      "select distinct layer_name from frame_slice";
+  private static final String LAYER_TRACKS_QUERY =
+      "select name, track_id from gpu_track inner join " +
+      "(select distinct track_id from frame_slice where layer_name = '%s') t" +
+      " on gpu_track.id = t.track_id";
+  private static final String TRACK_ID_QUERY =
+      "select group_concat(track_id) from (%s) where name GLOB '%s'";
```
rschiu, line 47: These name globbing bothers me.  It's alright for this CL, but remind me to talk with you about this.

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
```diff
@@ -31,70 +37,307 @@
...
+        }
+        return layers;
+      });
+    });
+  }
+
+  private static ListenableFuture<List<List<Event>>> createPhases(QueryEngine qe,
+      List<List<Event>> phases, List<String> layerNames, int idx) {
+    String layerName = layerNames.get(idx);
+    String baseName = layerName.replaceAll("[^A-Za-z0-9]", "");
```
rschiu, line 154: What happens if you do not filter out non-alphanumeric characters?

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
```diff
@@ -31,70 +37,307 @@
...
-  private FrameInfo(List<Buffer> buffers) {
-    this.buffers = buffers;
-  }
+  // Cast name as int for the phases. This is needed because experimental_slice_layout generates
+  // columns only based on slice_table. frame_slice is a child of slice_table and the extra columns
+  // such as frame_number and layer_name won't be generated. The alternate is doing a join on
+  // frame_slice table with slice_id but that query is very expensive in that it visibly slows down
+  // the loading of tracks.
+  private static final String PHASE_QUERY =
+      "select phase.*, cast(NAME as INT) as frame_number from " +
```
rschiu, line 60: s/NAME/name/

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
```diff
@@ -31,70 +37,307 @@
...
+  // same time.
 
-  private FrameInfo(List<Buffer> buffers) {
-    this.buffers = buffers;
-  }
+  // Cast name as int for the phases. This is needed because experimental_slice_layout generates
+  // columns only based on slice_table. frame_slice is a child of slice_table and the extra columns
+  // such as frame_number and layer_name won't be generated. The alternate is doing a join on
+  // frame_slice table with slice_id but that query is very expensive in that it visibly slows down
+  // the loading of tracks.
```
rschiu, line 58: Ugh....we need to figure out how to fix this.  This is a recipe for disaster.  Someone is going to change the name some day and break things because they did not know/forgot the name has to be the frame number

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -51,200 +44,116 @@
...
+        createWindow(window));
   }
 
   @Override
   public ListenableFuture<Data> computeData(DataRequest req) {
-    Window window = Window.compute(req, 5, FRAMELIFECYCLE_QUANTIZE_CUTOFF);
+    Window window = Window.compute(req, 5);
     return transformAsync(window.update(qe, tableName("window")),
-        $ -> window.quantized ? computeSummary(req, window) : computeSlices(req));
+        $ -> computeSlices(req));
```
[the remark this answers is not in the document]
silence-do-good: Quantized is needed only for summary and the summary we had was both wrong/not useful. So I removed summary altogether.

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
silence-do-good: The layer names will usually have lots of dots(.) in them and sql will treat them as databases. For example, com.app.layer1 - com will be treated as a db and app as a table in it and so on.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -51,200 +44,116 @@
...
+  private static final String STAT_COLUMNS =
+      "queue_to_acquire_time, acquire_to_latch_time, latch_to_present_time";
   private static final String SLICE_SQL =
-      "select " + BASE_COLUMNS + " from frame_slice where id = %d";
+      "select * from " +
+      "(select " + BASE_COLUMNS + " from frame_slice where id = %d) join " +
+      "(select " + STAT_COLUMNS + " from frame_slice " +
+        "where name = 'PresentFenceSignaled' and frame_number = " +
+          "(select frame_number from frame_slice where id = %d))";
   private static final String SLICES_SQL =
```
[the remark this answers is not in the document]
silence-do-good: Slices sql is for gathering the data. However, you point is correct. The RANGE_SQL below that handles range selection and I've updated the query to consider this as well. PTAL

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
silence-do-good: Added a TODO for this.

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
```diff
@@ -31,70 +37,308 @@
...
+                  currentLayerPhases.add(new Event("Composition", baseName + "_COMPOSITION",
+                      compositionDepth + 1, COMPOSITION_TOOLTIP));
+                  phases.add(currentLayerPhases);
+                  if (idx + 1 >= layerNames.size()) {
+                    return Futures.immediateFuture(phases);
+                  }
+                  return createPhases(qe, phases, layerNames, idx + 1);
+                });
+              });
             });
```
stellama0208, line 208: I'm a little worried about the 9 recursive transform function calling here, if they don't have time dependency, can we use some utility methods like `Futures.allAsList` to make it iterative rather than recursive? Not extremely sure, may need some suggestions from Pascal here.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsSelectionView.java
```diff
@@ -74,36 +71,23 @@ private void setSingleSliceView(State state, FrameEventsTrack.Slices slice) {
...
-    }
+    Composite stats = withLayoutData(createComposite(this, new GridLayout(2, false)),
+        withIndents(new GridData(SWT.LEFT, SWT.TOP, false, false), PANEL_INDENT, 0));
+    withLayoutData(createBoldLabel(stats, "Frame Stats:"),
+        withSpans(new GridData(), 2, 1));
+
+    createLabel(stats, "Frame number: ");
+    createLabel(stats, Long.toString(slice.frameNumbers.get(0)));
+
+    createLabel(stats, "Queue to Acquire: ");
```
stellama0208, line 83: Maybe also add descriptive string like Display, Application, Wait for GPU, Composition here, to help ramp up the users and let them know the track's meaning?

silence-do-good: The stats do not exactly completely to the 4 tracks that we are showing. There is some overlap, like queue to acquire and wait for GPU while the latch delay is not covered in the tracks. These stats are meant for those who know about the instant events and don't want to expand the buffers track fully to check on them.

gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -253,60 +180,53 @@ public Data(DataRequest request, long[] ids, long[] starts, long[] ends,
...
       add(row.getLong(0), row.getLong(1), row.getLong(2), row.getString(3), args,
-          row.getString(8), row.getString(9), frameStats);
+          row.getLong(8), layerName, eventName, row.getLong(9), row.getLong(10), row.getLong(11));
     }
 
-    public Slices(QueryEngine.Row row, Map<String, FrameStats> frameStats) {
+    public Slices(QueryEngine.Row row, String layerName, String eventName) {
       add(row.getLong(0), row.getLong(1), row.getLong(2), row.getString(3), ArgSet.EMPTY,
-          row.getString(8), row.getString(9), frameStats);
+          row.getLong(8), layerName, eventName, row.getLong(9), row.getLong(10), row.getLong(11));
```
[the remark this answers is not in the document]
silence-do-good: So I removed the second method since it was only being used previously.  The method below this is actually used for an entire query result, rather than a row. Technically, from the caller's perspective, there is no args that can be applied for multiple rows. So, the ArgSet.EMPTY should be applied in the constructor rather than sending it from the caller.

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
```diff
@@ -31,70 +37,308 @@
...
+    return transformAsync(data.qe.query(LAYER_NAME_QUERY), r -> {
+      r.forEachRow(($, row) -> {
+        layerNames.add(row.getString(0));
+      });
+      if (layerNames.isEmpty()) {
+        data.setFrame(FrameInfo.NONE);
+        return Futures.immediateFuture(data);
+      }
+      return transform(getLayers(data.qe, layers, layerNames), $ -> {
+        data.setFrame(new FrameInfo(layers));
```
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/perfetto/models/FrameInfo.java
silence-do-good: Good suggestion. I did not know about allAsList. I've updated the function now to use this. PTAL

PR #319, opened 2020-06-04
Add patch to fix gRPC compilation issue with gcc-9.2.1-31

pau-baiget:
Updating gRPC is turning to be a bit of a nightmare (https://github.com/google/agi/pull/310) so instead i'm testing whether we can just patch the existing gRPC dependency with the known fix from grpc/grpc@de62559 (and a similar fix needed somewhere else) and fix the linux build issues with gcc-9.2.1-31.


tobine:
I confirmed that this PR fixes my build issues on Linux

tools/build/third_party/com_github_grpc_grpc/com_github_grpc_grpc_fix.patch
```diff
@@ -0,0 +1,61 @@
...
+diff --git a/src/core/lib/gpr/log_linux.cc b/src/core/lib/gpr/log_linux.cc
+index 561276f0c2..81026e5689 100644
+--- a/src/core/lib/gpr/log_linux.cc
++++ b/src/core/lib/gpr/log_linux.cc
+@@ -40,7 +40,9 @@
+ #include <time.h>
+ #include <unistd.h>
+ 
+-static long gettid(void) { return syscall(__NR_gettid); }
++// Not naming it as gettid() to avoid duplicate declarations when complied with
```
hevrard, line 20: complied -> compiled

tools/build/third_party/com_github_grpc_grpc/com_github_grpc_grpc_fix.patch
```diff
@@ -0,0 +1,61 @@
...
+diff --git a/src/core/lib/iomgr/ev_epollex_linux.cc b/src/core/lib/iomgr/ev_epollex_linux.cc
+index c2d80c08dd..991c061219 100644
+--- a/src/core/lib/iomgr/ev_epollex_linux.cc
++++ b/src/core/lib/iomgr/ev_epollex_linux.cc
+@@ -1077,7 +1077,9 @@ static void end_worker(grpc_pollset* pollset, grpc_pollset_worker* worker,
+ }
+ 
+ #ifndef NDEBUG
+-static long gettid(void) { return syscall(__NR_gettid); }
++// Not naming it as gettid() to avoid duplicate declarations when complied with
```
hevrard, line 44: idem

tools/build/workspace.bzl
```diff
@@ -71,6 +71,7 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
         project = "grpc",
         commit = "c599e6a922a80e40e24a2d3c994a6dd51046796b",  # 1.22.1
         sha256 = "d17ead923510b3c8a03eec623fffe4cba64d43e10b3695f027a1c8f10c03756a",
+        patch_file = "@gapid//tools/build/third_party/com_github_grpc_grpc:com_github_grpc_grpc_fix.patch",
```
hevrard, line 74: Here add a comprehensive comment on why we do this, e.g. (please double check that I'm not saying wrong things in this):

```
This patch works around a naming conflict in grpc which leads to
compilation issues in recent gcc/glibc. This issue is fixed on recent
grpc versions (since
https://github.com/grpc/grpc/commit/de6255941a5e1c2fb2d50e57f84e38c09f45023d),
but updating our grpc version leads to errors in compiling the abseil
dependency of grpc
(https://github.com/abseil/abseil-cpp/issues/326). We tried to pull
abseil ourselves and patch it, but abseil also fails to compile with
gcc on windows, so we choose to patch grpc directly. Once grpc has a
version that builds fine on all our targets, we can update grpc and
drop this patch.
```

PR #320, opened 2020-06-04
Add OpenGL on ANGLE tracing option

tobine:
Add new tracing mode "OpenGL on ANGLE."
This is the Vulkan frame capture mode, but is intended to be used
on OpenGL apps and, prior to launching the app, ANGLE will be enabled
for the target app. AGI will then capture the Vulkan output from
ANGLE.

Here are the main components of the feature:
 * Enable feature w/ "--experimental-enable-angle-tracing" flags/experimental
 * This allows "OpenGL on Angle" tracing mode to appear
 * Detect ANGLE apk installed on device, favoring custom apk, then
 system ANGLE install
 * When Angle mode is selected, enable Angle for selected application
 * Frame capture proceeds as usual w/ Vulkan output of Angle
 * When complete, restore any original ANGLE settings to device

Bug: b/153447218

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -373,6 +373,58 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
 	return result, nil
 }
 
+// Currently using Android Q/10, API 29 as ANGLE support cut-off
+func (b *binding) SupportsAngle(ctx context.Context) bool {
+	os := b.Instance().GetConfiguration().GetOS()
+	return os.GetAPIVersion() >= 29
+}
+
+func (b *binding) QueryAnglePackageName(ctx context.Context) (string, error) {
+	if !b.SupportsAngle(ctx) {
+		return "", fmt.Errorf("ANGLE not supported on this device")
+	}
+	// ANGLE supported, so check for installed ANGLE package
+	// Favor custom installed package first, followed by default system package
+	custom := "com.chromium.angle"
+	system := "com.google.android.angle"
+	// Check installed packages for ANGLE package
+	packages, _ := b.InstalledPackages(ctx)
+	if pkg := packages.FindByName(custom); pkg != nil {
+		return custom, nil
+	}
+	if pkg := packages.FindByName(system); pkg != nil {
+		return system, nil
+	}
+	return "", fmt.Errorf("No ANGLE packages installed on this device")
+}
+
+func SetupAngle(ctx context.Context, d Device, p *android.InstalledPackage) (app.Cleanup, error) {
+	// Restore ANGLE settings during app cleanup
+	angleDriverValues, err := d.SystemSetting(ctx, "global", "angle_gl_driver_selection_values")
+	if err != nil {
+		return nil, log.Err(ctx, err, "Failed to get original ANGLE driver selection name.")
+	}
+	angleDriverPkgs, err := d.SystemSetting(ctx, "global", "angle_gl_driver_selection_pkgs")
+	if err != nil {
+		return nil, log.Err(ctx, err, "Failed to get original ANGLE enabled packages.")
+	}
+	anglePackage, err := d.SystemSetting(ctx, "global", "angle_debug_package")
+	if err != nil {
+		return nil, log.Err(ctx, err, "Failed to get original ANGLE package name.")
+	}
+	log.I(ctx, "Saved original ANGLE pkg %s for app %s to restore during cleanup.", anglePackage, angleDriverPkgs)
+	// We successfully saved old settings, now set new ANGLE values for tracing
+	d.SetSystemSetting(ctx, "global", "angle_gl_driver_selection_values", "angle")
+	d.SetSystemSetting(ctx, "global", "angle_debug_package", anglePackage)
+	d.SetSystemSetting(ctx, "global", "angle_gl_driver_selection_pkgs", p.Name)
+	// Return cleanup function to restore original ANGLE settings
+	return func(ctx context.Context) {
+		d.SetSystemSetting(ctx, "global", "angle_gl_driver_selection_values", angleDriverValues)
+		d.SetSystemSetting(ctx, "global", "angle_gl_driver_selection_pkgs", angleDriverPkgs)
+		d.SetSystemSetting(ctx, "global", "angle_debug_package", anglePackage)
+	}, nil
+}
+
 func extrasFlags(extras []android.ActionExtra) []string {
 	flags := []string{}
 	for _, e := range extras {
--- a/core/os/device/device.proto
+++ b/core/os/device/device.proto
@@ -173,6 +173,8 @@ message Configuration {
   Drivers drivers = 4;
   // Perfetto capability.
   PerfettoCapability perfetto_capability = 5;
+  // Angle package name
+  string angle_package = 6;
 }
 
 message VulkanProfilingLayers {
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -271,6 +271,7 @@ public class TracerDialog {
       private static final String DEFAULT_TRACE_FILE = "trace";
       private static final String TRACE_EXTENSION = ".gfxtrace";
       private static final String PERFETTO_EXTENSION = ".perfetto";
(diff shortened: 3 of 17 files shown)


gapis/trace/android/trace.go
```diff
@@ -283,6 +283,10 @@ func (t *androidTracer) TraceConfiguration(ctx context.Context) (*service.Device
 	*/
 	if len(t.b.Instance().GetConfiguration().GetDrivers().GetVulkan().GetPhysicalDevices()) > 0 && *flags.EnableVulkanTracing {
 		apis = append(apis, tracer.VulkanTraceOptions())
+		// If ANGLE is enabled and available, need to also append ANGLE trace mode
```
lpy, line 286: Do we want to put ANGLE flag behind vulkan tracing flag?

core/os/android/adb/commands.go
```diff
@@ -373,6 +373,31 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
 	return result, nil
 }
 
+// Currently using Android Q/10, API 29 as ANGLE support cut-off
```
lpy, line 376: How much do we care about capturing on Android 10?

lpy:
Please also add `Bug: b/BUG_ID` in the description message.

core/os/android/adb/commands.go
```diff
@@ -373,6 +373,31 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
...
+	return os.GetAPIVersion() >= 29
+}
+
+func (b *binding) QueryAnglePackageName(ctx context.Context) (string, error) {
+	if !b.SupportsAngle(ctx) {
+		return "", fmt.Errorf("ANGLE not supported on this device")
+	}
+	// ANGLE supported, so check for installed ANGLE package
+	//  Favor custom installed package first, followed by default system package
+	custom := "com.chromium.angle"
```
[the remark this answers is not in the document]
pmuetschard: these are constants, so this is fine.

gapis/service/service.proto
```diff
@@ -804,6 +804,7 @@ message ErrUnsupportedVersion {
 enum TraceType {
   Graphics = 0;
   Perfetto = 1;
+  Angle = 2;
```
pmuetschard, line 807: Remove this. I know the Type/API combo is a bit confusing as it is right now. I think, to stick with the current non-obvious way, would be to use type Graphics and API "OpenGL on ANGLE", the way you are mostly doing already.

Another option would be to simply make ANGLE a checkbox for a Vulkan trace. What do you think? would that be too confusing to select Vulkan for an OpenGL on ANGLE app?

core/os/device/device.proto
```diff
@@ -173,6 +173,8 @@ message Configuration {
   Drivers drivers = 4;
   // Perfetto capability.
   PerfettoCapability perfetto_capability = 5;
+  // Angle package name
+  string angle_package = 6;
```
silence-do-good, line 177: Not sure if this is still relevant. If you want to restore the angle package after tracing, this can be a custom message type similar to configuration, where it has the list of all the packages and whichever was enabled before taking the trace and so on. This can later be used to set and unset the package to be used for tracing.

gapii/client/adb.go
```diff
@@ -89,6 +89,36 @@ func Start(ctx context.Context, p *android.InstalledPackage, a *android.Activity
...
+			// TODO: Am I cleaning up these settings correctly?
+			angle_driver_values, _ := d.SystemSetting(ctx, "global", "angle_gl_driver_selection_values")
+			angle_driver_pkgs, _ := d.SystemSetting(ctx, "global", "angle_gl_driver_selection_pkgs")
+			angle_package, _ := d.SystemSetting(ctx, "global", "angle_debug_package")
+			log.I(ctx, "Saving original ANGLE pkg %s for app %s to restore during cleanup.", angle_package, angle_driver_pkgs)
+			cleanup = cleanup.Then(func(ctx context.Context) {
+				d.SetSystemSetting(ctx, "global", "angle_gl_driver_selection_values", angle_driver_values)
+				d.SetSystemSetting(ctx, "global", "angle_gl_driver_selection_pkgs", angle_driver_pkgs)
+				d.SetSystemSetting(ctx, "global", "angle_debug_package", angle_package)
+			})
```
silence-do-good, line 109: I think this cleanup needs to be invoked for it to execute.

core/os/android/adb/commands.go
```diff
@@ -373,6 +373,31 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
...
+	os := b.Instance().GetConfiguration().GetOS()
+	return os.GetAPIVersion() >= 29
+}
+
+func (b *binding) QueryAnglePackageName(ctx context.Context) (string, error) {
+	if !b.SupportsAngle(ctx) {
+		return "", fmt.Errorf("ANGLE not supported on this device")
+	}
+	// ANGLE supported, so check for installed ANGLE package
+	//  Favor custom installed package first, followed by default system package
```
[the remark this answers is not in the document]
tobine: Fixed

gapii/client/adb.go
```diff
@@ -89,6 +89,36 @@ func Start(ctx context.Context, p *android.InstalledPackage, a *android.Activity
...
+	// TODO: Need to clean this up and get it working
+	// If Angle was selected, then save current angle settings to restore and then set Angle for selected app
+	log.I(ctx, "Checking if ANGLE requested")
+	if o.EnableAngle {
+		log.I(ctx, "ANGLE is enabled")
+		if anglePackage := p.Device.Instance().GetConfiguration().AnglePackage; anglePackage != "" {
+			log.I(ctx, "Found ANGLE package %s, enabling it for %s", anglePackage, p.Name)
+			// Restore ANGLE settings during app cleanup
+			// TODO: Am I cleaning up these settings correctly?
+			angle_driver_values, _ := d.SystemSetting(ctx, "global", "angle_gl_driver_selection_values")
```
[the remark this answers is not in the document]
tobine: Done

core/os/android/adb/commands.go
tobine: We still use a reasonable amount of 10 on the ANGLE team so we care for now, at least. Could potentially update this once 11 lands publicly.

core/os/device/device.proto
tobine: This is currently where I store the detected ANGLE package name in device.go. Then, if ANGLE tracing mode is selected, I verify that we stored the detected ANGLE package in adb.go before enabling ANGLE.
So I could kill this but then I'd have to find a different place to store ANGLE package name.

gapii/client/adb.go
tobine: Done

gapis/service/service.proto
tobine: Yep, killed this and couple related functions that aren't needed with how I'm currently detecting ANGLE.

I personally like having the separate drop-down for "OpenGL on ANGLE" vs. a potential checkbox. Helps to clearly distinguish between an OpenGL app and Vulkan app, and also avoids users trying to check the ANGLE box for a Vulkan app.

gapis/trace/android/trace.go
tobine: ANGLE can currently only be used for Graphics tracing mode so this makes sense to me. If there comes a point when we want to enable ANGLE or Vulkan tracing by default without the other, we would have to refactor this.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -1007,6 +1010,10 @@ private File getOutputFile() {
       private static boolean isPerfetto(TraceTypeCapabilities config) {
         return config != null && config.getType() == TraceType.Perfetto;
       }
+
+      private static boolean isAngle(TraceTypeCapabilities config) {
+        return config != null && config.getApi().equalsIgnoreCase("OpenGL on ANGLE");
```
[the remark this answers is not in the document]
tobine: See comment from @pmuetschard below. I've killed Angle as a trace type in latest code. An Angle trace is really a Graphics (Vulkan) trace w/ API type "OpenGL on Angle"

tobine:
Addressed all of the feedback. PTAL

PR #321, opened 2020-06-05
Add C/C++ headers from externals to gofuse

hevrard:
Also fix a comment.

Test: I now have all C headers resolved in VSCode.


hevrard:
@stellama0208 the particular file that was failing for me was the `<grpc++/grpc++.h>` include in e.g. `gapir/cc/server.cpp`.

More generally, I think it would fail for headers of our external dependencies.

PR #322, opened 2020-06-08
Fix JRE copy on MacOS and remove some files

pau-baiget:


pau-baiget:
Proper fix in https://github.com/google/agi/pull/324

PR #328, opened 2020-06-10
Implement the prototype for counter performance tab.

stellama0208:
- Move the performance tree to a separate wide tab.
 - Show counter performance with a time-factored counter value average.
 - Bug: b/158055737.

### DIFF
--- a/core/app/flags/experimental.go
+++ b/core/app/flags/experimental.go
@@ -26,4 +26,5 @@ var (
 	EnableFrameLifecycle = flag.Bool("experimental-enable-frame-lifecycle", false, "Enable the experimental feature Frame Lifecycle.")
 	EnableVulkanTracing  = flag.Bool("experimental-enable-vulkan-tracing", false, "Enable the experimental feature Vulkan tracing.")
 	EnableAngleTracing   = flag.Bool("experimental-enable-angle-tracing", false, "Enable the experimental feature ANGLE tracing.")
+	EnablePerfTab        = flag.Bool("experimental-enable-perf-tab", false, "Enable the experimental feature Counter performance tab.")
 )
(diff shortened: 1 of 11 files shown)


stellama0208:
The above force push was only to rebase PR onto AGI tot.

stellama0208:
In a f2f sync on perfTab with Pascal, some suggestions were given to alter this PR into a final mergeable state. Above three new commits took care of those advices, ready for a new look, thanks! :)

core/app/flags/experimental.go
```diff
@@ -26,4 +26,5 @@ var (
 	EnableFrameLifecycle = flag.Bool("experimental-enable-frame-lifecycle", false, "Enable the experimental feature Frame Lifecycle.")
 	EnableVulkanTracing  = flag.Bool("experimental-enable-vulkan-tracing", false, "Enable the experimental feature Vulkan tracing.")
 	EnableAngleTracing   = flag.Bool("experimental-enable-angle-tracing", false, "Enable the experimental feature ANGLE tracing.")
+	EnablePerfTab        = flag.Bool("experimental-enable-perf-tab", false, "Enable the experimental feature Counter performance tab.")
```
pmuetschard, line 29: Is this flag used anywhere in the go side? I don't think we need it on the server, "Tab" is very much UI specific :)

stellama0208: It's not yet used in the server side. But I think when I move the counter sample digestion logic to gapis, this flag might be put into use then. 
Yes, maybe this flag should be given a more neutral name, maybe "experimental-enable-counter-perf"? Or any good naming idea? The only worry I have is that I may introduce too many alias around this project. Will do the renaming if necessary in the following logic movement PR.

PR #331, opened 2020-06-11
Check NDK version in Bazel

hevrard:
This makes sure we refuse to build with an invalid NDK version.

Bug: b/158669059

### DIFF
--- a/gapidapk/android/apk/rules.bzl
+++ b/gapidapk/android/apk/rules.bzl
@@ -110,7 +110,7 @@ def gapid_apk(name = "", abi = "", pkg = "", libs = {}, bins = {}):
             "//conditions:default": [],
         }),
         deps = select({
-            "//tools/build:android-" + name: ["@ndk_vk_validation_layer//:" + abi],
+            "//tools/build:android-" + name: ["@ndk_version_check//:ndk_version_check", "@ndk_vk_validation_layer//:" + abi],
             "//conditions:default": [],
         }),
     )
--- a/gapidapk/android/apk/rules.bzl
+++ b/gapidapk/android/apk/rules.bzl
@@ -110,7 +110,7 @@ def gapid_apk(name = "", abi = "", pkg = "", libs = {}, bins = {}):
             "//conditions:default": [],
         }),
         deps = select({
-            "//tools/build:android-" + name: ["@ndk_version_check//:ndk_version_check", "@ndk_vk_validation_layer//:" + abi],
+            "//tools/build:android-" + name: ["@ndk_vk_validation_layer//:" + abi],
             "//conditions:default": [],
         }),
     )
@@ -124,6 +124,7 @@ def gapid_apk(name = "", abi = "", pkg = "", libs = {}, bins = {}):
         assets = assets,
         assets_dir = abi,
         deps = [
+            "@ndk_version_check//:version_check",
             "//gapidapk/android/app/src/main:gapid",
             ":" + name + "_native",
         ],
--- a/kokoro/presubmit/build.sh
+++ b/kokoro/presubmit/build.sh
@@ -38,12 +38,13 @@ curl -L -k -s https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
 sudo apt-get update
 sudo apt-get install -y clang-format-6.0
 
-# Mock the correct NDK version.
-# Our Bazel scripts check that the correct NDK version is installed: rather
-# than downloading the 1+GB NDK just to have a matching version to pass this
-# presubmit test, we "mock" the version by overwriting it manually.
-export ANDROID_NDK_HOME=/opt/android-ndk-r16b
-cat > $ANDROID_NDK_HOME/source.properties <<EOF
+# Fake NDK with the expected version.
+# Our Bazel scripts check that a specific NDK version is installed: rather
+# than downloading the 1GB+ NDK just to have a matching version, we manually
+# create a fake NDK with only a version file.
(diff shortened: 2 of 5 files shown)


tools/build/rules/android.bzl
```diff
@@ -135,3 +135,30 @@ ndk_vk_validation_layer = repository_rule(
         "ANDROID_NDK_HOME",
     ],
 )
+
+# Enforce NDK version by checking $ANDROID_NDK_HOME/source.properties
+def _ndk_version_check(repository_ctx):
+    # This should be updated in sync with BUILDING.md documentation
+    version = "21.0.6113669"
```
yalcinmelihyasin, line 142: Should we have a specific version or should we accept anything starts with 26? Do we want to reject if there is a minor version difference?

hevrard: I guess you meant "that starts with 21" ?

I think we should impose specific versions. There are very few NDK versions made public, for instance in the NDK r21 series, there are only 2 public releases so far:
- `r21` with version `21.0.6113669` (used here)
- `r21d` with version `21.3.6528147`, freshly released a few days ago (updating AGI to use this one is tracked by b/158730767)

And I suspect the next NDK release will be `r22`, around October.

Because our users are unlikely to have many different minor versions of the NDK lying around, I think we should match a precise version.

yalcinmelihyasin: Seems fine for me.

kokoro/presubmit/build.sh
```diff
@@ -38,8 +38,17 @@ curl -L -k -s https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
...
+# Our Bazel scripts check that the correct NDK version is installed: rather
+# than downloading the 1+GB NDK just to have a matching version to pass this
+# presubmit test, we "mock" the version by overwriting it manually.
 export ANDROID_NDK_HOME=/opt/android-ndk-r16b
+cat > $ANDROID_NDK_HOME/source.properties <<EOF
+Pkg.Desc = Android NDK
+Pkg.Revision = 21.0.6113669
+EOF
+
+# Setup environment.
```
pmuetschard, line 51: What's going on here? if we are ok with an older ndk, why are we adding a check, and then by-passing it? FYI, we're already downloading the NDK on the build bots for the other OSes.

hevrard: This is only for `kokoro/presubmit`, where we don't really "build" anything. This avoids a 1GB download for the NDK on each presubmit, only to check the NDK version.

I agree this is hacky. Note that we did not download an up-to-date NDK on presubmit before this PR.

@pmuetschard do you think it is better to download 1GB on every presubmit for the sake of passing the version check?

pau-baiget: ping @pmuetschard

hevrard: Pascal and I chatted offline: Pascal is keen on having this rule being called only when we need to compile anything using the NDK.

@pmuetschard , this NDK check needs Bazel's [repository_ctx.read()](https://docs.bazel.build/versions/2.0.0/skylark/lib/repository_ctx.html#read) to check the content of the `source.properties` file, and I guess this is only accessible in a repository rule?

If that is so, I think you told me it's possible to pass a flag to Bazel to set the `android` argument of gapid_dependencies() to false?

gapidapk/android/apk/rules.bzl
```diff
@@ -110,7 +110,7 @@ def gapid_apk(name = "", abi = "", pkg = "", libs = {}, bins = {}):
             "//conditions:default": [],
         }),
         deps = select({
-            "//tools/build:android-" + name: ["@ndk_vk_validation_layer//:" + abi],
+            "//tools/build:android-" + name: ["@ndk_version_check//:ndk_version_check", "@ndk_vk_validation_layer//:" + abi],
```
pmuetschard, line 113: Fix the formatting here. Don't have multiple list items on a signle line.

kokoro/presubmit/build.sh
pmuetschard: Bazel has multiple phases of execution. I'm asking for the error failure to happen in a *later* phase. Alternatively, make it a warning rather than a failure.

You are already generating a dummy rule that we'll depend on. So, change the repository init code to create different dummies depending on whether it should fail or not. If it should fail, the dummy will then fail. That way it will only fail when the dummy is actually invoked.

hevrard:
@pmuetschard thanks for the Bazel guidance, I think I reached the setup you described.

hevrard:
@pmuetschard a friendly ping on this one

pmuetschard:
Make sure to squash.

pmuetschard:
FYI, our current version of bazel doesn't support 21 yet. I'm building with NDK 20 just fine, so I'm not sure why we think we *need* 21?

hevrard:
Thanks, we need NDK r21 in order to have access to the unified validation layer, it's documented in https://github.com/google/agi/blob/master/BUILDING.md. Bazel does not officially support NDK > r20, we asked for this support with https://github.com/bazelbuild/bazel/issues/10811.

pmuetschard:
> we need NDK r21

As I pointed out, it builds fine for me on 20, so I'm not sure why you say we *need* it.

hevrard:
It wasn't clear from my message: the unified Khronos verification layer ships only from NDK r21, and if you build with NDK < r21, you don't get it, and things go wrong when you run a report replay -- this is the scenario that prompter to file b/158669059, and hence this fix.

PR #335, opened 2020-06-12
Clarify error message upon replayer connection loss

hevrard:


gapir/client/connection.go
```diff
@@ -224,11 +224,11 @@ func (c *connection) HandleReplayCommunication(
 	}()
 	for {
 		if c.stream == nil {
-			return log.Errf(ctx, nil, "Replay stream connection lost")
+			return log.Errf(ctx, nil, "Replayer connection lost")
 		}
 		r, err := c.stream.Recv()
 		if err != nil {
-			return log.Errf(ctx, err, "Recv")
+			return log.Errf(ctx, err, "Replayer connection lost")
```
yalcinmelihyasin, line 231: I would *definitely* use different messages here. The two errors rise up from different reasons and it would really help to use different messages here.

hevrard: Good point, I've updated to disambiguate between the two errors.

PR #337, opened 2020-06-15
Simplify gapir socket naming

hevrard:
The socket used for GAPIS-GAPIR connection has an ABI-specific name:
this is not needed, because it is backed by a filesystem socket that
is created in the APK directory on device, and there is a distinct APK
for each ABI, so the name of the directory already makes it unique.

This change reduces the number of const strings that have to be kept
in sync between Go and C++ (here socket names), and properly documents
that these strings should be kept in sync.

### DIFF
--- a/gapir/client/device_connection.go
+++ b/gapir/client/device_connection.go
@@ -44,6 +44,8 @@ const (
 	deviceConnectionTimeout    = time.Second * 30
 	maxCheckSocketFileAttempts = 10
 	checkSocketFileRetryDelay  = time.Second
+	// socketName must match kSocketName in gapir/cc/main.cpp
+	socketName = "gapir-socket"
 )
 
 type deviceConnectionInfo struct {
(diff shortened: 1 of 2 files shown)


gapir/client/device_connection.go
```diff
@@ -233,13 +235,6 @@ func newHost(ctx context.Context, d bind.Device, abi *device.ABI, launchArgs []s
 	return &deviceConnectionInfo{port: port, authToken: authToken, cleanupFunc: cleanupFunc}, nil
 }
 
-var socketNames = map[device.Architecture]string{
```
purvisa-at-google-com, line 236: Is it safe to have one name for all sockets regardless of ISA? What do we lose?

hevrard: Good point, I try to explain that in the commit message: this socket name is used to decide on the file name for the unix socket, and this file is created on-device in a directory of the gapid APK, and we have a distinct APK per ABI. So the replayer for e.g. arm64 will listen on a unix socket backed by a file created under `/data/user/0/com.google.android.gapid.arm64v8a/`, and it will be a different directory for other ABIs (because they are different APKs).

The previous code was reflecting the ABI in the socket name, but that is redundant.

purvisa-at-google-com: Haha, sorry. I should have read that! LGTM

PR #339, opened 2020-06-15
WIP:Add testing for ANGLE tracing

tobine:
Some code and comments related to updating test framework
to handle ANGLE testing.
This should just be a frame capture test where API is set
to "OpenGL on ANGLE." Flagged some key code areas and I'm
hopeful that there's a simple way to just set verb.API to
"angle" for an ANGLE test and that will handle ANGLE test
as a custom frame capture variation.

### DIFF
--- a/cmd/gapit/flags.go
+++ b/cmd/gapit/flags.go
@@ -318,7 +318,7 @@ type (
 		No struct {
 			Buffer bool `help:"Do not buffer the output, this helps if the application crashes"`
 		}
-		API   string `help:"only capture the given API valid options are vulkan and perfetto"`
+		API   string `help:"only capture the given API valid options are vulkan, angle and perfetto"`
 		Local struct {
 			Port int `help:"connect to an application already running on the server using this port"`
 		}
--- a/cmd/gapit/benchmark.go
+++ b/cmd/gapit/benchmark.go
@@ -792,7 +792,7 @@ func (verb *benchmarkVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 }
 
 // This intentionally duplicates a lot of the gapit trace logic
-// so that we can independently chnage how what we do to benchmark
+// so that we can independently change what we do to benchmark
 // everything.
 func (verb *benchmarkVerb) doTrace(ctx context.Context, client client.Client, traceURI string) error {
 	ctx = status.Start(ctx, "Record Trace for %+v frames", verb.NumFrames)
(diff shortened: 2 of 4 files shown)


cmd/gapit/flags.go
```diff
@@ -318,6 +318,7 @@ type (
 		No struct {
 			Buffer bool `help:"Do not buffer the output, this helps if the application crashes"`
 		}
+		// TODO: I think we could add "angle" as valid option here for "OpenGL on ANGLE" mode
 		API   string `help:"only capture the given API valid options are vulkan and perfetto"`
```
hevrard, line 322: Yes, adding "angle" as an API is the right approach here.

test/swarming/bot-scripts/angle_profile.py
```diff
@@ -0,0 +1,64 @@
...
+    botutil.load_params(test_params, required_keys=required_keys)
+    assert os.path.isfile(test_params['gfxtrace'])
+
+    #### Profile
+    gapit = os.path.join(agi_dir, 'gapit')
+    # TODO: Need to add something here to enable ANGLE
+    #  Do I need an "--angle" option that will set ANGLE in the code...
+    #  Or is there a way to just set "verb.API" value as "angle" in test
+    #  json in which case we can just use existing frame_profile.py
+    #  script and customize json input(s) for ANGLE testing?
```
hevrard, line 50: Good question. I think we can add an `api` entry in the test params JSON.

Note that we probably want to focus on the `video.py` first, it is the one doing capture and replay. So in `video.py`, we can default to vulkan api in `gapit trace ...`, but override by the `api` entry in params JSON when present. See:
https://github.com/google/agi/blob/ca38aa96ef04e6ad432e3852212ad7afd1b00c88/test/swarming/bot-scripts/video.py#L57

`frame_profile.py` does not do a capture, it just takes a pre-captured `gfxtrace` as input and runs a profile on it. We should be able to reuse it without modification by just providing a gfxtrace file obtained with an Angle capture.

tobine:
Alright, I was able to verify that with my change to gapit/trace.go I can just add "api" : "angle" to the params.json file and use the existing video.py script to test ANGLE tracing.
That makes this change very straightforward.
Once we land this I can add cubey test to get started, which I've verified locally.

hevrard:
Sounds good. Currently I can only see changes in gapit, I'll wait for your changes in video.py to do a proper review.

tobine:
No video.py changes are required. I can use video.py script as-is and add "api": "angle" to params.json to turn on ANGLE tracing with just the gapit changes.

tobine:
Pushed update to make video.py script take "-api" as input param instead of being hard-coded. Thanks @hevrard !

Issue #340, opened 2020-06-16
cannot select device


uO-Ou:
hi, i cannot select device when "Capture a new trace".
![image](https://user-images.githubusercontent.com/16663574/84726281-19854480-afbf-11ea-89b4-2dc70015779c.png)

i'm sure my android device is connected.

this is my device-info:

`"configuration": {
  "OS": {
   "kind": "OSX",
   "name": "OSX",
   "majorVersion": 10,
   "minorVersion": 15,
   "pointVersion": 5
  },
  "hardware": {
   "name": "MacBookPro15,4",
   "CPU": {
    "name": "Intel(R) Core(TM) i7-8557U CPU @ 1.70GHz",
    "vendor": "GenuineIntel",
    "architecture": "X86_64",
    "cores": 8
   },
   "GPU": {}
  },
  "ABIs": [
   {
    "name": "x86_64",
    "OS": "OSX",
    "architecture": "X86_64",
    "memoryLayout": {
     "endian": "LittleEndian",
     "pointer": {
      "size": 8,
      "alignment": 8
     },
     "integer": {
      "size": 4,
      "alignment": 4
     },
     "size": {
      "size": 8,
      "alignment": 8
     },
     "char": {
      "size": 1,
      "alignment": 1
     },
     "i64": {
      "size": 8,
      "alignment": 8
     },
     "i32": {
      "size": 4,
      "alignment": 4
     },
     "i16": {
      "size": 2,
      "alignment": 2
     },
     "i8": {
      "size": 1,
      "alignment": 1
     },
     "f64": {
      "size": 8,
      "alignment": 8
     },
     "f32": {
      "size": 4,
      "alignment": 4
     },
     "f16": {
      "size": 2,
      "alignment": 2
     }
    }
   }
  ],
  "drivers": {},
  "perfettoCapability": {}`

hevrard:
Profiling with AGI dev-releases is currently restricted to a set of Android devices, as documented here: https://gpuinspector.dev/docs/requirements#supported-android-devices

It also requires beta GPU drivers, which are only accessible via the developer preview for which you can sign-up here: https://gpuinspector.dev/

Profiling on macOS is not supported, hence the empty "Device" menu in your case.

PR #341, opened 2020-06-16
Default trace dialog title to system profile

hevrard:
As graphics capture are currently experimental, default to "System
Profile" in the trace dialog window title.

Related: see screenshot of bug in #340

Bug: b/159096531

### DIFF
--- a/gapic/src/main/com/google/gapid/util/Messages.java
+++ b/gapic/src/main/com/google/gapid/util/Messages.java
@@ -48,6 +48,7 @@ public interface Messages {
   public static final String MEMORY_STRUCT_TAB_TEXT = "Struct";
   public static final String CAPTURE_TRACE_GRAPHICS = "Capture Graphics Trace";
   public static final String CAPTURE_TRACE_PERFETTO = "Capture System Profile";
+  public static final String CAPTURE_TRACE_SELECT = "Capture: Select Trace Type";
   public static final String CAPTURING_TRACE = "Capturing...";
   public static final String CAPTURE_DIRECTORY = "Capture output directory...";
   public static final String CAPTURE_EXECUTABLE = "Executable to trace...";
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -228,7 +228,7 @@ public void setDevices(List<DeviceCaptureInfo> devices) {
 
     @Override
     public String getTitle() {
-      return Messages.CAPTURE_TRACE_GRAPHICS;
+      return Messages.CAPTURE_TRACE_PERFETTO;
```
pmuetschard, line 231: This is the wrong "fix". When the "API" dropdown changes, we update the title:
```
getShell().setText(
            isPerfetto ? Messages.CAPTURE_TRACE_PERFETTO : Messages.CAPTURE_TRACE_GRAPHICS);
```
We need to make sure that this is also done when first initializing it. Otherwise we will show the wrong title again if for a user graphics trace is selected at init time. Maybe here we should give a default title of "Capture Trace" and then change it on dropdown init and selection change.

hevrard: I chose to display a default "select trace type" message on dialog creation, then the code you pointed out will update the dialog title when a given trace type is selected.

gapic/src/main/com/google/gapid/util/Messages.java
```diff
@@ -48,6 +48,7 @@
   public static final String MEMORY_STRUCT_TAB_TEXT = "Struct";
   public static final String CAPTURE_TRACE_GRAPHICS = "Capture Graphics Trace";
   public static final String CAPTURE_TRACE_PERFETTO = "Capture System Profile";
+  public static final String CAPTURE_TRACE_SELECT = "Capture: Select Trace Type";
```
pmuetschard, line 51: I don't think this really makes sense as a title. Maybe use `Capture Trace` instead? I don't want to bikeshed this, so...lgtm.

hevrard: Updated!

PR #344, opened 2020-06-18
Cleanup gapir/client package

hevrard:
This is a significant cleanup of the gapir/client package that should
not change the package behavior. The code essentially behaves in the
same way, it just has less indirections, less functions, less types,
and less interfaces. Function, type and package documentation has been
corrected and updated, some unused pieces of code have been removed,
and some functions, types and variables have been renamed to better
reflect what they represent.

This change is better reviewed as a clean reading of the client.go
file, which now contains the logic that was distributed across
client.go, background_connection.co and connection.go. Again, this
change does not change the code logic: there are not-so-good
interactions left over (like the fact that the Replay RPC stream is
created inside handleReplayCommunication() and creation status is
reported back via a channel, when it could be simply created in
Client.Connect() as part of a replayer initialization), for the sake
of preserving the behavior. This is a cleanup to have a cleaner state
from which we can achieve actual behavior refactoring.

Refactoring notes:

- the information related to a single GAPIR instance was stored across
the clientInfo, connection and backgroundConnection types, with some
duplicated fields. These three types are merged into the single
replayerInfo type, which can be seen as the spiritual descendant of
clientInfo that has absorbed connection and backgroundConnection.

- clientInfo (now replayerInfo) stored both a *device.ABI and
device.Architecture field, the architecture field has been removed
because it is trivially retrievable from the ABI field.

- an auth.Token was stored in deviceConnectionInfo and replicated in
connection, it is now stored only once in deviceConnectionInfo.

- the OS field of backgroundConnection is trivially retrievable from
the bind.Device field, hence the OS is not stored as a separate field.

- type deviceArch was declared only to be aliased to the exported
ConnectionKey type, we now just declare a ReplayerKey type that
effectively replaces ConnectionKey.

- the 'conn' variable name was abused to refer to either a connection
or a ConnectionKey, now all ReplayerKey variables are properly named
'key'.

- the Connection and ReplayResponseHandler interfaces defined in
api.go were not used outside of the gapir/client package. The actual
gapir-related functions used outside of gapir/client are the ones
exported by the Client type (namely New, BeginReplay, Connect,
PrewarmReplay, and SetReplayExecutor), and the ReplayExecutor
interface. Having the Connection and ReplayResponseHandler interface
being declared in gapir/api.go was somehow misleading in indicating
that these interfaces were relevant outside of the gapir/client
package. These two interfaces became obsolete and are now removed.

- most exported functions of Client receive a ReplayerKey: the lookup
of this key, and the subsequent check of the replayer found, was
duplicated in various places, it is now factored in
Client.getActiveReplayer().

- the former client.ping() function returned a time.Duration which was
never used, this time.Duration return value is removed.

- some small functions used only once have been inlined.

- some functions (e.g. BeginReplay and PrewarmReplay) were defined
three times for types Client, connection and backgroundConnection,
with the backgroundConnection being a wrapper around the connection
one, and Client being a wrapper around the backgroundConnection
one. This lead to intricate call hierarchies and redirections that are
hard to reason about. With the unification of types under
replayerInfo, most of these duplicates are gone.


gapir/client/doc.go
```diff
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package client provides helper methods and types for communicating with the
-// GAPIR service.
+// Package client provides helper methods and types for starting GAPIR
+// instances and communicating with them. The actual interaction is
+// achieved via the gRPC service defined in:
+// gapir/replay_service/service.proto
+// The client package abstracts the RPC handling details an offers a
```
pau-baiget, line 19: an -> and

gapir/client/client.go
```diff
@@ -17,80 +17,118 @@ package client
...
-	bgConnection         *backgroundConnection
+// ReplayExecutor must be implemented by replay executors to handle some live
+// interactions with a running replay.
+type ReplayExecutor interface {
+	// HandlePostData handles the given post data message.
+	HandlePostData(context.Context, *gapir.PostData) error
+	// HandleNotification handles the given notification message.
+	HandleNotification(context.Context, *gapir.Notification) error
+	// HandleFinished is notified when the given replay is finished.
+	HandleFinished(context.Context, error) error
```
yalcinmelihyasin, line 66: We may want to add another function here for any case when replay cannot be finished due to an error e.g. Connection, crash etc.

gapir/client/client.go
```diff
@@ -17,80 +17,118 @@ package client
...
+	mutex     sync.Mutex
+	replayers map[ReplayerKey]*replayerInfo
 }
 
-// New returns a newly construct Client.
+// New returns a new Client with no replayers.
 func New(ctx context.Context) *Client {
-	client := &Client{clientInfos: map[ConnectionKey]clientInfo{}}
+	client := &Client{replayers: map[ReplayerKey]*replayerInfo{}}
 	app.AddCleanup(ctx, func() {
```
yalcinmelihyasin, line 100: I don't know if this should be continue to reside here. We may want to consider the app calling shutdown directly, instead of adding this to the app's cleanup list. 

If app is creating a client then it owns the client and should take care of it.

gapir/client/client.go
```diff
@@ -102,33 +140,152 @@ func (client *Client) Connect(ctx context.Context, device bind.Device, abi *devi
 
 	log.I(ctx, "Waiting for connection to GAPIR...")
 
-	connection, err := newConnection(fmt.Sprintf("localhost:%d", newDeviceConnectionInfo.port), newDeviceConnectionInfo.authToken, connectTimeout)
+	// Create gRPC connection
+	conn, err := grpc.Dial(fmt.Sprintf("localhost:%d", newDeviceConnectionInfo.port),
```
yalcinmelihyasin, line 144: I don't know if it is a good idea to call grpc directly here. It seems to be violation of separation of concerns. For example, if we would want to test this function and want to use a mock instead of grpc, we cannot do it.

gapir/client/client.go
```diff
@@ -102,33 +140,152 @@ func (client *Client) Connect(ctx context.Context, device bind.Device, abi *devi
...
+				return
+			}
+		}
+	}
+}
+
+// getActiveReplayer returns the replayer identified by key only if this
+// replayer has an active connection to a GAPIR instance.
+func (client *Client) getActiveReplayer(ctx context.Context, key *ReplayerKey) (*replayerInfo, error) {
+	replayer, found := client.replayers[*key]
```
yalcinmelihyasin, line 212: This function should be protected by mutex.

If the heartbeat calls reconnect from a channel(I don't know the Go's channel model well but I suspect that it can be from another thread), when this method is in process, we may crash in for example in line 217 due to replayer being deleted meanwhile.

gapir/client/client.go
```diff
@@ -102,33 +140,152 @@ func (client *Client) Connect(ctx context.Context, device bind.Device, abi *devi
...
+		return nil, log.Err(ctx, nil, "Replayer has no active connection")
+	}
+
+	return replayer, nil
+}
+
+// BeginReplay sends a replay request to the replayer identified by key.
+func (client *Client) BeginReplay(ctx context.Context, key *ReplayerKey, payload string, dependent string) error {
+	ctx = log.Enter(ctx, "Starting replay on gapir device")
+	replayerInfo, err := client.getActiveReplayer(ctx, key)
```
yalcinmelihyasin, line 227: Consider multi-threading. 

We should check if hearbeat calls reconnect from another thread. If that is the case then every function that reads/writes replayer map should be mutex protected.

gapir/client/client.go
```diff
@@ -102,33 +140,152 @@ func (client *Client) Connect(ctx context.Context, device bind.Device, abi *devi
...
+
+	idReq := replaysrv.ReplayRequest{
+		Req: &replaysrv.ReplayRequest_Replay{
+			Replay: &replaysrv.Replay{
+				ReplayId:    payload,
+				DependentId: dependent,
+			},
+		},
+	}
+	err = replayerInfo.rpcStream.Send(&idReq)
```
yalcinmelihyasin, line 240: Even though I like removing files and reducing complexity, we may still want to separate rpc layer from this for both separation of concerns and testing.

gapir/client/client.go
```diff
@@ -102,33 +140,152 @@ func (client *Client) Connect(ctx context.Context, device bind.Device, abi *devi
...
+	err = replayerInfo.rpcStream.Send(&PrerunReq)
+	if err != nil {
+		return log.Err(ctx, err, "Sending replay payload")
+	}
+	return nil
+}
+
+// startReplayCommunicationHandler launches a background task which creates
+// the Replay RPC stream and starts to listen to it.
+func (replayer *replayerInfo) startReplayCommunicationHandler(ctx context.Context) error {
```
yalcinmelihyasin, line 288: I would prefer the definition of replayerInfo and the functions related with it in a different file then here both for readability and to reflect the actual structure.

gapir/client/client.go
hevrard: That sounds good, but it is a behavioral change, so I don't think it fits in this cleanup PR that tries hard to NOT change the package behavior/interface (beyond renaming).

gapir/client/client.go
hevrard: I think `core/app` is a framework to write apps (used also by `gapit`, `apic`, etc...), and it has no idea about the app internal details. As such, I don't think we should special-case the gapir client in this framework.

gapir/client/client.go
hevrard: In the case of gapir, we can create a mock GAPIR gRPC server, and pass its listening `port` to this piece of code. The mock creation would actually happen in `device_connection.go`.

pau-baiget:
```
there are not-so-good
interactions left over (like the fact that the Replay RPC stream is
created inside handleReplayCommunication() and creation status is
reported back via a channel, 
```

Maybe we should capture this nice-to-have remaining work in a separate ticket?

gapir/client/client.go
pau-baiget: I guess the purpose of this PR is not to fix existing issues, just to move code around and remove complexity. I see however a benefit of following up quickly on this feedback to actually fix the issues we're finding here.

gapir/client/client.go
hevrard: Agreed, moved to `replayer.go`

gapir/client/doc.go
hevrard: Good catch! fixed.

gapir/client/client.go
hevrard: I've add a mutex lock/unlock at the start of all exported functions.

hevrard:
@pau-baiget agreed, I have further plans to cleanup this code with behavioral changes once this PR is in, I maintain notes at go/agi-replay-error

PR #348, opened 2020-06-22
Clean up legacy names.

lpy:
Rename the tag name to AGI in the log.

Bug: N/A
Test: adb logcat


lpy:
This part was lifted out from another patch, ptal

PR #349, opened 2020-06-24
Do not fetch FB thumbnails when selected command has no attachments available

hliatis:
b/158822064

### DIFF
--- a/gapic/src/main/com/google/gapid/views/FramebufferView.java
+++ b/gapic/src/main/com/google/gapid/views/FramebufferView.java
@@ -133,6 +133,10 @@ public class FramebufferView extends Composite
   private final AttachmentPicker picker;
   private final Label pickerLabel;
 
+  private final GridData pickerGridData;
+  private final Button pickerToggle;
+  private boolean showAttachments;
+
   public FramebufferView(Composite parent, Models models, Widgets widgets) {
     super(parent, SWT.NONE);
     this.models = models;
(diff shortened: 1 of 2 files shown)


gapis/resolve/framebuffer_attachment.go
```diff
@@ -31,6 +31,15 @@ import (
...
 func FramebufferAttachments(ctx context.Context, p *path.FramebufferAttachments, r *path.ResolveConfig) (interface{}, error) {
+	cmd, err := Cmd(ctx, p.After, r)
+	if err != nil {
+		return nil, err
+	}
+
+	if !cmd.CmdFlags().IsExecutedDraw() && !cmd.CmdFlags().IsEndOfFrame() {
+		return nil, &service.ErrDataUnavailable{Reason: messages.ErrNotADrawCall()}
+	}
+
```
pmuetschard, line 42: This should not be here. The resolvable should return this error instead. I know we used to have this functionality, at least in GLES, where it would give a good error message (such as "there's no framebufer, something, something").

It should be the resolvable that determines when an FB is available.

hliatis: Non-draw/non-present calls technically do have resolvable framebuffer attachments. Unless you mean I physically copy-and-paste those line to the resolve functon.

pmuetschard: No, my point is exactly that we should still show them in those cases. I am pointing out that it should be the act of resolving that decides whether a FB is available and applicable at a specific command. It should not be the API-agnostic part of the code that decides.

The issue is when you select a command where there isn't an FB, the error is not nice and we should hide the thumbnails.

hliatis: I see. Well, that's easy to fix: I just get rid of these lines. The UI code responds the same way regardless of error.

gapic/src/main/com/google/gapid/views/FramebufferView.java
```diff
@@ -133,6 +133,11 @@ private RenderSetting(int maxWidth, int maxHeight, Path.DrawMode drawMode) {
   private final AttachmentPicker picker;
   private final Label pickerLabel;
 
+  private final GridData pickerGridData;
+  private final SashForm splitter;
```
stellama0208, line 137: It looks like the splitter is only accessed locally at initialization, and doesn't need to be member variable?

gapis/resolve/framebuffer_attachment.go
```diff
@@ -39,6 +39,15 @@ func FramebufferAttachments(ctx context.Context, p *path.FramebufferAttachments,
 }
 
 func (r *FramebufferAttachmentsResolvable) Resolve(ctx context.Context) (interface{}, error) {
+	cmd, err := Cmd(ctx, r.Path.After, r.Config)
+	if err != nil {
+		return []*service.FramebufferAttachment{}, err
+	}
+
+	if cmd.CmdFlags().IsSubmission() {
+		return []*service.FramebufferAttachment{}, &service.ErrDataUnavailable{Reason: messages.ErrMessage("Please select a draw-call or VkQueuePresent")}
```
pmuetschard, line 48: This is in API-agnostic code. So I would rename it to "a draw-call or present".

PR #350, opened 2020-06-24
Fix depth attachment error when requesting a thumbnail smaller than the actual attachment

hliatis:
When requesting a thumbnail for depth attachments, the image will return as an error due to size disparities. This should fix that.

Bug: b/155322016

### DIFF
--- a/gapis/api/vulkan/read_framebuffer.go
+++ b/gapis/api/vulkan/read_framebuffer.go
@@ -107,7 +107,7 @@ func keyFromIndex(idx api.SubCmdIdx) string {
 	return fmt.Sprintf("%v", idx)
 }
 
-func (t *readFramebuffer) Depth(ctx context.Context, id api.SubCmdIdx, idx uint32, res replay.Result) {
+func (t *readFramebuffer) Depth(ctx context.Context, id api.SubCmdIdx, requestWidth, requestHeight, idx uint32, res replay.Result) {
 	t.injections[keyFromIndex(id)] = append(t.injections[keyFromIndex(id)], injection{res,
 		func(ctx context.Context, cmd *InsertionCommand, res replay.Result, out transform.Writer) error {
 			s := out.State()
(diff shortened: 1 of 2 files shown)


gapis/api/vulkan/read_framebuffer.go
```diff
@@ -152,7 +152,7 @@ func (t *readFramebuffer) Depth(ctx context.Context, id api.SubCmdIdx, idx uint3
 			// first one.
 			// TODO: support multi-layer rendering.
 			layer := imageViewDepth.SubresourceRange().BaseArrayLayer()
-			return t.postImageData(ctx, cb, id, s, cmd.cmdBuffer, cmd.pendingCommandBuffers, depthImageObject, imageViewDepth.Fmt(), VkImageAspectFlagBits_VK_IMAGE_ASPECT_DEPTH_BIT, layer, level, w, h, w, h, out, res)
+			return t.postImageData(ctx, cb, id, s, cmd.cmdBuffer, cmd.pendingCommandBuffers, depthImageObject, imageViewDepth.Fmt(), VkImageAspectFlagBits_VK_IMAGE_ASPECT_DEPTH_BIT, layer, level, w, h, width, height, out, res)
```
pau-baiget, line 155: Since the two pairs of size params of this function are

```
func (t *readFramebuffer) postImageData(ctx context.Context,
...
	imgWidth,
	imgHeight,
	requestWidth,
	requestHeight uint32,
...
```
could you please rename the ```w,h``` and ```width, height``` accordingly, for clarify purposes?

pmuetschard: I'm ok if you want to leave `w` and `h` as is, to not pollute the change. But you could use `req[uest]Width` and `req[uest]Height` for the new parameters.

pmuetschard:
Please do give the PR a better description and title.

PR #352, opened 2020-06-24
Fix copying of packed depth/stencil images.

pmuetschard:
When computing the offset in the image buffer, the (larger) element size in the buffer was used, which caused a read out of bounds.

Bug: http://b/158838840


gapis/api/vulkan/api/copy_clear_commands.api
```diff
@@ -627,7 +627,7 @@ sub void copyImageBuffer(VkBuffer buffer, VkImage image, VkImageLayout layout, m
     aspectBit := aspectBits[0]
```
hevrard, line 627: Nit: there should be one more level of indentation in the block starting here, as we're in the `else` part of the `if` in the line above. See the `}}` near the end of this `sub`.

PR #354, opened 2020-06-26
Improve image <-> buffer copy mutate performance.

pmuetschard:
This improves the copy from/to image to/from buffer for packed depth formats by lifting out the expensive `readMemoryInBuffer` and `getBufferBoundMemoryPiecesInRange` calls out of the per-pixel loop and only calling them per row. Improvement on a representative trace on my machine is ~4x.

### DIFF
--- a/gapis/api/vulkan/api/copy_clear_commands.api
+++ b/gapis/api/vulkan/api/copy_clear_commands.api
@@ -676,19 +676,23 @@ sub void copyImageBuffer(VkBuffer buffer, VkImage image, VkImageLayout layout, m
           yInExtent := y - yStart
           rowStartByte := depthStartByte + y * imageLayout.rowPitch
           rowStartBlockInExtent := ((zInExtent * imageHeight) + yInExtent) * rowLength
+          bufStart := as!VkDeviceSize(bufferLayerOffset + rowStartBlockInExtent * elementSizeInBuffer)
+          bufSize := as!VkDeviceSize((xEnd - xStart) * elementSizeInBuffer)
+          readMemoryInBuffer(bufferObject, bufStart, bufSize)
+          bufMemPieces := getBufferBoundMemoryPiecesInRange(bufferObject, bufStart, bufSize)
+          imgStart := rowStartByte + xStart * elementSizeInImage
           if isPackedDepth {
-            for x in (xStart .. xEnd) {
-              imgStart := as!VkDeviceSize(rowStartByte + x * elementSizeInImage)
-              rowStartInExtent := (rowStartBlockInExtent + x) * elementSizeInBuffer
-              bufStart := as!VkDeviceSize(bufferLayerOffset + rowStartInExtent)
-              readMemoryInBuffer(bufferObject, bufStart, as!VkDeviceSize(elementSizeInBuffer))
-              // Discard the 4th byte in the buffer for current pixel.
-              bufMemPieces := getBufferBoundMemoryPiecesInRange(bufferObject, bufStart, as!VkDeviceSize(elementSizeInImage))
-              for _ , _ , piece in bufMemPieces {
-                bufMemStart := piece.MemoryOffset
-                bufMemEnd := bufMemStart + piece.Size
-                imgPieceStart := (piece.ResourceOffset - bufStart) + imgStart
-                imgPieceEnd := imgPieceStart + piece.Size
+            for _ , _ , piece in bufMemPieces {
+              // The below assumes that each memory piece is aligned to a multiple of
+              // elementSizeInBuffer. This has to be true, since the smallest alignment of pages is
+              // 64 bytes, and elementSizeInBuffer is 4.
+              pieceXStart := as!u64(piece.ResourceOffset - bufStart) / elementSizeInBuffer
+              pieceXEnd := as!u64(piece.Size) / elementSizeInBuffer
+              for x in (0 .. (pieceXEnd - pieceXStart)) {
+                bufMemStart := piece.MemoryOffset + as!VkDeviceSize(x * elementSizeInBuffer)
+                bufMemEnd := bufMemStart + as!VkDeviceSize(elementSizeInImage)
+                imgPieceStart := imgStart + x * elementSizeInImage
+                imgPieceEnd := imgPieceStart + elementSizeInImage
                 if isSrcBuffer {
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/api/copy_clear_commands.api
```diff
@@ -676,19 +676,23 @@ sub void copyImageBuffer(VkBuffer buffer, VkImage image, VkImageLayout layout, m
...
-              bufMemPieces := getBufferBoundMemoryPiecesInRange(bufferObject, bufStart, as!VkDeviceSize(elementSizeInImage))
-              for _ , _ , piece in bufMemPieces {
-                bufMemStart := piece.MemoryOffset
-                bufMemEnd := bufMemStart + piece.Size
-                imgPieceStart := (piece.ResourceOffset - bufStart) + imgStart
-                imgPieceEnd := imgPieceStart + piece.Size
+            for _ , _ , piece in bufMemPieces {
+              // The below assumes that each memory piece is alligned to a multiple of
+              // elementSizeInBuffer. This has to be true, since the smallest alignment of pages is
+              // 64 bytes, and elementSizeInBuffer is 4.
```
pau-baiget, line 688: if ```elementSizeInBuffer is 4``` why do we have this stored in a variable?

gapis/api/vulkan/api/copy_clear_commands.api
```diff
@@ -676,19 +676,23 @@ sub void copyImageBuffer(VkBuffer buffer, VkImage image, VkImageLayout layout, m
...
+              pieceXStart := as!u64(piece.ResourceOffset - bufStart) / elementSizeInBuffer
+              pieceXEnd := as!u64(piece.Size) / elementSizeInBuffer
+              for x in (0 .. (pieceXEnd - pieceXStart)) {
+                bufMemStart := piece.MemoryOffset + as!VkDeviceSize(x * elementSizeInBuffer)
+                bufMemEnd := bufMemStart + as!VkDeviceSize(elementSizeInImage)
+                imgPieceStart := imgStart + x * elementSizeInImage
+                imgPieceEnd := imgPieceStart + elementSizeInImage
                 if isSrcBuffer {
                   copy(imageLevel.Data[imgPieceStart:imgPieceEnd], piece.DeviceMemory.Data[bufMemStart:bufMemEnd])
```
AWoloszyn, line 697: I imagine this is really going to hurt your pools as well. (Just an aside for performance) You are going to end up with one slice in the pool per pixel in the image :\

gapis/api/vulkan/api/copy_clear_commands.api
pmuetschard: because it can have different values in different parts of the code.

gapis/api/vulkan/api/copy_clear_commands.api
pmuetschard: Yes, it still slow, but much faster than before. I'm mulling some other ideas that I will bounce off of people, when they are a bit more formed.

Do you see any issues with the change otherwise?

AWoloszyn: Nope this looks reasonable as far as I can tell.

gapis/api/vulkan/api/copy_clear_commands.api
pmuetschard: also, and sorry if that's what you originally meant to ask about, not using `4` here, even though we know it's `4` is because `elementSizeInBuffer` is more descriptive (esp next to `elementSizeInImage`) than `4`.

pau-baiget: Would you care about adding an assert there to make sure it's _always_ 4?

pmuetschard:
ping @purvisa-at-google-com

gapis/api/vulkan/api/copy_clear_commands.api
```diff
@@ -676,19 +676,23 @@ sub void copyImageBuffer(VkBuffer buffer, VkImage image, VkImageLayout layout, m
...
-              readMemoryInBuffer(bufferObject, bufStart, as!VkDeviceSize(elementSizeInBuffer))
-              // Discard the 4th byte in the buffer for current pixel.
-              bufMemPieces := getBufferBoundMemoryPiecesInRange(bufferObject, bufStart, as!VkDeviceSize(elementSizeInImage))
-              for _ , _ , piece in bufMemPieces {
-                bufMemStart := piece.MemoryOffset
-                bufMemEnd := bufMemStart + piece.Size
-                imgPieceStart := (piece.ResourceOffset - bufStart) + imgStart
-                imgPieceEnd := imgPieceStart + piece.Size
+            for _ , _ , piece in bufMemPieces {
+              // The below assumes that each memory piece is alligned to a multiple of
```
purvisa-at-google-com, line 686: nit: alligned -> aligned

gapis/api/vulkan/api/copy_clear_commands.api
purvisa-at-google-com: I guess getBufferBoundMemoryPiecesInRange is the real cost we have hoisted out of the loop? It's worrying that it's more expensive than having a pool entry per pixel!

If it's empirically faster then we should commit, but I agree. It doesn't sound like we've got all the performance pathologies out of this code just yet. Can we write the 4th pixel channel just to get all the pool entries to consolidate? I think you talked to me about this already.

AWoloszyn: So this change does not CHANGE the problem with pool write per pixel. It is just ANOTHER thing that would make this code faster.

purvisa-at-google-com: Ah, ok, that makes more sense!

pmuetschard: There currently are two per-pixel things, this removes one of them. Removing the other is a bigger change that has a wider effect and needs changes in a few places.

gapis/api/vulkan/api/copy_clear_commands.api
pmuetschard: Done.

pmuetschard:
@purvisa-at-google-com  PTAL.

gapis/api/vulkan/api/copy_clear_commands.api
pmuetschard: We don't have an assert mechanism in API file, as far as I know.

PR #355, opened 2020-06-29
Use GCC 8 on Windows builds

hevrard:
This is to stay in sync with the version used in Linux builds.

Bug: b/157873397

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -55,9 +55,9 @@ Using the msys64 shell (installed by choco at `C:\tools\msys64\mingw64`):
 1. Update MSYS with: `pacman -Syu`.
 2. If the update ends with “close the window and run it again”, close and reopen the window and repeat 1.
 3. Fetch required tools with: `pacman -S curl git zip unzip patch`
-4. Download gcc with: `curl -O http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-gcc-9.2.0-2-any.pkg.tar.xz`
-5. Download gcc-libs with: `curl -O http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-gcc-libs-9.2.0-2-any.pkg.tar.xz`
-6. Install gcc with: `pacman -U mingw-w64-x86_64-gcc*-9.2.0-2-any.pkg.tar.xz`
+4. Download gcc with: `curl -O http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-gcc-8.3.0-2-any.pkg.tar.xz`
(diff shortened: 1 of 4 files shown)


johnplate:
There is a valid reason to stay on gcc8 for the Linux releases. The question now is how many of our team are using gcc9 on Linux because it is the default gLinux compiler? If most of us are using gcc9, why not use gcc9 also on Windows?

hevrard:
Thanks both for the comments, it prompted me to lookup why we didn't use gcc9 on Linux: https://github.com/google/agi/pull/37 reminded me that using gcc9 on Linux leads to a dependency on a recent version on GLIBC that may not be present on various Linux installs, and that is not present on the Swarming bots in particular.

So to recap: we use gcc8 on Linux to be able to use the builds on Swarming, and here I downgrade to gcc8 on windows for the sake of using the same version between these two platforms, as discussed in b/157873397.

Now we also had some issue with grpc, see https://github.com/google/agi/pull/319 , that prompted the use of gcc9 on windows. Let's see if this change compiles fine on windows, in which case we can use gcc8 everywhere. If not, we can properly document why we don't have the same versions.

johnplate:
We can also try to link executables on Linux with the flag -static-libgcc. I remember I used it years ago to solve the issue with the dependency.

hevrard:
> We can also try to link executables on Linux with the flag -static-libgcc. I remember I used it years ago to solve the issue with the dependency.

That may be possible technically, yet it would not make things simpler. The main goal of this PR is to reduce discrepancies by using the same gcc major version on both windows and linux, and this works with gcc8 without compiler flag tweaks. Hence I think it is better to use gcc8 rather than adding some special flags to enable gcc9.

johnplate:
I don't mind to switch to gcc8 on Windows, but I don't think it will make things easier, because I assume that most of our development environments both on Linux and Windows are currently using gcc9, especially since the gLinux default was switched to gcc 9.3. It is not helpful if kokoro uses a different major compiler version than the developers.

hevrard:
I get your point of having Kokoro builds be close to what developers use, but I would rather ask developers to install gcc8 than trying to statically link a glibc, which is a notorious can of worm (see e.g. https://stackoverflow.com/a/26306630). More generally, Kokoro builds are done using not-too-recent OSes and compilers in order to produce binaries that are usable on not-too-recent platforms, like Swarming bots.

johnplate:
We should discuss this topic with the team before we ask them all to change their compilers

hevrard:
See go/agi-gcc-version

hevrard:
After chatting with the team, the decision is to keep gcc9 on Windows in order to catch gcc9-related bugs in the CI.

PR #356, opened 2020-06-29
Remove Graphics driver override.

lpy:
Previously we added an override to allow fast development and iteration
with partners, now that the integration is done, remove this hack.
Rationale: Graphics driver has special permission than other components,
for example application apk. Having a way to override this exposes some
security concerns and we will trap ourselves by having to do hacks in
order to make the namespace work properly.

Minor: Do not return the prerelease driver if it's preinstalled, because
it is empty.

Bug: b/148970509

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -48,7 +48,6 @@ const (
 
 	perfettoPort = NamedFileSystemSocket("/dev/socket/traced_consumer")
 
-	driverOverride = "agi.driver_package_override"
 	driverProperty = "ro.gfx.driver.1"
 )
 
@@ -443,7 +442,9 @@ func (b *binding) resolveDriverPath(ctx context.Context, driver string) (Driver,
 		return Driver{}, nil
 	}
 	path = path[8:]
-	if path == "" {
+	// If the driver package path doesn't have the /data/app prefix, it means the returned
+	// path is the preinstalled emtpy prerelease driver.
+	if path == "" || !strings.HasPrefix(path, "/data/app") {
 		return Driver{}, nil
 	}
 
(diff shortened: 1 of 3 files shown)


core/os/android/adb/commands.go
```diff
@@ -443,7 +442,9 @@ func (b *binding) resolveDriverPath(ctx context.Context, driver string) (Driver,
 		return Driver{}, nil
 	}
 	path = path[8:]
-	if path == "" {
+	// If the driver package path doesn't have the /data/app prefix, it means the returned
+	// path is the preinstalled emtpy prerelease driver.
+	if path == "" || !strings.HasPrefix(path, "/data/app") {
```
pmuetschard, line 447: Should we do this the other way around, i.e. `if strings.HasPrefix(path, "/vendor/app")` ignore it?

lpy: It is vendor dependent, vendors can put it inside `/system/priv-app` as well.

lpy:
> Please keep `GraphicsDriver` and drop `PrereleaseGraphicsDriver`, which should have been called `GraphicsDriverIgnoringOverride`, sigh.

Wouldn't that be confusing since `GraphicsDriver` returns a path to an apk instead of something `/vendor/lib64/`?

pmuetschard:
> Wouldn't that be confusing since GraphicsDriver returns a path to an apk instead of something /vendor/lib64/?

I don't understand that comment. All I was asking is that instead of changing all the references to `PrereleaseGraphicsDriver` keep the `GraphicsDriver` function and drop `PrereleaseGraphicsDriver` rather than keeping `PrereleaseGraphicsDriver` and dropping `GraphicsDriver`. I.e. just rename your new `PrereleaseGraphicsDriver` to `GraphicsDriver`.

lpy:
> I don't understand that comment. All I was asking is that instead of changing all the references to `PrereleaseGraphicsDriver` keep the `GraphicsDriver` function and drop `PrereleaseGraphicsDriver` rather than keeping `PrereleaseGraphicsDriver` and dropping `GraphicsDriver`. I.e. just rename your new `PrereleaseGraphicsDriver` to `GraphicsDriver`.

My bad, sorry for the confusing wordings. What I mean is that `GraphicsDriver` only returns the info about the prerelease driver, so why not make it more explicit even though I have to change a bunch of references? Or do you anticipate we may end up extending this function to return info about other graphics drivers?

pmuetschard:
> Or do you anticipate we may end up extending this function to return info about other graphics drivers?

My idea is that what ever logic we'll use that selects what driver to use, will live in this function. We shouldn't really have had more than one version in the first place.

lpy:
Thanks! PTAL!

core/os/android/adb/commands.go
lpy: Marking this as resolved. :)

PR #358, opened 2020-06-29
Fix a couple of straggling user-facing "GAPID" references.

pmuetschard:


lpy:
Driven-by comment: why not take this chance to rename GapidVersion.in?

gapic/src/main/com/google/gapid/util/Flags.java
```diff
@@ -130,7 +130,7 @@ private static boolean isBooleanFlag(Flag<?> flag) {
   }
 
   private static void printVersion(PrintStream out) {
-    out.println("GAPID version " + GAPID_VERSION);
+    out.println("AGI version " + GAPID_VERSION);
```
hevrard, line 133: Also rename GAPID_VERSION to AGI_VERSION ?

pmuetschard:
Neither of those are user-visible, so it can wait.

PR #360, opened 2020-06-30
Fix two draw command grouping bugs.

stellama0208:
- Fix wrong cmd nesting relationship between state setting cmds 
and renderpass cmds, when draw grouping get interrupted by 
renderpass/debugmarker grouping.
- Fix wrong cmd nesting for continuous draw commands scenario, 
which is resulted from extra stack popping.
- Bug: b/159466831.

### DIFF
--- a/gapis/api/vulkan/vulkan.go
+++ b/gapis/api/vulkan/vulkan.go
@@ -238,6 +238,36 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
 			d.SubcommandLookup.SetValue(nv, ref)
 			refs = append(refs, ref)
 
+			// Handle draw commands grouping.
+			cmdName := cb.CommandReferences().Get(uint32(i)).Type().String()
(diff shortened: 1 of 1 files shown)


stellama0208:
The above force push only updates the commit messages.

hliatis:
In the Subpasses demo from Sascha WIllems, some of the grouping still looks confusing

stellama0208:
> In the Subpasses demo from Sascha WIllems, some of the grouping still looks confusing

Thank you Hernan for testing it out on Subpasses SW sample! I have been concerned about possible command list with a non-stack-structured grouping for a while, but haven't been able to really find a corner case, but this is the one! 

Subpasses SW sample's problem requires a deeper and heavier design change, I've documented the causing problem in b/160276657. Since that change would require a design doc first, I will confine the scope of this PR to improving compatibility of draw grouping with other types of groupings. 

This PR will be merged to deal with some existing grouping issue that can be observed even in stack-structured command list. (Observable in Bloom SW sample and NetEase Viking Village sample).

PR #362, opened 2020-06-30
Make sure renderpasses with fewer attachments than later ones show FBs

hliatis:
Bug: b/160157241


hliatis:
Made a new push since I didn't handle the case of zero attachments being added

PR #364, opened 2020-07-01
* Fourth part of the new transform system

yalcinmelihyasin:
It contains:

- FrameBuffer Replay and all the transforms that required for this replay type.

This will be followed by the timestamp query and some other minor changes..

Currently, it contains code duplication as we need to keep old transform
system running but they will be removed when the transition is over.

For Q&A:
go/agi-transform-review


yalcinmelihyasin:
There will be a benchmark comparison document to be added here.

hevrard:
Some tests are failing, see Ci-linux-test log:
```
/tmpfs/tmp/bazel_out/sandbox/linux-sandbox/4210/execroot/gapid/gapis/api/transform2/transform_test.go:35:39: cannot use &multiplier_transform literal (type *multiplier_transform) as type transform2.Transform in slice literal:
	*multiplier_transform does not implement transform2.Transform (missing RequiresInnerStateMutation method)
/tmpfs/tmp/bazel_out/sandbox/linux-sandbox/4210/execroot/gapid/gapis/api/transform2/transform_test.go:37:42: not enough arguments in call to transform2.CreateTransformChain
	have (commandGenerator.CommandGenerator, []transform2.Transform, *testWriter)
	want (context.Context, commandGenerator.CommandGenerator, []transform2.Transform, transform2.Writer)
/tmpfs/tmp/bazel_out/sandbox/linux-sandbox/4210/execroot/gapid/gapis/api/transform2/transform_test.go:52:39: cannot use &multiplier_transform literal (type *multiplier_transform) as type transform2.Transform in slice literal:
	*multiplier_transform does not implement transform2.Transform (missing RequiresInnerStateMutation method)
/tmpfs/tmp/bazel_out/sandbox/linux-sandbox/4210/execroot/gapid/gapis/api/transform2/transform_test.go:54:42: not enough arguments in call to transform2.CreateTransformChain
	have (commandGenerator.CommandGenerator, []transform2.Transform, *testWriter)
	want (context.Context, commandGenerator.CommandGenerator, []transform2.Transform, transform2.Writer)
```

yalcinmelihyasin:
Here are some results from the `gapit benchmark` command.

Old is the commit `ffcead0e9616e3e9488596930d6192f1172cc9a9` in master and the new is the pr directly applied to the particular commit. 

- The max memory usage is affected from when the GC is triggered so Andrew suggested me to use the highest value.
- As the values are affected from the external factors e.g. clock speed, I have put 3 metrics that is related with replay and 1 metric related with trace. So that it is clear that even without any code effecting the trace, results can be quite different. 
- I do not see any significant change between the new results and the old ones. Please point if you thing there are any.

```
                                         Old                New      

Highest Max Memory Usage :             1306 MB            1334 MB
Average Before MEC Frame Time :        60.2 ms             63.2 ms
Average Trailing Frame Time :          140.5 ms           145.9 ms
Average Interaction :                   3.5 s              3.63 s
____________________________________________________________________

Average Trace Frame Time :            325.88 ms           340.46 ms
```

yalcinmelihyasin:
I have updated the `average interactive` metric. It is interesting that while nearly every other metric is going down, it went up. Although most of it comes from just one spike during the old replay, might be an outlier.

yalcinmelihyasin:
And numbers are average of 5 benchmark for each.

gapis/api/vulkan/transform_read_framebuffer.go
```diff
@@ -51,73 +51,103 @@ type pendingRead struct {
...
-	if len(t.pendingReads) > 0 {
-		if id != api.CmdNoID {
-			return t.FlushPending(ctx, out)
+
+		if _, ok := cmd.(*VkQueueSubmit); ok {
+			if len(framebufferTransform.pendingReads) > 0 {
+				if err := framebufferTransform.FlushPending(ctx, id, inputState); err != nil {
+					return nil, err
+				}
+			}
```
hevrard, line 135: There seems to be a change of behavior here, the old code called FlushPending as soon as `id != api.CmdNoID`, the new code waits for a VkQueueSubmit. Is this intended?

gapis/api/transform2/transform_chain.go
```diff
@@ -105,29 +123,35 @@ func (chain *TransformChain) endChain(ctx context.Context) error {
...
+		}
 
-	for _, transform := range chain.transforms {
-		cmds, err = transform.TransformCommand(ctx, id, cmds, chain.out.State())
+		chain.currentTransformIndex = i
+		var err error
+		inputCmds, err = transform.TransformCommand(ctx, id, inputCmds, chain.out.State())
 		if err != nil {
-			log.W(ctx, "Error on Transform on cmd [%v:%v] with transform [%v] : %v", id, inputCmd, transform, err)
+			log.W(ctx, "Error on Transform on cmd [%v:%v] with transform [:v:%v] : %v", id, inputCmds, i, transform, err)
```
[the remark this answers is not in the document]
yalcinmelihyasin: It would, because with cmdID, index of the transform and the existing input command, we would have a good idea of where it failed.

gapis/api/vulkan/transform_capture_log.go
```diff
@@ -50,6 +50,14 @@ func (logTransform *captureLog) RequiresAccurateState() bool {
 	return false
 }
 
+func (logTransform *captureLog) RequiresInnerStateMutation() bool {
+	return false
+}
+
+func (logTransform *captureLog) SetInnerStateMutationFunction(mutator transform2.StateMutator) {
+	// This transform do not require inner state mutation
```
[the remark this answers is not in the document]
yalcinmelihyasin: Hmmm. We can discuss this. It may or may not be a good idea.

gapis/api/vulkan/transform_wireframe.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (wireframe *wireframeTransform) EndTransform(ctx context.Context, inputCommands []api.Cmd, inputState *api.GlobalState) ([]api.Cmd, error) {
+	return inputCommands, nil
+}
+
+func (wireframe *wireframeTransform) ClearTransformResources(ctx context.Context) {
+	// Melih TODO: This memory never been released.
+	// Check if it's intentional
+	// wireframe.allocations.FreeAllocations()
```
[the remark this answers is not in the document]
yalcinmelihyasin: There are a few of these, I will talk with awolozsyn@ tomorrow.

gapis/api/transform2/transform_chain.go
```diff
@@ -105,29 +123,35 @@ func (chain *TransformChain) endChain(ctx context.Context) error {
...
-	if err = mutateAndWrite(ctx, id, cmds, chain.out); err != nil {
+	if err := mutateAndWrite(ctx, id, inputCmds, chain.out); err != nil {
 		return err
 	}
 
-	for _, transform := range chain.transforms {
-		transform.ClearTransformResources(ctx)
+	if beginTransformIndex == 0 {
+		for _, transform := range chain.transforms {
+			transform.ClearTransformResources(ctx)
```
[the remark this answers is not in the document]
yalcinmelihyasin: So this guarantees that we don't clear resources when we branched off from our transform chain.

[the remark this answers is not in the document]
yalcinmelihyasin: And it only clears the resources when all the transforms for that command has processed and mutated. Any temporal resources for the branched mutation should be cleared explicitly in transform.

gapis/api/vulkan/transform_read_framebuffer.go
```diff
@@ -51,73 +51,103 @@ type pendingRead struct {
...
-		if r, ok := t.injections[idx_string]; ok {
-			// If this command is FOR an EOF command, we want to mutate it, so that
-			// we have the presentation info available.
-			if cmd.callee != nil && cmd.callee.CmdFlags().IsEndOfFrame() {
-				cmd.callee.Mutate(ctx, id, out.State(), nil, nil)
-			}
-			for _, injection := range r {
-				if err := injection.fn(ctx, cmd, injection.res, out); err != nil {
-					return err
+func (framebufferTransform *readFramebuffer) TransformCommand(ctx context.Context, id api.CmdID, inputCommands []api.Cmd, inputState *api.GlobalState) ([]api.Cmd, error) {
```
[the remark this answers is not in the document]
yalcinmelihyasin: The ones that I have kept are the mostly the methods that I did not touch. Maybe we should rename them all sometime in the future but I think we should steer away from one letter variables where it makes sense.

gapis/api/vulkan/transform_read_framebuffer.go
yalcinmelihyasin: Yes, because actually the intention of the old code was doing "FlushPending" on the VkQueueSubmit because the only way you can call flushPending while having a pending read is when you have vkQueueSubmit.

This is the scenario: 
   - In command splitter you add an insertion command while you are transforming VkQueueSubmit and you replace it here with an actual command list(see color and depth functions).
   - InsertionCommand you inject creates a pending read and the first command without an id of cmdNoID is the vkQueueSubmit that actually generated that insertion command in command splitter.

I couldn't see if there would be a pending read without a command splitter. But I will double check with Andrew just in case.

gapis/api/transform2/transform_chain.go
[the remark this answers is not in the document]
yalcinmelihyasin: That's actually a very valid point. I will think about it how we can show it more clear. Maybe we can show it like, here this is the cmdID, this is the cmd, this is the transform index, this is the transform and this is the latest status of input commands.

gapis/api/vulkan/transform_command_splitter.go
```diff
@@ -153,123 +96,377 @@ type commandSplitter struct {
...
+	}
+
+	insertionCommand := &InsertionCommand{
+		VkCommandBuffer(0),
+		append([]VkCommandBuffer{}, t.pendingCommandBuffers...),
+		topCut,
+		inputCommands[endOfFrameCmdID],
+	}
+
+	if isEndOfFrame {
```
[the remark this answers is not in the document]
yalcinmelihyasin: Yes, we can put a comment here and what you say makes sense as a reason. But let's take the Andrew's input if there is any other reason.

gapis/api/transform2/transform_chain.go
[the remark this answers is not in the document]
yalcinmelihyasin: Let's do this in the next PR because there are things that we have to consider and let's talk about them when I am back from vacation. As it's already the behaviour in master, I don't think this requires very immediate attention.

gapis/api/transform2/transform.go
```diff
@@ -45,8 +49,15 @@ type Transform interface {
 	// Resources are needed for the state mutation, therefore this should be called after mutation.
 	ClearTransformResources(ctx context.Context)
 
-	// RequiresAccurateState returns true if transform needs the observe the accurate state.
+	// RequiresAccurateState returns true if the transform needs to observe the accurate state.
 	RequiresAccurateState() bool
+
+	// RequiresInnerStateMutation returns true if the transform needs to mutate state during the transformation
+	RequiresInnerStateMutation() bool
```
purvisa-at-google-com, line 56: I'm not sure about this name. It's pretty cryptic. What about ReadsStateBetweenCommands() or DoesYield()?

gapis/api/transform2/transform.go
```diff
@@ -45,8 +49,15 @@ type Transform interface {
...
+
+	// SetInnerStateMutationFunction sets a mutator function for making a branch in the transform
+	// to transform and mutate pending commands to be able to continue transform.
+	SetInnerStateMutationFunction(stateMutator StateMutator)
```
purvisa-at-google-com, line 60: With my C++ head on I'm not sure I like this being public. But I guess we have fewer choices in go.

gapis/api/transform2/transform_chain.go
```diff
@@ -26,30 +26,48 @@ import (
...
+		if t.RequiresInnerStateMutation() {
+			t.SetInnerStateMutationFunction(chain.mutator)
+		}
+	}
+
+	return &chain
 }
 
 func (chain *TransformChain) beginChain(ctx context.Context) error {
-	chain.handleInitialState(chain.out.State())
```
purvisa-at-google-com, line 51: Why does this get deleted? Seems unrelated to the CL

gapis/api/transform2/transform_chain.go
```diff
@@ -105,29 +123,35 @@ func (chain *TransformChain) endChain(ctx context.Context) error {
 	return nil
 }
 
-func (chain *TransformChain) transformCommand(ctx context.Context, inputCmd api.Cmd, id api.CmdID) error {
-	var err error
-	cmds := []api.Cmd{inputCmd}
+func (chain *TransformChain) transformCommands(ctx context.Context, id api.CmdID, inputCmds []api.Cmd, beginTransformIndex int) error {
+	for i, transform := range chain.transforms {
```
purvisa-at-google-com, line 127: Could this be rewritten with an explicit iteration variable to be shorter and not need to "continue" over the first "beginTransformIndex" commands?

gapis/api/transform2/transform_chain.go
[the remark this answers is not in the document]
purvisa-at-google-com: A comment to make this explicit would be nice! Maybe we can phrase it in terms of "If we are in the middle of processing some commands through the chain, then it's too early to release resources. Instead let's wait until the start of the next pass though the chain to clear"?

gapis/api/transform2/transform_chain.go
```diff
@@ -157,23 +181,27 @@ func (chain *TransformChain) GetNextTransformedCommands(ctx context.Context) err
...
+	err := chain.transformCommands(ctx, chain.currentCommandID, inputCmds, 0)
 	if err != nil {
 		log.E(ctx, "Replay error (%v:%v): %v", chain.currentCommandID, currentCommand, err)
 	}
+
 	chain.currentCommandID++
 	return err
 }
 
-func (chain *TransformChain) handleInitialState(state *api.GlobalState) (*api.GlobalState, error) {
```
purvisa-at-google-com, line 168: I already mentioned it above, but even though this function does nothing at the moment, I'd prefer to keep it because it makes a strong statement of intent about what will happen in the future and where it should happen.

gapis/api/transform2/transform_chain.go
```diff
@@ -157,23 +181,27 @@ func (chain *TransformChain) GetNextTransformedCommands(ctx context.Context) err
...
-	for _, t := range chain.transforms {
-		if t.RequiresAccurateState() {
-			// Melih TODO: Temporary check until we implement accurate state
-			panic("Implement accurate state")
-		}
+func (chain *TransformChain) stateMutator(ctx context.Context, id api.CmdID, cmds []api.Cmd) error {
+	beginTransformIndex := chain.currentTransformIndex + 1
```
purvisa-at-google-com, line 196: I'd add some comments making the saving and restoring of member variables to the call stack as a means of recursion explicit. It's very easy to miss reading this. It's a very short and simple looking function, but actually hides quite a lot of subtle complexity. Would be nice to document!

gapis/api/transform2/transform_test.go
```diff
@@ -112,6 +112,14 @@ func (t *multiplier_transform) RequiresAccurateState() bool {
 	return false
 }
 
+func (t *multiplier_transform) RequiresInnerStateMutation() bool {
+	return false
+}
+
+func (t *multiplier_transform) SetInnerStateMutationFunction(stateMutator transform2.StateMutator) {
```
purvisa-at-google-com, line 119: It's a shame that every transform needs to implement this function. I guess go style polymorphism makes it hard to avoid this.

purvisa-at-google-com: Actually can we just pass this in to TransformCommands as an argument. The transforms do not need to store it locally since they will only call it inside that one function. So it could just be provided from the outside to avoid all these trivial functions.

gapis/api/vulkan/transform_capture_log.go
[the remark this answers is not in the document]
purvisa-at-google-com: See my comment above on this function. I'm not sure we should have it in the first place. There are a lot of redundant implementations of "do nothing" in each transform. I think we can centralise this functionality on the transform chain so it only exists in one place!

gapis/api/transform2/transform.go
yalcinmelihyasin: It's not a read, it's actually a write operation on state. It would be a misleading name. For yield, I think it is already have a well known meaning and this really does not satisfy the meaning(as there is no actual yield).

We may change the name but I think it should be a significantly better name in terms of explaining what is happening.

gapis/api/transform2/transform.go
yalcinmelihyasin: I am not very comfortable with Go but I am not sure if it is not public then can you implement this interface from another package because the transforms are under Vulkan package not transform.

gapis/api/transform2/transform_chain.go
yalcinmelihyasin: I think we had talked about it. This is not in use therefore I removed it.

gapis/api/transform2/transform_chain.go
yalcinmelihyasin: If it matters yes we can but I would suggest to adding it to the next PR so that we don't wait for a rebuild. I don't think it is very important.

gapis/api/transform2/transform_chain.go
yalcinmelihyasin: Well, I think we have talked about not copying the state because it is very expensive to do so. Do we really want to copy it in the future? I thought that we had dropped the idea completely.

gapis/api/transform2/transform_chain.go
yalcinmelihyasin: Sure. But as parallel with the previous command, can we add it to the next PR. I am planning to open it today.

gapis/api/transform2/transform_test.go
yalcinmelihyasin: Let's talk about this in detail later and if we decide it can go to another PR.

gapis/api/transform2/transform_chain.go
purvisa-at-google-com: It's required for functions that NeedAccurateState(). Cloning is expensive, but we can't have accurate state without doing it. So we will need to implement this in the future.

gapis/api/transform2/transform_chain.go
purvisa-at-google-com: I've commented below. I believe we should keep it in as a stub. We will need to implement this functionality later for transforms that NeedAccurateState()

gapis/api/transform2/transform.go
purvisa-at-google-com: If we do as commented below and pass this in to TransformCommands as an extra argument, then the problem goes away. I think we should do that.

gapis/api/transform2/transform.go
purvisa-at-google-com: Again if we pass the state mutator in to TransformCommands() as an extra argument this can just go away. Then the name is irrelevant.

purvisa-at-google-com: Sorry, that comment is in reference to comments below!

gapis/api/transform2/transform.go
```diff
@@ -23,6 +23,10 @@ import (
 	"github.com/google/gapid/gapis/api"
 )
 
+// StateMutator is a function that allows transformation to mutate state during transformation
+// TODO: Refactor the transforms that uses this to remove this behaviour
+type StateMutator func(id api.CmdID, cmds []api.Cmd) error
```
purvisa-at-google-com, line 28: This function should not take an "id" argument. That should be stored inside the TransformChain so that a transform cannot change the ID when it performs a state mutation. This design was created in-part to standardise command id generation to a single source of truth within the program and the argument here undoes that aspect of the design. See the implementation of "func (chain *TransformChain) stateMutator" for how a Transform currently transforming commands for ID=5 may pass its output on to a next transform in the chain under an incorrect ID of say 66.

gapis/api/vulkan/transform_command_splitter.go
```diff
@@ -0,0 +1,934 @@
...
+	}
+
+	commandBufferAllocateInfo := NewVkCommandBufferAllocateInfo(inputState.Arena,
+		VkStructureType_VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO, // sType
+		NewVoidᶜᵖ(memory.Nullptr),                                      // pNext
+		commandPoolID,                                                  // commandPool
+		VkCommandBufferLevel_VK_COMMAND_BUFFER_LEVEL_PRIMARY,           // level
+		1, // commandBufferCount
+	)
+	commandBufferID := VkCommandBuffer(newUnusedID(true, func(x uint64) bool {
```
AWoloszyn, line 833: These read the state, Have you made sure to mutate up to this point in order to not get a clashing commandbufferID?

PR #365, opened 2020-07-01
Fix wrong tab layout when user open AGI the first time.

stellama0208:
- b/157252944.


stellama0208:
To reproduce the bug and test out the PR, we could delete the file `.apic` under the home directory, to mimic opening AGI for the first time.

hevrard:
There was some CI flakes, I restarted the failing tracks.

PR #366, opened 2020-07-03
Match replay device vulkan driver version with trace

hevrard:
Do not allow to replay on a device which reports a different Vulkan
driver version. Also, show the driver version in the replay device
selection dialog.

Bug: b/148563197

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Devices.java
+++ b/gapic/src/main/com/google/gapid/models/Devices.java
@@ -250,7 +250,7 @@ public class Devices {
   public static String getVulkanDriverVersions(Device.Instance dev) {
     StringBuilder version = new StringBuilder("N/A");
     VulkanDriver vkDriver = dev.getConfiguration().getDrivers().getVulkan();
-    Boolean first = true;
+    boolean first = true;
     for (int i = 0; i < vkDriver.getPhysicalDevicesCount(); i++) {
       if (first) {
         version.setLength(0);
--- a/gapic/src/main/com/google/gapid/models/Devices.java
+++ b/gapic/src/main/com/google/gapid/models/Devices.java
@@ -29,6 +29,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.google.gapid.proto.SettingsProto;
 import com.google.gapid.proto.device.Device;
 import com.google.gapid.proto.device.Device.Instance;
+import com.google.gapid.proto.device.Device.VulkanDriver;
 import com.google.gapid.proto.service.Service;
 import com.google.gapid.proto.service.Service.Value;
 import com.google.gapid.proto.service.path.Path;
@@ -244,6 +245,15 @@ public class Devices {
 
   public void removeListener(Listener listener) {
     listeners.removeListener(listener);
+  }
+
+  public static String getVulkanDriverVersion(Device.Instance dev) {
+    String version = "N/A";
+    VulkanDriver vkDriver = dev.getConfiguration().getDrivers().getVulkan();
+    if (vkDriver.getPhysicalDevicesCount() > 0) {
+      version = Integer.toUnsignedString(vkDriver.getPhysicalDevices(0).getDriverVersion());
+    }
+    return version;
   }
 
   public static String getLabel(Device.Instance dev) {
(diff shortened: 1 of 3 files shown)


gapis/api/vulkan/replay.go
```diff
@@ -90,6 +90,9 @@ func (a API) GetReplayPriority(ctx context.Context, i *device.Instance, h *captu
 				if devPhyInfo.GetDeviceId() != tracePhyInfo.GetDeviceId() {
 					continue
 				}
+				if devPhyInfo.GetDriverVersion() != tracePhyInfo.GetDriverVersion() {
```
pau-baiget, line 93: So when something fails we always return 1 here. Does this allow to return a proper error message to the user that's different when the device doesn't match and when the driver doesn't match?

hevrard: When a device is compatible we return 1, else we return 0 (the `return 1` is actually reached only if all checks pass). The semantic of GetReplayPriority is to return a priority which must be > 0 for the device to be considered valid.

This change does not enable us to return a precise error message for the reason why the device is not considered valid. The current architecture is such that for a given trace, the server returns the list of valid replay devices. Invalid replay devices are not listed, and so there is not even a chance to explain why they might be invalid. So the current architecture does not enable to say "we see you have a deviceABC plugged in, but it is not suitable for this trace". This limitation is tackled in "phase 2" of go/agi-replay-device.

The bug was originally filed because AGI would let the user try and replay on a device which had a different Vulkan driver version, this PR here disallow this, so I think it is a good first fix.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -246,6 +247,15 @@ public void removeListener(Listener listener) {
     listeners.removeListener(listener);
   }
 
+  public static String getVulkanDriverVersion(Device.Instance dev) {
+    String version = "N/A";
+    VulkanDriver vkDriver = dev.getConfiguration().getDrivers().getVulkan();
+    if (vkDriver.getPhysicalDevicesCount() > 0) {
+      version = Integer.toUnsignedString(vkDriver.getPhysicalDevices(0).getDriverVersion());
+    }
```
stellama0208, line 255: It seems that the result returned here may be different from what agi actually used in the gapis? 
We directly return the version of the first vulkan physical device here, while in gapis we iterate and do compatibility checks which may actually pick other vulkan physical devices? 
If it's true, before we find a mechanism to transfer back the information of which driver we actually used, would it be better if we concat all the possible driver versions together and show them here? (If too long maybe hide with a shortened string and show in full at hover?)

hevrard: Yes, we list only the driver version of the first physical device, I assumed it was OK since the few mobile devices that we support only have one Vulkan physical device, so the trace will report only one vk physical device.

But your point is valid, I updated the PR to list the driver version of all vulkan physical devices recorded in the trace. I think it's OK to make the string long in that case, because multi-device capture would be very rare (I'm not even sure it is properly supported by AGI now).

gapis/api/vulkan/replay.go
pau-baiget: You're right i wrongly read the code, thanks for the thorough explanation!

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -246,6 +247,22 @@ public void removeListener(Listener listener) {
     listeners.removeListener(listener);
   }
 
+  public static String getVulkanDriverVersions(Device.Instance dev) {
+    StringBuilder version = new StringBuilder("N/A");
+    VulkanDriver vkDriver = dev.getConfiguration().getDrivers().getVulkan();
+    Boolean first = true;
```
stellama0208, line 253: nit: can use `boolean` instead in the future (in Java, primitive types started with a capital letter (`Boolean`) are boxed type, they would have some affiliated helper functions, if in no need of those functions, then the non-capital types (`boolean`) would be enough.

hevrard: Good catch, I've updated to a simple `boolean`, thanks!

PR #367, opened 2020-07-06
Restart GAPIR only upon a new replay request

hevrard:
When a ping to the replayer fails, just remove this replayer, do not
proactively restart it. Removing it from the Client replayers list
ensures that it will be restarted upon the Connect() of the next
replay request.

Bug: b/159997874
Test: manual

### DIFF
--- a/gapir/client/client.go
+++ b/gapir/client/client.go
@@ -152,22 +152,14 @@ func (client *Client) Connect(ctx context.Context, device bind.Device, abi *devi
 	return &key, nil
 }
 
-// removeConnection closes and removes a GAPIR instance.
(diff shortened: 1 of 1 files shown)


yalcinmelihyasin:
I am not sure if this the right direction to take. What is the advantage that we gain from this?

hevrard:
@yalcinmelihyasin see the motivation in b/159997874. The rationale is that currently, if you want to e.g. capture a new trace on a device where you did the replay, the new trace will fail because the replayer will be restarted and take over the app being traced.

yalcinmelihyasin:
I see. Makes sense.

PR #368, opened 2020-07-06
Migrate LUCI tools from lucy-py to luci-go

hevrard:
LUCI tools have been reimplemented in Go, and now only the Go version
is recommended. This update follows a request from the LUCI team.

Fetch the luci-go tools using CIPD, as recommended, and update various
flags to match the new tools.

Bug: b/159313821
Test: test/swarming/manual-run.sh

### DIFF
--- a/kokoro/linux/build.sh
+++ b/kokoro/linux/build.sh
@@ -114,8 +114,8 @@ export PATH="`pwd`/depot_tools:$PATH"
 # Note that the '${platform}' must appear as-is in the ensure file,
 # hence the single quotes.
 (
-  echo 'infra/tools/luci/isolate/${platform} git_revision:227ed2302d4b77f5a9823c223d5c1571b45dafd4'
-  echo 'infra/tools/luci/swarming/${platform} git_revision:227ed2302d4b77f5a9823c223d5c1571b45dafd4'
+  echo 'infra/tools/luci/isolate/${platform} git_revision:fe882fb29f1eaa8d1a4cfc77af4d11bcf7d59318'
+  echo 'infra/tools/luci/swarming/${platform} git_revision:fe882fb29f1eaa8d1a4cfc77af4d11bcf7d59318'
 ) > ${LUCI_ROOT}/ensure_file.txt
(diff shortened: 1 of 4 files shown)


kokoro/linux/build.sh
```diff
@@ -101,12 +101,23 @@ $SRC/kokoro/linux/package.sh $BUILD_ROOT/out
...
+git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+export PATH="`pwd`/depot_tools:$PATH"
+
+# You can get valid git revision by looking up infra/tools/... at:
+# https://chrome-infra-packages.appspot.com/
+# Note that the '${platform}' must appear as-is in the ensure file,
+# hence the single quotes.
+(
+  echo 'infra/tools/luci/isolate/${platform} git_revision:227ed2302d4b77f5a9823c223d5c1571b45dafd4'
+  echo 'infra/tools/luci/swarming/${platform} git_revision:227ed2302d4b77f5a9823c223d5c1571b45dafd4'
```
pau-baiget, line 118: Just double checking you meant to put the same revision here.

hevrard: Yep, it's the same revision.

test/swarming/collect.py
```diff
@@ -132,10 +131,8 @@ def main():
         f.write(json.dumps(results, sort_keys=True, indent=2))
 
     #### Exit code
-    # Return non-zero if the task failed, but silence failures due to expiration
-    # of an internal failure, because we do not want Swarming infrastructure
-    # failures to be reported as build failures.
-    if status == 'pass' or status == 'expired' or status == 'internal_failure':
+    # Return non-zero if the task failed, but silence failures due to expiration:
+    if status == 'pass' or status == 'expired':
```
yalcinmelihyasin, line 135: What do we do now, when we have internal failure in task_result[]?

hevrard: Good point!

This field did not seem to be present in the JSON returned by the (updated) LUCI tools. I asked the LUCI team, see the discussion in b/159313821, but basically such fields are not present unless there is actually an internal failure (or a regular failure, etc).

I've updated this PR to properly check for field existence before their value, and the check for internal_failure is back.

PR #369, opened 2020-07-07
Fix wrong usage of defer.

lpy:
Bug: 148970509
Test: bazel run gapit -- device --os android


lpy:
I believe I mis-understood the scope of `defer`. Please feel free to merge this patch once it's LGTMed before rolling out a dev release with this commit https://github.com/google/agi/commit/41f745f95ce0d74751481987da1310098417003b

PR #373, opened 2020-07-08
Go back to using a 1.8 JDK/JRE.

pmuetschard:
SWT requires 1.8 and has issues with 11. The original assumption that 1.8 doesn't notarize is invalid, it does indeed notarize.


kokoro/macos/package.sh
```diff
@@ -57,8 +57,9 @@ done
 iconutil -c icns -o AGI.app/Contents/Resources/AGI.icns AGI.iconset
 
 # Move the JRE's legal notices to the Resources folder, so signing doesn't complain.
-mkdir AGI.app/Contents/Resources/jre
-mv AGI.app/Contents/MacOS/jre/legal AGI.app/Contents/Resources/jre
+# Commented out, as JRE 1.8 does not have such a folder, JRE 11, does.
```
pau-baiget, line 60: Why don't we just remove it instead of commenting it?

PR #374, opened 2020-07-08
Move the hover card position to the left of the hover point.

lpy:
Previously when hover point moves towards to the right boundary of the window,
the hover card will stick on the right edge of the window which blocks the view
of the current hover point and hence the selection. This patch modifies the
position of the hover card and moves the position of the hover card to the left
side of the hover point.

Bug: b/158506210

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/BatterySummaryPanel.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/BatterySummaryPanel.java
@@ -135,7 +135,10 @@ public class BatterySummaryPanel extends TrackPanel<BatterySummaryPanel> impleme
       // Draw hovered card.
       if (hovered != null) {
         double cardW = hovered.allSize.w + 3 * HOVER_PADDING + LEGEND_SIZE;
-        double cardX = Math.min(mouseXpos + HOVER_MARGIN, w - cardW);
+        double cardX = mouseXpos + CURSOR_SIZE / 2 + HOVER_MARGIN;
+        if (cardX >= w - cardW) {
+          cardX = mouseXpos - CURSOR_SIZE / 2 - HOVER_MARGIN - cardW;
+        }
         ctx.setBackgroundColor(colors().hoverBackground);
         ctx.fillRect(cardX, mouseYpos, cardW, hovered.allSize.h);
 
@@ -187,7 +190,18 @@ public class BatterySummaryPanel extends TrackPanel<BatterySummaryPanel> impleme
       @Override
       public Area getRedraw() {
         double redrawW = CURSOR_SIZE + HOVER_MARGIN + hovered.allSize.w + 3 * HOVER_PADDING + LEGEND_SIZE;
-        double redrawX = Math.min(mouseXpos - CURSOR_SIZE, state.getWidth() - redrawW);
+        double redrawX = mouseXpos - CURSOR_SIZE / 2;
+        if (redrawX >= state.getWidth() - redrawW) {
+          redrawX = mouseXpos + CURSOR_SIZE / 2.0 - redrawW;
+
+          // If the hover card is drawn on the left side of the hover point, when moving the mouse
+          // from left to right, the right edge of the cursor doesn't seem to get redrawn all the
+          // time, this looks like a precision issue. This also happens when cursor is now on the
+          // right side of the hover card, and the mouse moving from right to left there seems to
+          // be a precision issue on the right edge of the cursor, hence extend the redraw with by
+          // plusing the radius of the cursor.
+          redrawW += CURSOR_SIZE / 2;
+        }
         return new Area(redrawX, -TRACK_MARGIN, redrawW, HEIGHT + 2 * TRACK_MARGIN);
       }
 
--- a/gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
@@ -138,22 +138,28 @@ public class CounterPanel extends TrackPanel<CounterPanel> implements Selectable
 
         ctx.setBackgroundColor(colors().hoverBackground);
         double bgH = Math.max(hovered.size.h, trackHeight);
-        double x = Math.min(hovered.getTooltipX(), w - (2 * HOVER_PADDING + hovered.size.w));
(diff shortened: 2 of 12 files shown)


pmuetschard:
@stellama0208 can you please review this?

gapic/src/main/com/google/gapid/perfetto/views/CpuPanel.java
```diff
@@ -138,7 +139,12 @@ private void renderSummary(RenderContext ctx, CpuTrack.Data data, double w, doub
       double mouseX = state.timeToPx(tStart + hovered.bucket * data.bucketSize + data.bucketSize / 2);
       double dx = HOVER_PADDING + hovered.size.w + HOVER_PADDING;
       double dy = HOVER_PADDING + hovered.size.h + HOVER_PADDING;
-      double cardX = Math.min(mouseX + HOVER_MARGIN, w - dx);
+      // By default, the hover card left boundary x-axis coordinate is the current hover point plus
+      // CURSOR_SIZE and HOVER_MARGIN.
+      double cardX = mouseX + CURSOR_SIZE + HOVER_MARGIN;
```
stellama0208, line 144: Sometimes we have the cursor size counted in but sometimes we don't, if we can be more consistent it would be great.

gapic/src/main/com/google/gapid/perfetto/views/BatterySummaryPanel.java
```diff
@@ -187,7 +190,12 @@ protected Hover onTrackMouseMove(TextMeasurer m, double x, double y, int mods) {
       @Override
       public Area getRedraw() {
         double redrawW = CURSOR_SIZE + HOVER_MARGIN + hovered.allSize.w + 3 * HOVER_PADDING + LEGEND_SIZE;
-        double redrawX = Math.min(mouseXpos - CURSOR_SIZE, state.getWidth() - redrawW);
+        double redrawX = mouseXpos - CURSOR_SIZE;
+        if (redrawX >= state.getWidth() - redrawW) {
+          redrawX = mouseXpos - redrawW;
+          // The cursor is now on the right side of the hover card, hence extend the redraw width.
+          redrawW += CURSOR_SIZE;
```
stellama0208, line 197: This panel doesn't have the logic of doubling up the redrawW, but somehow it still works. 
I was a little bit curious so I looked into it and found the base Canvas `PanelCanvas` will handle combining together the old and new redraw areas. In this way, we could remove the logic of doubling up the redrawW in other panels.

gapic/src/main/com/google/gapid/perfetto/views/FrameEventsPanel.java
```diff
@@ -157,7 +157,11 @@ public void renderSlices(RenderContext ctx, FrameEventsTrack.Data data, double w
     }
 
     if (hoveredTitle != null) {
-      double cardX = Math.min(mouseXpos + HOVER_MARGIN, w - (hoveredSize.w + 2 * HOVER_PADDING));
+      // double cardX = Math.min(mouseXpos + HOVER_MARGIN, w - (hoveredSize.w + 2 * HOVER_PADDING));
```
stellama0208, line 160: I suppose this line is meant to be deleted.

gapic/src/main/com/google/gapid/perfetto/views/CpuFrequencyPanel.java
```diff
@@ -31,6 +31,8 @@
  */
 public class CpuFrequencyPanel extends TrackPanel<CpuFrequencyPanel> {
   private static final double HEIGHT = 30;
+  private static final double CURSOR_SIZE = 5;
+  private static final double HOVER_PADDING = 8;
```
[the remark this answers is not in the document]
lpy: This panel doesn't have margin. I added one

gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
```diff
@@ -138,22 +139,28 @@ protected void renderTrack(RenderContext ctx, Repainter repainter, double w, dou
...
+          hx = hovered.mouseX - CURSOR_SIZE - HOVER_MARGIN - 2 * HOVER_PADDING - hovered.size.w;
+        }
+        ctx.fillRect(hx, Math.min((trackHeight - bgH) / 2, 0),
             2 * HOVER_PADDING + hovered.size.w, bgH);
         ctx.setForegroundColor(colors().textMain);
-        x += HOVER_PADDING;
+        // The left x-axis coordinate of the left labels.
+        double x = hx + HOVER_PADDING;
         y = (trackHeight - hovered.size.h) / 2;
+        // The difference between the x-axis coordinate of the left labels and the right labels.
```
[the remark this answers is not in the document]
lpy: The margin of the further side of the hover point doesn't matter in all panels, hence we always use one `HOVER_MARGIN`, which is the one between the hover point and the hover card.

The CounterPanel HoverCard technically has 5 `HOVER_PADDING`. The order of them from left to right is: `HOVER_PADDING` -> `LEFT_LABEL_NAME:` -> `HOVER_PADDING` -> `VALUE` -> `HOVER_PADDING` -> `RIGHT_LABEL_NAME:` -> `HOVER_PADDING` -> `VALUE` -> `HOVER_PADDING`. So you are right about HoverCard has 3 `HOVER_PADDING`.

None of the hover card in this patch has included the `HOVER_PADDING` of the left and right boundaries inside hover card.

lpy:
Thanks, ptal

gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
```diff
@@ -191,12 +197,48 @@ protected Hover onTrackMouseMove(Fonts.TextMeasurer m, double x, double y, int m
...
+          // with the right boundary of the default redrawn area when the hover card is drawn on
+          // the left side of the hover point.
+          redrawRx = Math.max(hovered.mouseX + CURSOR_SIZE / 2, endX);
+
+          // Finally, re-calculate the redrawn width, plus radius of the cursor to avoid the
+          // precision issue at the right edge of the cursor.
+          redrawW = redrawRx - redrawLx + CURSOR_SIZE / 2;
+        }
+
+        return new Area(redrawLx, -TRACK_MARGIN, redrawW, trackHeight + 2 * TRACK_MARGIN);
```
stellama0208, line 241: nit: If you'd like to, you can make this part conciser.

PR #375, opened 2020-07-08
Update NDK to r21d

hevrard:


hevrard:
The intended change is the last commit, it is currently based on top of #331 which should be merged before this PR can be properly rebased on master.

hevrard:
Rebased, it is now a single short commit.

PR #376, opened 2020-07-08
Clean up GPU profiling integration with BSP drivers.

lpy:
Instead of trying to provide an option to profile against system driver, AGI
should always query and use the latest available drivers on the device. When
developer driver is present, AGI will always try to set up and use it, and when
fail to do so, fall back to use GPU profiling libraries in system image. If
developer driver is not installed or the device doesn't support developer
driver, always attempt to use GPU profiling libraries in system image.

Bug: b/148970509
Test: Perform below commands on supported devices.
Test: bazel run gapit -- device --os android

### DIFF
--- a/gapidapk/deviceinfo.go
+++ b/gapidapk/deviceinfo.go
@@ -104,29 +104,51 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
 	// Close any previous runs of the apk
 	apk.Stop(ctx)
 
-	driver, err := d.GraphicsDriver(ctx)
-	if err != nil {
-		return err
-	}
-
 	var cleanup app.Cleanup
 
-	// Set up device info service to use prerelease driver.
-	nextCleanup, err := adb.SetupPrereleaseDriver(ctx, d, apk.InstalledPackage)
-	cleanup = cleanup.Then(nextCleanup)
+	packages := []string{}
+	driver, err := d.GraphicsDriver(ctx)
 	if err != nil {
-		cleanup.Invoke(ctx)
-		return err
+		// If there's an error, keep going to attempt to use GPU profiling
+		// libraries in system image.
+		log.W(ctx, "Failed to query developer driver: %v, assuming no developer driver found.", err)
+	}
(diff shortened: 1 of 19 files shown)


lpy:
This patch is partially a revert of of #326, ptal!

gapidapk/deviceinfo.go
```diff
@@ -104,30 +104,52 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
...
+	driver, err := d.GraphicsDriver(ctx)
 	if err != nil {
-		cleanup.Invoke(ctx)
-		return err
+		// If there's an error, keep going to attempt to use GPU profiling
+		// libraries in system image.
+		log.W(ctx, "Failed to query developer driver: %v, assuming no developer driver found.", err)
+	}
+	if driver.Package != "" {
+		log.I(ctx, "Use GPU profiling libraries from developer driver package.")
```
pmuetschard, line 117: nit: "Using..."

And, since we're already logging, we should log the package name:
```suggestion
		log.I(ctx, "Using GPU profiling libraries from developer driver package: %v", driver.Package)
```

(same below)

gapidapk/deviceinfo.go
```diff
@@ -104,30 +104,52 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
...
+		// Set up device info service to use prerelease driver.
+		nextCleanup, err := adb.SetupPrereleaseDriver(ctx, d, apk.InstalledPackage)
+		cleanup = cleanup.Then(nextCleanup)
+		if err != nil {
+			cleanup.Invoke(ctx)
+			return err
+		}
+		packages = append(packages, driver.Package)
+	} else {
+		log.I(ctx, "No developer driver found, attempt to use GPU profiling libraries in system image.")
```
pmuetschard, line 128: nit
```suggestion
		log.I(ctx, "No developer driver found, attempting to use GPU profiling libraries in system image.")
```

(same below)

lpy:
Thanks, nits addressed :)

PR #378, opened 2020-07-10
Limit geometry view to only draw calls

hliatis:
Bug: b/150769254

### DIFF
--- a/gapis/api/vulkan/draw_call_mesh.go
+++ b/gapis/api/vulkan/draw_call_mesh.go
@@ -22,7 +22,9 @@ import (
 	"github.com/google/gapid/core/stream"
 	"github.com/google/gapid/core/stream/fmts"
 	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/messages"
 	"github.com/google/gapid/gapis/resolve"
+	"github.com/google/gapid/gapis/service"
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/draw_call_mesh.go
```diff
@@ -42,6 +44,15 @@ func drawCallMesh(ctx context.Context, dc *VkQueueSubmit, p *path.Mesh, r *path.
...
 	c := getStateObject(s)
 
+	cmd, err := resolve.Cmd(ctx, cmdPath, r)
+	if err != nil {
+		return nil, err
+	}
+
+	if !cmd.CmdFlags().IsExecutedDraw() || len(cmdPath.Indices) == 1 {
+		return nil, &service.ErrDataUnavailable{Reason: messages.ErrMessage("Please select a draw call")}
+	}
```
pmuetschard, line 54: This is the wrong error (there is an existing "not a draw call" error) to return and probably the wrong place as well. We should not even get here for anything other than a draw call.
There is an issue with how meshes were implemented in Vulkan compared to GLES, which is now hard to see, as the GLES code is gone. There is a loop in the resolve that tries each node in the requested range that expects certain behavior of things below it. These expectations are currently not met, which is why we are seeing issues.

pmuetschard:
superseded by #390

PR #380, opened 2020-07-14
Surface error from trace processor

rschiu:
Bug: b/148797225

### DIFF
--- a/gapic/src/main/com/google/gapid/views/StatusBar.java
+++ b/gapic/src/main/com/google/gapid/views/StatusBar.java
@@ -44,9 +44,11 @@ import org.eclipse.swt.widgets.Link;
  */
 public class StatusBar extends Composite {
   private final Composite memoryStatus;
+  private final Composite perfettoStatus;
   private final Composite replayStatus;
   private final Composite serverStatus;
   private final HeapStatus heap;
+  private final Label perfettoStatusText;
   private final Label serverPrefix;
   private final Label server;
   private final Label replay;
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/views/StatusBar.java
```diff
@@ -101,6 +113,19 @@ public void setNotification(String text, Runnable onClick) {
...
   }
 
+  public void setPerfettoStatus(String errorMsg) {
+    System.out.println("setPerfettoStatus");
+    System.out.println("errorMsg: " + errorMsg);
+    if (errorMsg == null) {
+      perfettoStatus.setVisible(false);
+    } else {
+      perfettoStatus.setVisible(true);
+      perfettoStatusText.setToolTipText(errorMsg);
```
rschiu, line 123: @pmuetschard @stellama0208 Could you help me out a bit here?
I am working on b/148797225 and the idea I have is to show a warning symbol in the status bar if there is any error while processing the trace.  When the user hover over the warning symbol, the tooltip will describe the problem.  However, setToolTipText doesn't seem to work.  Nothing shows up when hovering over the symbol.

Would you know why that is the case?  Or do you have alternative suggestions on how I should approach this?

pmuetschard:
Tooltip seems to work for me:
image

I just added the following to the `StatusBar` constructor (and adjusted the grid layout size):
```
Label test = createLabel(this, "", theme.error());
test.setToolTipText("Hello World!");
```

gapic/src/main/com/google/gapid/views/StatusBar.java
stellama0208: Is that because you are using `perfettoStatusText.setToolTipText(errorMsg);` rather than `perfettoStatus.setToolTipText(errorMsg);`, it seems that now the tooltips is on the text and not the symbol?

PR #384, opened 2020-07-15
Give informative message when there's no available geometry.

stellama0208:
- Fix a server exception hiding problem inside geometry view.
   (We have better server error visualization inside gapic now.)
 - Fix an early return logic, to allow showing original geometry
   mesh, when there's only original but no faceted geometry mesh.
 - Bug: b/159016023.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Geometries.java
+++ b/gapic/src/main/com/google/gapid/models/Geometries.java
@@ -141,8 +141,7 @@ public class Geometries
       return success(result.get());
     } catch (DataUnavailableException e) {
       DeviceDependentModel.Source<Source> s = getSource();
-      // TODO: don't assume that it's because of not selecting a draw call.
-      return success(new Data(s.device, s.source.semantics, null, null));
+      return error(Loadable.Message.error(e));
     } catch (RpcException e) {
       LOG.log(WARNING, "Failed to load the geometry", e);
       return error(Loadable.Message.error(e));
--- a/gapic/src/main/com/google/gapid/views/GeometryView.java
+++ b/gapic/src/main/com/google/gapid/views/GeometryView.java
@@ -292,9 +292,8 @@ public class GeometryView extends Composite
     }
 
     Geometries.Data meshes = models.geos.getData();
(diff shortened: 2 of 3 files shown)


gapic/src/main/com/google/gapid/models/Geometries.java
```diff
@@ -141,8 +141,7 @@ protected boolean isSourceComplete(DeviceDependentModel.Source<Source> source) {
       return success(result.get());
     } catch (DataUnavailableException e) {
       DeviceDependentModel.Source<Source> s = getSource();
-      // TODO: don't assume that it's because of not selecting a draw call.
-      return success(new Data(s.device, s.source.semantics, null, null));
+      return error(Loadable.Message.error(e));
```
pmuetschard, line 144: `DataUnavailableException` shouldn't be treated as an error. We treat it differently in the UI everywhere else. It should not show the exclamation point. Also, is the message from the server here actually helpful?

gapic/src/main/com/google/gapid/views/GeometryView.java
```diff
@@ -292,9 +292,8 @@ public void onGeometryLoaded(Message error) {
     }
 
     Geometries.Data meshes = models.geos.getData();
-    if (!meshes.hasFaceted()) {
-      loading.showMessage(Info, Messages.SELECT_DRAW_CALL);
-      // ?? saveItem.setEnabled(false);
+    if (!meshes.hasOriginal() && !meshes.hasFaceted()) {
+      loading.showMessage(Info, Messages.NO_MESH_DATA);
```
pmuetschard, line 296: Why the rewording of the message? It was more specific/helpful before.

pmuetschard:
Just a quick status update: talked f2f about this and we realized that this PR doesn't actually fix the bug of geometry not showing up when it should / showing up when it shouldn't. So, there is more work to be done on the server, which will clean up the error communication as well.

stellama0208:
Thanks for appending the summary and take a look into Geometry! Will close this PR for now.

PR #385, opened 2020-07-15
Provide restart server option in error dialog when gapis lost.

stellama0208:
- Bug: b/159788419.

### DIFF
--- a/gapic/src/main/com/google/gapid/Main.java
+++ b/gapic/src/main/com/google/gapid/Main.java
@@ -187,9 +187,9 @@ public class Main {
     @Override
     public void onServerExit(int code, String panic) {
       scheduleIfStillOpen(shell -> showErrorDialogWithTwoButtons(
-          shell, getAnalytics(), "The gapis server has exited with an error code of: " + code, panic,
+          shell, getAnalytics(), String.format(Messages.SERVER_ERROR_MESSAGE, code), panic,
           IDialogConstants.RETRY_ID, "Restart Server", this::restartServer,
-          IDialogConstants.CLOSE_ID, IDialogConstants.CLOSE_LABEL, window::close)
+          IDialogConstants.CLOSE_ID, "Exit", window::close)
       );
     }
 
--- a/gapic/src/main/com/google/gapid/Main.java
+++ b/gapic/src/main/com/google/gapid/Main.java
@@ -172,9 +172,6 @@ public class Main {
         shell.getDisplay().asyncExec(() -> showFirstTimeDialog(shell, models, widgets, onStart));
       }
 
-      // TODO: Only for easy gapis crash testing, will delete before check in.
-      window.statusBar.addListener(SWT.MouseDown, $ -> server.disconnect());
-
       // Add the links on Loading Screen after the server set up.
       window.updateLoadingScreen(client, models, widgets);
     }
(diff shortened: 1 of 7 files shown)


gapic/src/main/com/google/gapid/Main.java
```diff
@@ -165,24 +168,44 @@ private void uiStartup(Shell shell) {
         shell.getDisplay().asyncExec(() -> showFirstTimeDialog(shell, models, widgets, onStart));
       }
 
+      // TODO: Only for easy gapis crash testing, will delete before check in.
+      window.statusBar.addListener(SWT.MouseDown, $ -> server.disconnect());
+
```
stellama0208, line 173: For easy testing, a click on the status message composite can trigger a quick gapis losing failure :) This part will be deleted before check in.

gapic/src/main/com/google/gapid/MainWindow.java
```diff
@@ -267,6 +270,62 @@ private Point getDefaultInitialLocation(Point size) {
...
+        }
+      }
+    });
+
+    if (OS.isMac) {
+      MacApplication.init(shell.getDisplay(),
+          () -> showAbout(shell, models.analytics, widgets),
+          () -> showSettingsDialog(shell, models, widgets.theme),
+          file -> models.capture.loadCapture(new File(file)));
+    }
```
pmuetschard, line 314: This doesn't quite work correctly. `MacApplication.init` should only be called once. It adds listeners to the menu each time it is called.

pmuetschard:
We talked f2f to explore an alternative where we do not need to allocate new instances of the `Client`, `Models` and `Widgets` classes, simplifying the code a bit.

stellama0208:
Thanks for the f2f comments, they provide a great help to make this PR better! Those comments are addressed with new commits, ready for another look!

gapic/src/main/com/google/gapid/server/Client.java
```diff
@@ -64,12 +64,18 @@
...
-  private final GapidClient client;
+  private GapidClient client;
+
+  public Client() { /* Initialize an empty wrapper. */}
 
   public Client(GapidClient client) {
     this.client = client;
   }
 
+  public void initializeGapidClient(GapidClient client) {
```
pmuetschard, line 75: Just call it `setGapidClient`

gapic/src/main/com/google/gapid/util/Messages.java
```diff
@@ -75,6 +75,10 @@
   public static final String SETTINGS_TITLE = "Modify Settings";
   public static final String ERROR_MESSAGE =
       "The application encountered an error:\n%s\n\nPlease check the logs for details.";
+  public static final String SERVER_ERROR_MESSAGE =
+      "\nThe server has exited with an error code of: %s \n" +
+      "Most functions in AGI are unavailable without an server. \n" +
+      "You can restart server, exit, or close the dialog to save your unfinished work.";
```
pmuetschard, line 81:
```suggestion
  public static final String SERVER_ERROR_MESSAGE =
      "\nThe server has exited with an error code of: %s \n" +
      "Most functions in AGI are unavailable without a server. \n" +
      "You can restart the server, exit AGI, or close the dialog to continue without a server.";
```

PR #388, opened 2020-07-15
Better horizontal scrolling in the system profiler.

pmuetschard:
When the scrolling is done via the touch pad or mouse wheel, handle it ourselves, rather than going via the scrollbar. The scrollbar size is limited, and thus when zoomed in, the scroll accuracy is limited and leads to scrolling too fast/far. We can scroll better by handling it directly if the event comes from the touch pad or mouse wheel.

Bug: http://b/160731902

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/TraceComposite.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/TraceComposite.java
@@ -21,6 +21,8 @@ import static com.google.gapid.perfetto.views.StyleConstants.KB_PAN_FAST;
 import static com.google.gapid.perfetto.views.StyleConstants.KB_PAN_SLOW;
 import static com.google.gapid.perfetto.views.StyleConstants.KB_ZOOM_FAST;
 import static com.google.gapid.perfetto.views.StyleConstants.KB_ZOOM_SLOW;
+import static com.google.gapid.perfetto.views.StyleConstants.TP_PAN_FAST;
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/perfetto/views/TraceComposite.java
```diff
@@ -82,9 +84,18 @@ public TraceComposite(Composite parent, Analytics analytics, Theme theme) {
...
-      if ((e.stateMask & SWT.MODIFIER_MASK) == SWT.MOD1) {
-        // Ignore horizontal scroll, only when zooming.
+      int mods = e.stateMask & SWT.MODIFIER_MASK;
+      if (mods == SWT.MOD1) {
+        // Ignore horizontal scroll, when zooming.
+        e.doit = false;
+      } else {
+        // Treat horizontal touch pad/scroll wheel ourselves, rather than going through the
+        // scrollbar, since the scrollbar's size is limited.
         e.doit = false;
```
hevrard, line 94: It looks like there is `e.doit = false` in both branches, could this be refactored as:

```
e.doit = false;
if (mods != SWT.MOD1) {
    if (state.dragX(...)) {
        ...
    }
}
```

pmuetschard: Done. I saw that, but had decided to not write it that way as I thought it was confusing. I think I found an ok way to write it now that explains properly why things are the way they are.

PR #390, opened 2020-07-16
Geometry View Fixes

pmuetschard:
- Fixes how we communicate in Vulkan whether a mesh is available at a requested command/sub-command.
- Does some refactoring to make it more obvious how to communicate mesh un-availability
- Verifies that we are on a draw call inside a queue submit (this is inspired from PR #378)
- Fixes the UI to show a proper error if the mesh is empty
- Resets the UI's status bar so that it doesn't keep showing the stats from the last successfully loaded model

Bug: http://b/159016023, http://b/150769254

### DIFF
--- a/gapis/resolve/mesh.go
+++ b/gapis/resolve/mesh.go
@@ -30,15 +30,10 @@ func Mesh(ctx context.Context, p *path.Mesh, r *path.ResolveConfig) (*api.Mesh, 
 	if err != nil {
 		return nil, err
 	}
-	mesh, err := meshFor(ctx, obj, p, r)
-	switch {
-	case err != nil:
-		return nil, err
-	case mesh != nil:
-		return mesh, nil
-	default:
-		return nil, &service.ErrDataUnavailable{Reason: messages.ErrMeshNotAvailable()}
+	if mesh, err := meshFor(ctx, obj, p, r); err != api.ErrMeshNotAvailable {
+		return mesh, err
 	}
+	return nil, &service.ErrDataUnavailable{Reason: messages.ErrMeshNotAvailable()}
 }
 
 func meshFor(ctx context.Context, o interface{}, p *path.Mesh, r *path.ResolveConfig) (*api.Mesh, error) {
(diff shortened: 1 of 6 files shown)


gapis/resolve/mesh.go
```diff
@@ -30,15 +30,10 @@ func Mesh(ctx context.Context, p *path.Mesh, r *path.ResolveConfig) (*api.Mesh,
...
-	mesh, err := meshFor(ctx, obj, p, r)
-	switch {
-	case err != nil:
-		return nil, err
-	case mesh != nil:
-		return mesh, nil
-	default:
-		return nil, &service.ErrDataUnavailable{Reason: messages.ErrMeshNotAvailable()}
+	if mesh, err := meshFor(ctx, obj, p, r); err != api.ErrMeshNotAvailable {
+		return mesh, err
```
stellama0208, line 34: It looks like if a non-ErrMeshNotAvailable error happen, the mesh will still get returned, is this expected?

gapis/resolve/mesh.go
```diff
@@ -60,29 +55,19 @@ func meshFor(ctx context.Context, o interface{}, p *path.Mesh, r *path.ResolveCo
...
-				p := o.Commands.Capture.Command(o.Commands.From[0], cmd...).Mesh(p.Options)
-				if mesh, err := meshFor(ctx, cmds[o.Commands.From[0]], p, r); mesh != nil || err != nil {
-					return mesh, err
-				}
+		cmd := append([]uint64{}, o.Commands.From...) // make a copy of o.Commands.From
+		for i := o.Commands.To[lastSubcommand]; i >= o.Commands.From[lastSubcommand]; i-- {
+			cmd[lastSubcommand] = i
+			p := o.Commands.Capture.Command(cmd[0], cmd[1:]...).Mesh(p.Options)
+			if mesh, err := meshFor(ctx, cmds[o.Commands.From[0]], p, r); err != api.ErrMeshNotAvailable {
+				return mesh, err
```
stellama0208, line 70: Same question here, are we doing it deliberately to let the call function deal with the non-ErrMeshNotAvailable error?

gapis/resolve/mesh.go
pmuetschard: Yes, because then the err is either nil and the mesh is good, or there was another kind of error, which should be surfaced.

gapis/resolve/mesh.go
pmuetschard: Yes, same answer.

PR #394, opened 2020-07-20
Do not request Framebuffer Attachments if resources are still loading

hliatis:
Prevents a null pointer exception. Other views have similar checks.

Bug: b/161699220

### DIFF
--- a/gapic/src/main/com/google/gapid/views/PipelineView.java
+++ b/gapic/src/main/com/google/gapid/views/PipelineView.java
@@ -170,8 +170,11 @@ public class PipelineView extends Composite
   }
 
   private void updatePipelines() {
-    if (models.resources.isLoaded() && models.commands.getSelectedCommands() != null) {
-      loading.startLoading();
+    loading.startLoading();
+
+    if (models.commands.getSelectedCommands() == null) {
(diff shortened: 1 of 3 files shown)


hliatis:
I agree. Other views such as the Pipeline View and Texture View use that message as well.

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -170,7 +170,9 @@ public void onCommandsSelected(CommandIndex path) {
   }
 
   private void updatePipelines() {
-    if (models.resources.isLoaded() && models.commands.getSelectedCommands() != null) {
+    if (!models.resources.isLoaded()) {
+      loading.startLoading();
+    } else if (models.commands.getSelectedCommands() != null) {
       loading.startLoading();
```
hevrard, line 176: Here `loading.startLoading()` is on both branches, and that may makes sense to make the logic clearer, but do you think this could be refactored?

stellama0208: The flow in FramebufferView.java looks great, show hinting message when `models.commands.getSelectedCommands() == null`, and if all check passes, the else clause lead the following main part.

PR #400, opened 2020-07-21
Improve replay device selection dialog

hevrard:
This is still a WIP.

Design doc: go/agi-replay-device

Replace the combo selector by a pair of tables to show compatible and
incompatible devices. The replay device can be selected by cliking on
a row in the compatible table.

Major things to clarify:

- refreshing the tables currently requires two RPC calls, we should
  probably refactor to get all info in one RPC -- not specifically for
  performance, but also to simplify the UI code here.

- Device validation: we currently cannot disambiguate between "we
  didn't try to validate this device yet" and "we tried in this
  session and it failed", it may be nice to have this info to show as
  a column in the table. More generally, we probably want to move
  runValidationCheck() to the Devices.java model. It might be
  interesting to try and validate all compatible devices in parallel,
  but then we need to not use the SingleInFlight rpc controller in
  Devices.


hevrard:
@pmuetschard @stellama0208 @hliatis this is a draft w.r.t. replay device selection "v2", could one of you have a look and tell me what you think of:

- merging gapis RPCs `GetDevices()` and `GetDevicesForReplay()` into a single RPC?

- try and validate all compatible devices in parallel?

See details in the comment above and in the code draft. Thanks!

hevrard:
As discussed

> * merging gapis RPCs `GetDevices()` and `GetDevicesForReplay()` into a single RPC?

Keep two RPCs, augment the proto message returned to indicate device compatibility, and the reason for not being compatible.

> * try and validate all compatible devices in parallel?

No, let have the user click on a device to check validation status.

hevrard:
@pmuetschard @stellama0208 @hliatis This is now ready for review. I've squashed all in a single commit with this message, which can help reviewing:

```
    Improve replay device selection dialog
    
    Improve replay device selection dialog by diplaying all known devices
    in two tables: compatible devices which can be used for replay, and
    incompatible devices, alongside the incompatibility reason to explain
    for each why it is not fit to replay the trace.
    
    This change introduces a new "Incompatibility" protobuf enum, and
    augment the Devices proto message with a list of IncompatibleDevice,
    such that when GAPIS is queried for the list of devices available to
    replay a given trace, it also returns the list of incompatible
    devices, along with the incompatility reason.
    
    On the GAPIS side, the incompatibility reason is set in
    gapis/api/vulkan.GetReplayPriority(), and propagated to the answer to
    a GetDevicesForReplay() RPC response.
    
    On the GAPIC side, models.Devices.loadReplayDevices() handles the
    augmented RPC reply to store both compatible and incompatible
    devices. Each set of devices is presented in a table in
    views.DeviceDialog. Selecting a compatible replay device triggers a
    validation check, and if validation pass then the user can open the
    trace with the selected device.
```

The replay device selection dialog looks like this when there is no compatible device plugged in:

nodevice

And then if you plug a compatible device, it shows up and is selectable:

okdevice

Selecting a compatible device enabled the "OK" button which opens the trace.

hevrard:
ptal

core/os/device/device.proto
```diff
@@ -56,6 +56,21 @@ enum OSKind {
   Stadia = 5;
 }
 
+// Incompatibility list possible reasons why a device is not compatible to
+// replay a given trace.
+enum Incompatibility {
+  Compatible = 0;  // Default, for compatible devices
+  ABIOS = 1;
+  ABIArchitecture = 2;
+  ABIMemoryLayout = 3;
```
pmuetschard, line 65: Remove the `ABI` part of these enums.

core/os/device/device.proto
```diff
@@ -56,6 +56,21 @@ enum OSKind {
...
+  GPUVendorID = 4;
+  GPUDeviceID = 5;
```
pmuetschard, line 67: Combine into one `GPU`.

core/os/device/device.proto
```diff
@@ -56,6 +56,21 @@ enum OSKind {
...
+  VkDriverVersion = 6;
+  VkAPIVersion = 7;
```
pmuetschard, line 69: Remove the 'Vk' part. Where we can, let's keep the gapis API graphics API agnostic.

core/os/device/device.proto
```diff
@@ -56,6 +56,21 @@ enum OSKind {
...
+  NoTraceVkDriver = 8;
+  NoVkSupport = 9;
```
pmuetschard, line 71: Combine into `NoApiSupport`. In the UI, we can show it as "This device doesn't support the captured API." If we used the `stringtable.Msg` approach, we could have the server stuff "Vulkan" into the message.

core/os/device/device.proto
```diff
@@ -56,6 +56,21 @@ enum OSKind {
   Stadia = 5;
 }
 
+// Incompatibility list possible reasons why a device is not compatible to
+// replay a given trace.
+enum Incompatibility {
```
pmuetschard, line 61: I'm not a big fan of the double-negative. I'd prefer calling this `ReplayCompatability` and possibly prefixing (all but the first) enums with a negative (i.e. kind of like `VkResult`). However, I see why'd you choose this, so, up to you.

gapis/service/service.proto
```diff
@@ -50,6 +50,7 @@ message Commands {
 }
 message Devices {
   repeated path.Device list = 1;
+  repeated IncompatibleDevice incompatibleList = 2;
```
pmuetschard, line 53: I'd prefer to just have a `repated device.Incompatability` field here, where the two lists are the same length and correspond to each other. This is a common proto practice. Otherwise just a repeated message with both fields. I.e. one of these:
```
message Devices {
  repeated path.Device list = 1;
  repeated device.Incompataiblity replayCompatability = 2;
}
```
or
```
message Device {
  path.Device device = 1;
  device.Incompataiblity replayCompatability = 2;
}

message Devices {
  repeated Device list = 1;
}
```

 Also please document here that this second field is only set for the `GetDevicesForReplay` call, and not `Get(Path.Devices)`.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -324,6 +358,33 @@ protected void setValidationStatus(DeviceValidationResult result) {
...
+          return "Vulkan API version";
+        case Device.Incompatibility.NoTraceVkDriver_VALUE:
+          return "Capture has no Vulkan driver info";
+        case Device.Incompatibility.NoVkSupport_VALUE:
+          return "Device has no Vulkan support";
+        default:
+          return "Unknown incompatibility";
+      }
+    }
+
```
pmuetschard, line 387: I'd prefer we use the `stringtable.Msg` mechanism here to get more details from the server. For example, instead of just saying "ABI architecture", which users will not understand what it means, we can say "Device doesn't support the arm64-v8a ABI", etc. The reality is that this dialog will most often be shown when someone opens a trace that can no longer replay on the one device they have connected (i.e. not to select from multiple devices), so we need to be really clear why we are refusing to open a user's trace.

As it is, we are just translating enums to strings, which is what the stringtable is supposed to do for us.

core/os/device/device.proto
hevrard: We can't, it leads to errors like:

```
core/os/device/device.proto:63:3: "OS" is already defined in "device".
core/os/device/device.proto:63:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "OS" must be unique within "device", not just within "Incompatibility".
core/os/device/device.proto:64:3: "Architecture" is already defined in "device".
core/os/device/device.proto:64:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "Architecture" must be unique within "device", not just within "Incompatibility".
core/os/device/device.proto:65:3: "MemoryLayout" is already defined in "device".
core/os/device/device.proto:65:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "MemoryLayout" must be unique within "device", not just within "Incompatibility".
```

That's why I added the "ABI" prefix.

But I think that by renaming this enum to `ReplayCompatibility` as you suggest, we may use enum names like `BadOS` or `InvalidOS`.

pmuetschard: 🙄  I see. I'd prefer something like `Incompatable<foo>` or `<foo>Mismatch`. The OS is neither bad nor invalid. :)

hevrard: :+1:

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -173,54 +173,89 @@ protected Control createDialogArea(Composite parent) {
...
+      incompatibleDeviceTable.setLabelProvider(new LabelProvider() {
+        @Override
+        public String getText(Object element) {
+          return Devices.getLabel(((IncompatibleDeviceInfo) element).device);
+        }
+      });
+      Widgets.createTableColumn(incompatibleDeviceTable, "Name", dev -> ((IncompatibleDeviceInfo)dev).device.getName());
+      Widgets.createTableColumn(incompatibleDeviceTable, "Serial", dev -> ((IncompatibleDeviceInfo)dev).device.getSerial());
+      Widgets.createTableColumn(incompatibleDeviceTable, "GPU", dev -> ((IncompatibleDeviceInfo)dev).device.getConfiguration().getHardware().getGPU().getName());
+      Widgets.createTableColumn(incompatibleDeviceTable, "Driver version", dev -> Integer.toUnsignedString(((IncompatibleDeviceInfo)dev).device.getConfiguration().getDrivers().getVulkan().getPhysicalDevices(0).getDriverVersion()));
```
stellama0208, line 245: I think we should do a size check before `getPhysicalDevices(0)`, met out of boundary exception here sometimes.

gapis/api/vulkan/replay.go
```diff
@@ -49,59 +49,77 @@ var (
...
 	}
 
 	// The device does not support Vulkan
 	if devVkDriver == nil {
-		return 0
+		return 0, device.Incompatibility_NoVkSupport
 	}
 
+	var incompatibility device.Incompatibility
 	for _, abi := range devAbis {
```
stellama0208, line 71: Since here we iterate through all the ABIs, the returned incompatibility reason might only record the reason found in the last ABI. it might be a nit improvement, but maybe we could rank the incompatibility reason and return the most detailed one throughout the iteration?

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -173,54 +173,89 @@ protected Control createDialogArea(Composite parent) {
...
+
+      // Refresh button
+      refreshDeviceButton = Widgets.createButton(composite, Messages.SELECT_DEVICE_REFRESH_TABLE,
+          e -> {
+            refreshDeviceButton.setText(Messages.SELECT_DEVICE_TABLE_REFRESHING);
+            refreshDeviceButton.setEnabled(false);
+            logFailure(LOG,
+                Scheduler.EXECUTOR.schedule(
+                    () -> models.devices.loadReplayDevices(models.capture.getData().path),
+                    300, TimeUnit.MILLISECONDS));
```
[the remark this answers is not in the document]
hevrard: The reason is to give some feedback to the user that something is happening: with a delay of zero, if the list of replay device does not change, the user might think pressing the button does not do anything. The small delay makes sure the button changes to "refreshing devices" for a split-second. I added a comment to clarify.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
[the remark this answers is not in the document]
hevrard: Moved to a separate function.

hevrard: Done, in the new function Devices.GetDriverVersion()

core/os/device/device.proto
hevrard: done

core/os/device/device.proto
hevrard: done

core/os/device/device.proto
hevrard: Renamed to ReplayCompatibility

core/os/device/device.proto
hevrard: done

gapis/service/service.proto
hevrard: I changed to the first of the options you suggested:

```
message Devices {
  repeated path.Device list = 1;
  repeated device.Incompatibility replayCompatibilityList = 2;
}
```

Note that this impacts the semantics of GetDevicesForReplayResponse, where the device list now contains both compatible and incompatible devices.  They are sorted to make sure compatible devices always come first.

core/os/device/device.proto
hevrard: Renamed.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
hevrard: I've laid down the basis to use stringtable, but I'm not sure I'm doing it right -- it seems to slow down the apparition of the pop-up.

hevrard: The incompatibility reason messages are now properly generated on the server side.

gapis/api/vulkan/replay.go
hevrard: Good point! I added a "reasonDepth" check such that we retain the incompatibility reason with the deepest depth, i.e. the most detailed one. I hope this covers your remark.

hevrard:
The incompatibility is now properly explained:

screenshot

Maybe we could even drop the "GPU" and "Driver version" columns? (I'm keen on keeping "Serial" as it properly disambiguates devices of the same model).

hevrard:
PTAL

gapis/api/vulkan/replay.go
```diff
@@ -340,57 +343,109 @@ func getReplayTypeName(request replay.Request) string {
...
+		depth++
+		devOS := abi.GetOS()
+		traceOS := h.GetABI().GetOS()
+		if devOS != traceOS {
+			if reasonDepth < depth {
+				reasonDepth = depth
+				reason = messages.ReplayCompatibilityIncompatibleOs(devOS.String(), traceOS.String())
+			}
+			continue
+		}
```
hevrard, line 381: Extract that out of the ABI loop.

gapis/api/vulkan/replay.go
```diff
@@ -340,57 +343,109 @@ func getReplayTypeName(request replay.Request) string {
...
+
 		// Memory layout must match.
+		depth++
 		if !abi.GetMemoryLayout().SameAs(h.GetABI().GetMemoryLayout()) {
+			if reasonDepth < depth {
+				reasonDepth = depth
+				reason = messages.ReplayCompatibilityIncompatibleMemoryLayout()
```
hevrard, line 400: Remove the memory layout check, the architecture check is enough.

gapis/api/vulkan/replay.go
```diff
@@ -340,57 +343,109 @@ func getReplayTypeName(request replay.Request) string {
...
-		return 0
+		return 0, messages.ReplayCompatibilityIncompatibleApi()
 	}
 
+	// As we iterate over several ABIs, remember the most detailed incompatibility reason
+	reasonDepth := 0
+	var reason *stringtable.Msg
+
 	for _, abi := range devAbis {
+		depth := 0
```
pmuetschard, line 369:
```
incompatibility = nil
for each abi {
  if dev.abi.arch != trace.abi.arch {
    continue
  }
  // do other checks, setting incompatibility
  // else if matches return "compatable"
}
if incompatibility = nil {
  incompatibility = "{trace.abi.arch} is not supported by device"
}
return incompatability
```

gapis/api/vulkan/replay.go
hevrard: done

gapis/api/vulkan/replay.go
hevrard: done

gapis/api/vulkan/replay.go
hevrard: done

hevrard:
@pmuetschard PTAL

gapis/resolve/framebuffer_attachment.go
```diff
@@ -92,11 +92,11 @@ func typeToString(t api.FramebufferAttachmentType) string {
...
-		devices, err := devices.ForReplay(ctx, p.After.Capture)
+		devices, compatibilities, _, err := devices.ForReplay(ctx, p.After.Capture)
 		if err != nil {
 			return nil, err
 		}
-		if len(devices) == 0 {
+		if len(compatibilities) == 0 || !compatibilities[0] {
 			return nil, fmt.Errorf("No compatible devices found")
 		}
 		r.ReplayDevice = devices[0]
```
pmuetschard, line 102: Yikes. This needs to be removed. Can you add a TODO on top of the if for this, please?

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -99,21 +102,59 @@ public void onCaptureLoaded(Loadable.Message error) {
   protected void resetReplayDevice() {
     replayDevices = null;
     selectedReplayDevice = null;
+    incompatibleReplayDevices = null;
+  }
+
+  public class ReplayDeviceInfo {
```
pmuetschard, line 108: This should be static. And can you move it to the bottom please. I'd say, put it right after the `Listener` interface.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -99,21 +102,59 @@ public void onCaptureLoaded(Loadable.Message error) {
...
+    public Device.Instance instance;
+    public Boolean compatible;
+    public Stringtable.Msg reason;
```
pmuetschard, line 111: This should all be final.

pmuetschard (commented):
Looks good, only minor comments.

gapic/src/main/com/google/gapid/views/DeviceDialog.java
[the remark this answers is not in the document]
stellama0208: I see, thanks for explaining! This detail is thoughtful!

stellama0208 (commented):
Looks mostly ready to go!

gapis/resolve/framebuffer_attachment.go
hevrard: You mean remove the whole `if r.ReplayDevice == nil {` ?

gapic/src/main/com/google/gapid/models/Devices.java
hevrard: done

gapic/src/main/com/google/gapid/models/Devices.java
hevrard: done

gapic/src/main/com/google/gapid/views/DeviceDialog.java
```diff
@@ -173,54 +174,91 @@ protected Control createDialogArea(Composite parent) {
...
+      incompatibleDeviceTable.setLabelProvider(new LabelProvider() {
+        @Override
+        public String getText(Object element) {
+          return Devices.getLabel(((ReplayDeviceInfo) element).instance);
+        }
+      });
+      Widgets.createTableColumn(incompatibleDeviceTable, "Name", dev -> ((ReplayDeviceInfo)dev).instance.getName());
+      Widgets.createTableColumn(incompatibleDeviceTable, "Serial", dev -> ((ReplayDeviceInfo)dev).instance.getSerial());
+      Widgets.createTableColumn(incompatibleDeviceTable, "GPU", dev -> ((ReplayDeviceInfo)dev).instance.getConfiguration().getHardware().getGPU().getName());
+      Widgets.createTableColumn(incompatibleDeviceTable, "Driver version", dev -> Integer.toUnsignedString(((ReplayDeviceInfo)dev).instance.getConfiguration().getDrivers().getVulkan().getPhysicalDevices(0).getDriverVersion()));
```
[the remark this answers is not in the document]
hevrard: Woops, done.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -99,21 +102,59 @@ public void onCaptureLoaded(Loadable.Message error) {
...
+
+            return MoreFutures.combine(Arrays.asList(allDevices), both -> {
+              MoreFutures.Result<List<Device.Instance>> devRes = both.get(0);
+              if (devRes.hasFailed()) {
+                throw devRes.error;
+              }
+              List<ReplayDeviceInfo> replayDevs = Lists.newArrayList();
+              for (int i = 0; i < devRes.result.size(); ++i) {
+                replayDevs.add(new ReplayDeviceInfo(devRes.result.get(i), compatibilities.get(i), reasons.get(i)));
+              }
```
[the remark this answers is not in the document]
hevrard: Thanks for this! The combine was a leftover from previous code which needed it to deal with a double list of devices. Using your suggestion works fine, thanks.

stellama0208 (approved):
LGTM and verified on my end! The new dialog with device failing reason is real helpful!

gapis/resolve/framebuffer_attachment.go
hevrard: TODO added.

PR #401, opened 2020-07-21
Basic track filtering feature.

pmuetschard:
Filters tracks based on track title. If a group's title matches the search term, all children are included, even if a child may not match.

If a group contains a child that matches, the group itself is shown as well, but only containing children that match. This is so that context is not lost.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/canvas/PanelGroup.java
+++ b/gapic/src/main/com/google/gapid/perfetto/canvas/PanelGroup.java
@@ -23,16 +23,18 @@ import java.util.List;
 /**
  * Contains a list of child {@link Panel panels}, which are laid out vertically.
  */
-public class PanelGroup extends Panel.Base {
+public class PanelGroup extends Panel.Base implements Panel.Grouper {
   private final List<PanelGroup.Child> panels = Lists.newArrayList();
 
   public PanelGroup() {
   }
 
-  public int size() {
+  @Override
+  public int getPanelCount() {
     return panels.size();
   }
 
+  @Override
   public Panel getPanel(int idx) {
     return panels.get(idx).panel;
   }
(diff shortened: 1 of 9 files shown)


gapic/src/main/com/google/gapid/perfetto/canvas/PanelGroup.java
```diff
@@ -167,17 +175,31 @@ private Child findPanel(double y) {
...
     public double getNextY() {
       return y + h;
     }
+
+    public void setVisibile(boolean visible) {
+      this.visible = visible;
+    }
+
+    public void setFiltered(boolean include) {
+      this.include = include;
```
silence-do-good, line 198: Can we rename **include** to **filtered** in this class? It will be consistent with setFiltered then.

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
```diff
@@ -208,7 +213,7 @@ public void stop() {
   }
 
   private static class Group<T extends CopyablePanel<T> & TitledPanel, D extends CopyablePanel<D>>
-      extends Panel.Base implements CopyablePanel<Group<T, D>> {
+      extends Panel.Base implements CopyablePanel<Group<T, D>>, Panel.Grouper, FilterablePanel {
     private final T summary;
     private final CopyablePanel.Group detail;
     protected final BiConsumer<CopyablePanel.Group, Boolean> filter;
```
silence-do-good, line 219: Unable to highlight more unchanged lines here. But can we rename the members filter and filtered in this class since their purpose is for expand/collapse? It's a little confusing as we are using the word filter differently now.

gapic/src/main/com/google/gapid/perfetto/canvas/PanelGroup.java
pmuetschard: Done.

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
pmuetschard: Done.

pmuetschard:
@silence-do-good What you are describing is done on purpose, but we can change the behavior if desired (i.e. pinning a group pins all its children, not just the currently visible ones). Note, what's confusing you is the rendering of groups followed by tracks. There is no clear distinction where the group ends, so if you pin a group and a single track, it looks kinda like the track belongs to the pinned group, while it doesn't. This has always existed and you can see it if pinning a group and track even without filtering. Maybe there is some rendering change we could make to make it more apparent where a group ends.

Currently you can only pin something once. If you pin a group and a track from within the group, the track does show up twice: once in the group if expanded and once on its own. This is due to some groups, when collapsed, showing their own data. E.g. a process track is a group and shows the process's CPU usage when collapsed. So it's helpful to be able to pin the collapsed group and possibly one of its children to see them side-by-side. Since you can only pin a group once, you can't do filter by vertex->pin group, filter by fragment->pin group again and now have two groups.

PR #402, opened 2020-07-21
Add gpu memory counter tracks

silence-do-good:
This PR includes the following changes:

1. Counter tracks for Global GPU memory usage and Per process total usage. Design doc : [https://docs.google.com/document/d/149ArKJi1E3ZJNcrp3Tfo9oj9FLRMomdBqtFMknE-m_w/edit#heading=h.h4nppfches2w](url)
2. Update perfetto to get the trace processor related changes for GPU memory support
3. Fix a typo for a static variable name

Bug: 157152869

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
@@ -58,7 +58,7 @@ import java.util.stream.Collectors;
  * Determines what tracks to show for a trace.
  */
 public class Tracks {
-  private static final Logger LOG = Logger.getLogger(CounterInfo.class.getName());
+  private static final Logger LOG = Logger.getLogger(Tracks.class.getName());
 
   // CPU usage percentage at which a process/thread is considered idle.
   private static final double IDLE_PERCENT_CUTOFF = 0.001; // 0.1%.
--- a/gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
@@ -16,9 +16,10 @@
 package com.google.gapid.perfetto.models;
 
 import static com.google.gapid.perfetto.views.StyleConstants.DEFAULT_COUNTER_TRACK_HEIGHT;
-import static com.google.gapid.perfetto.views.StyleConstants.PROCESS_COUNTER_TRACK_HIGHT;
+import static com.google.gapid.perfetto.views.StyleConstants.PROCESS_COUNTER_TRACK_HEIGHT;
 import static com.google.gapid.perfetto.views.TrackContainer.group;
 import static com.google.gapid.perfetto.views.TrackContainer.single;
+import static java.util.logging.Level.WARNING;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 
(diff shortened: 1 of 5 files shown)


lpy:
Add bug id please

zhangyiwei:
Hi AGI folks, can we have a quick review on this PR? Some vendors are waiting for it to be released so to verify their tracepoint locally. Thanks!

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -50,12 +51,15 @@
...
 import java.util.Map;
 import java.util.Objects;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 /**
  * Determines what tracks to show for a trace.
  */
 public class Tracks {
+  private static final Logger LOG = Logger.getLogger(CounterInfo.class.getName());
```
stellama0208, line 61: By convention this logger is shared by all things within this class, so I think maybe the input parameter should be `Tracks.class.getName()` if you are really using the logger.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -133,13 +137,29 @@ private Tracks() {
         .filter(c -> c.count > 0)
         .collect(toMap(c -> c.id, c -> c));
 
-    if (counters.isEmpty() && data.getGpu().isEmpty()) {
+    List<CounterInfo> gpuMemGlobalCounter = data.getCounters().values().stream()
```
stellama0208, line 140: nit: it looks like an early filtering of `data.getCounters(CounterInfo.Type.Gpu)` can be helpful to avoid filtering too many counters.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
silence-do-good: Nice catch. Updated it now.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
silence-do-good: The GPU memory counters are separate from the GPU counters that come out from the vendor producer. The one that I added is a global counter and since its basically related to the GPU,  we are putting it inside the GPU group. This check is needed because even if the GPU counter producer is not present, this can output data and hence the GPU group should be present in the UI. Its a little repetitive because a stream cannot be operated more than once. The alternate I can think of is by iterating through the map and manually putting it to either the GPU counters or the global gpu memory usage.

lpy:
LGTM on the way we enable GPU memory ftrace tracepoint, I will defer the UI code review to others.

zhangyiwei:
I've attached a screenshot for the UI with this tracepoint.

image

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
stellama0208: Sorry I pasted the wrong thing, I was meaning `data.getCounters(CounterInfo.Type. Global)`, which takes advantage of the existing map, and may ease some burden on java stream's filtering when there are too many counters.

silence-do-good: Discussed this offline. Closing this for now.

PR #403, opened 2020-07-21
Fifth part of the new transform system

yalcinmelihyasin:
It contains:

- Timestamp Replay and all the transforms that required for this replay type
- Refactor of the replay file as this is the last replay type that we have

This will be followed by a cleanup to remove the remnants of the old transform system.

For Q&A:
go/agi-transform-review


yalcinmelihyasin:
This is the last big PR. So I would be happy if we get a review sooner.

gapis/api/vulkan/transform_query_timestamps.go
```diff
@@ -0,0 +1,624 @@
...
+		cmd := cb.VkDestroyQueryPool(
+			info.device,
+			info.queryPool,
+			memory.Nullptr,
+		)
+		outputCmds = append(outputCmds, cmd)
+	}
+
+	// Melih TODO: ask about this
+	// for commandPoolkey, commandPool := range timestampTransform.commandPools {
```
AWoloszyn, line 612: ???

yalcinmelihyasin: PM

hevrard: Has this been clarified?

Also, I think it would be better to have a discussion here (rather than on private messages) for all reviewers to benefit from it :)

yalcinmelihyasin:
> It looks like the main big changes are gapis/api/vulkan/query_timestamps.go becoming gapis/api/vulkan/transform_query_timestamp.go, and the `API.Query*()` function being moved from gapis/api/vulkan/replay.go into the new gapis/api/vulkan/replay_types.go.
> 
> I'm surprised to see many smaller files being deleted:
> 
> ```
> capture_log.go
> file_log.go
> end_of_replay.go
> mapping_exporter.go
> wait_for_fence.go
> ```
> 
> As this is a massive change to review, could you explain a bit what's happening to the code in those files? Is it being deleted because it's now unused, or being refactored in some other places?

I think we have discussed this in previous meetings and PR. Because they are used by different types of replay, newer versions have been introduced previously. As this is the last replay type, now their old versions are being removed.

newer versions have the name transform_old_name.

gapis/api/transform2/command_id.go
```diff
@@ -0,0 +1,50 @@
...
+
+func NewCommandID(id api.CmdID, cmdType CommandType) CommandID {
+	return CommandID{
+		id:          id,
+		commandType: cmdType,
+	}
+}
+
+func (c *CommandID) GetID() api.CmdID {
+	return c.id
```
purvisa-at-google-com, line 45: Should panic, error or do something else if it's BeginCommand or EndCommand

gapis/api/transform2/command_id.go
```diff
@@ -0,0 +1,50 @@
...
+// and a transform chain to run all of them.
+package transform2
+
+import (
+	"github.com/google/gapid/gapis/api"
+)
+
+type CommandType uint32
+
+const (
```
purvisa-at-google-com, line 26: BeforeCommands, RealCommand, AfterCommands? The names here are a little confusing.

gapis/api/transform2/command_id.go
```diff
@@ -0,0 +1,50 @@
...
+	TransformCommand CommandType = 1
+	EndCommand       CommandType = 2
+)
+
+type CommandID struct {
+	id          api.CmdID
+	commandType CommandType
+}
+
+func NewCommandID(id api.CmdID, cmdType CommandType) CommandID {
```
purvisa-at-google-com, line 37: I think we should have different constructors for the basic command types. NewBeforeCommands() NewRealCommand(uint32 id) NewAfterCommand()

As it is, you have to specify an id argument even when it doesn't make sense to for Begin and End.

gapis/api/transform2/transform_chain.go
```diff
@@ -185,7 +187,7 @@ func (chain *TransformChain) GetNextTransformedCommands(ctx context.Context) err
 		log.E(ctx, "Replay error (%v:%v): %v", chain.currentCommandID, currentCommand, err)
 	}
 
-	chain.currentCommandID++
+	chain.currentCommandID.id++
```
purvisa-at-google-com, line 190: I think we should have an id.NextAfter() that returns a new ID and errors if you're trying to NextAfter on a Begin or End.

gapis/api/transform2/command_id.go
yalcinmelihyasin: Open to any suggestions.

gapis/api/transform2/command_id.go
yalcinmelihyasin: Makes sense.

gapis/api/transform2/transform_chain.go
yalcinmelihyasin: I am not very sure about making the commandID struct too clever but I will take a look at this.

gapis/api/transform2/command_id.go
yalcinmelihyasin: Same with the last comment. But I will take a look at this.

gapis/api/transform2/command_id.go
yalcinmelihyasin: Only thing. I refrain to use "real" because it does not really implies what it means.

gapis/api/transform2/transform_chain.go
yalcinmelihyasin: I just renamed NextAfter to Increment

gapis/api/transform2/command_id.go
yalcinmelihyasin: I though about this but if we change this, now we get a map that beginTransform() produces BeforeCommand, commandGenerator produces RealCommand and EndTransform() produces AfterCommand which I do not think it makes it better and actually makes it even more confusing. 

If you disagree, we can talk about renaming when you are back and we can change it with a later pr. It is a small change to implement anyway.

PR #404, opened 2020-07-22
Make sure to report devices that do not support profiling

hevrard:
Android devices that do not support GPU profiling should still be
listed by GAPIS. In particular, this is needed to list them in the UI
when selecting a replay device: they will be listed as invalid
devices, but at least the user can see that AGI is aware of these
devices.

In practice, this change makes sure fetchDeviceInfo does not return an
error when there is not GPU profiling support.

Bug: b/161878283
Test: manual

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -496,14 +496,7 @@ func (b *binding) DriverVersionCode(ctx context.Context) (int, error) {
 	return ip.VersionCode, err
 }
 
-// PrepareGpuProfiling queries GPU profiling support, and when profiling is supported it sets up
-// the device for profiling of installedPackage app. It returns:
-// - a bool which is true when GPU profiling is supported and the setup is done without errors
-// - a string that contains the name of the package where to find profiling layers, if profiling
-//   is supported by the system image, then this string is empty
-// - a cleanup function to revert the device settings after profiling is done
-// - an error to indicate if anything went wrong
-// The returned bool disambiguates between "an error happened" and "profiling is not supported".
+// PrepareGpuProfiling implements the adb.Device interface.
 func (b *binding) PrepareGpuProfiling(ctx context.Context, installedPackage *android.InstalledPackage) (bool, string, app.Cleanup, error) {
 	driver, err := b.GraphicsDriver(ctx)
 	if err != nil {
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -499,7 +499,8 @@ func (b *binding) DriverVersionCode(ctx context.Context) (int, error) {
 // PrepareGpuProfiling queries GPU profiling support, and when profiling is supported it sets up
 // the device for profiling of installedPackage app. It returns:
 // - a bool which is true when GPU profiling is supported and the setup is done without errors
-// - a string that contains the name of the package where to find profiling layers
+// - a string that contains the name of the package where to find profiling layers, if profiling
+//   is supported by the system image, then this string is empty
 // - a cleanup function to revert the device settings after profiling is done
 // - an error to indicate if anything went wrong
 // The returned bool disambiguates between "an error happened" and "profiling is not supported".
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -496,29 +496,49 @@ func (b *binding) DriverVersionCode(ctx context.Context) (int, error) {
 	return ip.VersionCode, err
 }
 
-// HasGpuProfilingSupportInSystemImage returns whether system image has GPU profiling support.
-func (b *binding) HasGpuProfilingSupportInSystemImage(ctx context.Context) (bool, error) {
+// PrepareGpuProfiling queries GPU profiling support, and when profiling is supported it sets up
+// the device for profiling of installedPackage app. It returns:
+// - a bool which is true when GPU profiling is supported and the setup is done without errors
+// - a string that contains the name of the package where to find profiling layers
+// - a cleanup function to revert the device settings after profiling is done
+// - an error to indicate if anything went wrong
+// The returned bool disambiguates between "an error happened" and "profiling is not supported".
+func (b *binding) PrepareGpuProfiling(ctx context.Context, installedPackage *android.InstalledPackage) (bool, string, app.Cleanup, error) {
+	driver, err := b.GraphicsDriver(ctx)
+	if err != nil {
+		// If there's an error, keep going to attempt to use GPU profiling
+		// libraries in system image.
+		log.W(ctx, "Failed to query developer driver: %v, assuming no developer driver found.", err)
+	}
+
+	if driver.Package != "" {
+		log.I(ctx, "Using GPU profiling libraries from developer driver package: %v.", driver.Package)
+
+		// Set up device info service to use prerelease driver.
+		cleanup, err := SetupPrereleaseDriver(ctx, b, installedPackage)
+		if err != nil {
+			return false, "", cleanup, err
+		}
+		return true, driver.Package, cleanup, nil
+	}
+
+	// For Android 11, GPU profiling could be part of the system driver. It can be implemented as
+	// a vulkan layer. Hence check whether GPU profiling is supported as part of the system driver,
+	// and append the vulkan profiling layer apk package name as a place to discover the vulkan
+	// profiling library.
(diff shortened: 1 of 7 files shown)


gapidapk/deviceinfo.go
```diff
@@ -144,7 +151,7 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
 	}
 	defer cleanup.Invoke(ctx)
 
-	if d.Instance().GetConfiguration().GetOS().GetAPIVersion() >= 29 {
+	if gpuProfilingSupported && d.Instance().GetConfiguration().GetOS().GetAPIVersion() >= 29 {
```
hevrard, line 154: Maybe we can get rid of the API level check here, since gpuProfilingSupported should suffice?

gapidapk/deviceinfo.go
```diff
@@ -108,6 +108,7 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
 
 	packages := []string{}
 	driver, err := d.GraphicsDriver(ctx)
+	gpuProfilingSupported := false
```
lpy, line 111: This block of code is used in two places I think, which I plan to clean up. So maybe Hugues you could step in to lift this block into a function? Sorry for adding more work, otherwise I can accept it as a temporary workaround to unblock people

hevrard: I can have a go at this tomorrow.

hevrard:
@lpy I've tried a refactoring, please take a look. I've squashed all in a single commit, copied here:

===========

Android devices that do not support GPU profiling should still be
listed by GAPIS. In particular, this is needed to list them in the UI
when selecting a replay device: they will be listed as invalid
devices, but at least the user can see that AGI is aware of these
devices.

In practice, this change refactors the GPU profiling setup functions,
and makes sure gapidapk.fetchDeviceInfo() does not return an error
when there is not GPU profiling support.

We had a very similar code duplicated in gapidapk.fetchDeviceInfo(),
gapidapk.ensurePerfettoProducerLaunched() and
gapis/perfetto/android/trace.go:Start() to query the support of GPU
profiling and setup the device accordingly. This change refactors this
code in a single new function: adb.Device.PrepareGpuProfiling().

This new function does both profiling support query and device setup,
since we never query without doing the setup if support is
detected. Both HasGpuProfilingSupportInSystemImage() and
GetGpuProfilingLayerPackageName() become obsolete and are inlined in
the new function, since this is the only place where they are used.

Profiling support is not checked in ensurePerfettoProducerLaunched()
since this function is only called from fetchDeviceInfo(), and only if
support has already been checked.

Bug: b/161878283
Test: manual

core/os/android/adb/commands.go
```diff
@@ -496,29 +496,49 @@ func (b *binding) DriverVersionCode(ctx context.Context) (int, error) {
 	return ip.VersionCode, err
 }
 
-// HasGpuProfilingSupportInSystemImage returns whether system image has GPU profiling support.
-func (b *binding) HasGpuProfilingSupportInSystemImage(ctx context.Context) (bool, error) {
-	supported, err := b.SystemProperty(ctx, systemImageGpuProfilerSupportProperty)
+// PrepareGpuProfiling queries GPU profiling support, and when profiling is supported it sets up
+// the device for profiling of installedPackage app. It returns:
+// - a bool which is true when GPU profiling is supported and the setup is done without errors
+// - a string that contains the name of the package where to find profiling layers
```
lpy, line 502: let's explicitly call out that empty string means BSP driver by default if GPU profiling is supported.

gapidapk/deviceinfo.go
```diff
@@ -106,47 +106,26 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
...
+		return err
 	}
-	if driver.Package != "" {
-		log.I(ctx, "Using GPU profiling libraries from developer driver package: %v.", driver.Package)
-
-		// Set up device info service to use prerelease driver.
-		nextCleanup, err := adb.SetupPrereleaseDriver(ctx, d, apk.InstalledPackage)
+	if supported {
+		// Set driver package
+		nextCleanup, err := android.SetupLayers(ctx, d, apk.Name, []string{packageName}, []string{})
```
pmuetschard, line 117: We used to check that `packageName` was not empty before calling `SetupLayers`. You should do that, too, or document in the commit message why you are changing this functionality in a refactor.

gapis/perfetto/android/trace.go
```diff
@@ -132,36 +132,15 @@ func Start(ctx context.Context, d adb.Device, a *android.ActivityAction, opts *s
...
+		cleanup = cleanup.Then(nextCleanup)
+		if err != nil || !supported {
+			return nil, cleanup.Invoke(ctx), log.Err(ctx, err, "GPU profiling is not supported")
 		}
-		hasRenderStages := hasDataSourceEnabled(opts.PerfettoConfig, gpuRenderStagesDataSourceName)
 
 		// Setup the profiling layers.
-		nextCleanup, err := setupProfileLayers(ctx, d, a.Package.Name, hasRenderStages, abi, packages, layers)
+		hasRenderStages := hasDataSourceEnabled(opts.PerfettoConfig, gpuRenderStagesDataSourceName)
+		nextCleanup, err = setupProfileLayers(ctx, d, a.Package.Name, hasRenderStages, abi, []string{packageName}, layers)
```
pmuetschard, line 143: Same thing here, and here I think it actually would be a problem, since in `setupProfileLayers` we append another string to the list, and thus, if `packageName` is empty, it would try to set the property to `:<gapid.package>`, which is more likely going to fail than succeed.

core/os/android/adb/commands.go
hevrard: done

gapis/perfetto/android/trace.go
hevrard: Thanks for this, I noticed a similar issue existing before my refactor in `SetupProfileLayersSource()`, where we don't check if `driver.Package` is an empty string or not: https://github.com/google/agi/blob/master/gapis/perfetto/android/trace.go#L76-L80

Rather than checking in various places that we don't add an empty string in the layer/package lists, I decided to cleanup these two lists directly in `android.SetupLayers()` by removing empty names.

Let me know if that works for you.

gapidapk/deviceinfo.go
hevrard: See other conversation.

hevrard:
@pmuetschard I edited in the way we discussed, PTAL.

core/os/android/adb/commands.go
```diff
@@ -496,29 +496,50 @@ func (b *binding) DriverVersionCode(ctx context.Context) (int, error) {
...
 }
 
-// HasGpuProfilingSupportInSystemImage returns whether system image has GPU profiling support.
-func (b *binding) HasGpuProfilingSupportInSystemImage(ctx context.Context) (bool, error) {
-	supported, err := b.SystemProperty(ctx, systemImageGpuProfilerSupportProperty)
+// PrepareGpuProfiling queries GPU profiling support, and when profiling is supported it sets up
+// the device for profiling of installedPackage app. It returns:
+// - a bool which is true when GPU profiling is supported and the setup is done without errors
+// - a string that contains the name of the package where to find profiling layers, if profiling
+//   is supported by the system image, then this string is empty
```
pmuetschard, line 503: If profiling support from the system image requires a layer, it's not empty. I think the wording here should not invoke the system image support, but rather say that the package string may be empty if no layer is required for profiling. You can document the system image vs pre-release driver part inside the function.

core/os/android/adb/commands.go
```diff
@@ -496,29 +496,50 @@ func (b *binding) DriverVersionCode(ctx context.Context) (int, error) {
...
+// - a cleanup function to revert the device settings after profiling is done
+// - an error to indicate if anything went wrong
+// The returned bool disambiguates between "an error happened" and "profiling is not supported".
```
pmuetschard, line 506: This documentation should be on the interface not here, I think.

core/os/android/adb/commands.go
hevrard: done

core/os/android/adb/commands.go
hevrard: done

PR #405, opened 2020-07-22
add mutex to grpcServer.interrupters

johnplate:
Concurrent status requests can cause gapis to crash, because the write access to grpcServer.interrupters in gapis/server/grpc.go is not thread safe and needs to be protected with a mutex.

Bug: b/161880755

### DIFF
--- a/gapis/server/grpc.go
+++ b/gapis/server/grpc.go
@@ -86,10 +86,10 @@ func NewWithListener(ctx context.Context, l net.Listener, cfg Config, srvChan ch
 		return err
 	case <-task.ShouldStop(ctx):
 		s.mutex.Lock()
+		defer s.mutex.Unlock()
 		for _, f := range s.interrupters {
 			f()
 		}
-		s.mutex.Unlock()
 		return <-done
 	}
 }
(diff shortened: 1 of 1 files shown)


gapis/server/grpc.go
```diff
@@ -85,9 +85,11 @@ func NewWithListener(ctx context.Context, l net.Listener, cfg Config, srvChan ch
 	case err := <-done:
 		return err
 	case <-task.ShouldStop(ctx):
+		s.mutex.Lock()
```
AWoloszyn, line 88:
```
s.mutex.Lock()
defer s.mutex.Unlock()
```

gapis/server/grpc.go
```diff
@@ -179,10 +182,16 @@ func (s *grpcServer) stopOnInterrupt(ctx context.Context, server *grpc.Server, s
 }
 
 func (s *grpcServer) addInterrupter(f func()) (remove func()) {
+	s.mutex.Lock()
```
AWoloszyn, line 185:
```
s.mutex.Lock()
defer s.mutex.Unlock()
```

gapis/server/grpc.go
johnplate: changed

gapis/server/grpc.go
johnplate: changed

pau-baiget:
Good catch!!

PR #407, opened 2020-07-23
Don't use debugfs node in Android 11 and beyond.

lpy:
Previously to avoid the gap at the beginning of the trace due to setup time for
ftrace when deferred start is used, AGI waits until a ftrace is enabled by
checking the node. However, debugfs node was deprecated, hence update the path.

Bug: b/161645305


lpy:
PTAL

PR #409, opened 2020-07-24
Use StyledText for various slices selection panel to allow selection.

lpy:
Label is not a selectable widget, values AGI shows in the selection panel
should be selectable by users. This patch swtiches those from Label to
StyledText.

Minor: Show command buffer handle in hex in vkQueueSubmit event slice.

Bug: b/162029786

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
@@ -19,6 +19,7 @@ import static com.google.gapid.perfetto.TimeSpan.timeToString;
 import static com.google.gapid.widgets.Widgets.createBoldLabel;
 import static com.google.gapid.widgets.Widgets.createComposite;
 import static com.google.gapid.widgets.Widgets.createLabel;
+import static com.google.gapid.widgets.Widgets.createSelectableLabel;
 import static com.google.gapid.widgets.Widgets.createTreeColumn;
 import static com.google.gapid.widgets.Widgets.createTreeViewer;
 import static com.google.gapid.widgets.Widgets.packColumns;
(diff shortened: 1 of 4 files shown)


lpy:
PTAL

gapic/src/main/com/google/gapid/widgets/Widgets.java
```diff
@@ -453,6 +456,23 @@ public static Text createTextarea(Composite parent, String text) {
     return result;
   }
 
+  public static StyledText createStyledText(Composite parent, String text, boolean editable) {
```
silence-do-good, line 459: I don't think editable is needed since there is never going to be a use case where we allow the user to edit these values. We can just silently do setEditable(false) everywhere. If at all we hit some feature where we allow the user to edit these things, we can enable then.

lpy:
> Is there any reason to use label for any of the text on the canvas?

I don't have an answer, maybe @pmuetschard can answer. My first thought is there are indeed a lot of places we don't need text to be selectable.

lpy:
PTAL

lpy:
I only checked on mac

gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
```diff
@@ -19,6 +19,8 @@
 import static com.google.gapid.widgets.Widgets.createBoldLabel;
 import static com.google.gapid.widgets.Widgets.createComposite;
 import static com.google.gapid.widgets.Widgets.createLabel;
+import static com.google.gapid.widgets.Widgets.createStyledText;
+import static com.google.gapid.widgets.Widgets.createTextbox;
```
pmuetschard, line 23: unused?

gapic/src/main/com/google/gapid/widgets/Widgets.java
```diff
@@ -453,6 +456,23 @@ public static Text createTextarea(Composite parent, String text) {
     return result;
   }
 
+  public static StyledText createStyledText(Composite parent, String text) {
```
pmuetschard, line 459: Let's give this function a better name, since it creates a very specific `StyledText`. Maybe `createSelectableLabel`?

lpy:
Thanks for the feedback and thank you so much for checking on other platforms! addressed comments! PTAL!

PR #410, opened 2020-07-24
Add duration in renderstages hover

hevrard:
Add duration in the tooltip showed when hovering a renderstage or
Vulkan event.

Bug: b/161698450
Test: manual

Before:
hover-duration-before

After:
hover-duration-after

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/GpuQueuePanel.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/GpuQueuePanel.java
@@ -206,7 +206,7 @@ public class GpuQueuePanel extends TrackPanel<GpuQueuePanel> implements Selectab
           hoveredCategory = "";
         }
         hoveredTitle = buildSliceTitle(hoveredTitle, data.args[i]);
-        hoveredTitle += " (" + TimeSpan.timeToShortString(tEnd -tStart) + ")";
+        hoveredTitle += " (" + TimeSpan.timeToString(tEnd - tStart) + ")";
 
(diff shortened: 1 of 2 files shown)


hevrard:
Woops, the duration also appears in the renderstage "body text", I'll fix this such that it appears only in the hover.

hevrard:
Now the duration appears only in the hover:

hover

gapic/src/main/com/google/gapid/perfetto/views/GpuQueuePanel.java
```diff
@@ -204,6 +206,7 @@ protected Hover onTrackMouseMove(Fonts.TextMeasurer m, double x, double y, int m
           hoveredCategory = "";
         }
         hoveredTitle = buildSliceTitle(hoveredTitle, data.args[i]);
+        hoveredTitle += " (" + TimeSpan.timeToShortString(tEnd -tStart) + ")";
```
lpy, line 209: whitespace before `-`

pmuetschard, line 209: Use `timeToString` instead of the short version. Precision here matters and slices are unlikely to be too long to warrant the short version. (ditto below)

hevrard: done

hevrard: done

PR #411, opened 2020-07-27
Pipeline View links to Input Attachments go to Framebuffer View

hliatis:
Bug: b/156906215

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Follower.java
+++ b/gapic/src/main/com/google/gapid/models/Follower.java
@@ -237,6 +237,9 @@ public class Follower {
       case RESOURCE_DATA:
         onFollowResource(path.getResourceData());
         break;
+      case FRAMEBUFFER_ATTACHMENT:
+        listeners.fire().onFramebufferAttachmentFollowed(path.getFramebufferAttachment());
+        break;
       default:
         LOG.log(WARNING, "Unknown follow path result: " + path);
     }
@@ -282,7 +285,15 @@ public class Follower {
      */
     public default void onMemoryFollowed(Path.Memory path)  { /* empty */ }
 
+    /**
+     * Event indicating that a link to the given resource was followed.
+     */
     public default void onTextureFollowed(Service.Resource resource) {}
+
+    /**
+     * Event indicating that a link with the given path to a framebuffer attachment was followed.
+     */
+    public default void onFramebufferAttachmentFollowed(Path.FramebufferAttachment path) {}
   }
 
   public static interface Prefetcher<K> {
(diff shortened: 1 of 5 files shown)


gapis/api/vulkan/resources.go
```diff
@@ -1147,6 +1147,36 @@ func commonShaderDataGroups(ctx context.Context,
...
+
+									if fbViewHandle == viewHandle {
+										paths = append(paths, cmd.FramebufferAttachmentAfter(v).Path())
+										break
+									}
+								}
+							}
+
+							currentSetData = append(currentSetData, api.CreateLinkedDataValue("url", paths, api.CreatePoDDataValue("VkImageView", viewHandle)))
+
```
pau-baiget, line 1175: Why all those extra empty lines?

hliatis: I guess I can get rid of the empty line at 1175. 1162 and 1173 separates the logic to find the input attachment index, I guess.

pau-baiget:
LGTM, but i will wait for @stellama0208 and @pmuetschard to give the actual +1 to this change. Good stuff! :)

gapis/service/path/path.go
```diff
@@ -651,6 +651,13 @@ func (n *Command) FramebufferAttachmentsAfter() *FramebufferAttachments {
 	}
 }
 
+func (n *Command) FramebufferAttachmentAfter(index uint32) *FramebufferAttachment {
```
stellama0208, line 654: nit: it would be great that these two adjacent functions could get quick annotations as others!

gapic/src/main/com/google/gapid/models/Follower.java
```diff
@@ -283,6 +286,8 @@ public default void onStateFollowed(Path.Any path) { /* empty */ }
     public default void onMemoryFollowed(Path.Memory path)  { /* empty */ }
 
     public default void onTextureFollowed(Service.Resource resource) {}
+
+    public default void onFramebufferAttachmentFollowed(Path.Any path) {}
```
stellama0208, line 290: nit: it looks like we can parse the Path.Any earlier to make type restriction here stricter?

pmuetschard: Yes, this should accept a `Path.FramebufferAttachment`

gapis/service/path/path.go
pmuetschard: Yes, all public go functions should have a comment. The comment should be in the godoc format (i.e. start with the name of the method).

PR #418, opened 2020-07-29
1 Frame Graphics Traces by Default

pmuetschard:
Bug: http://b/159113865


lpy:
A random question: Should we prevent users from inputting non-positive number?

pmuetschard:
> A random question: Should we prevent users from inputting non-positive number?

We do. The minimum has always been one frame/second.

PR #419, opened 2020-07-30
Add a short comment to some gapis RPC

hevrard:
Bug: N/A

### DIFF
--- a/gapis/service/service.proto
+++ b/gapis/service/service.proto
@@ -733,7 +733,8 @@ service Gapid {
       returns (stream GetTimestampsResponse) {
   }
 
-  // ValidateDevice checks the GPU profiling capabilities of the requested device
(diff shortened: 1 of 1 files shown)


gapis/service/service.proto
```diff
@@ -731,6 +733,8 @@ service Gapid {
       returns (stream GetTimestampsResponse) {
   }
 
+  // ValidateDevice runs validation of the GPU driver on the requested
+  // device to check that GPU counters are within an expected range.
```
lpy, line 737: to validate the GPU profiling capabilities of the requested device

hevrard: done

johnplate: Unfortunately now the line is too long :)

hevrard: Indeed! fixed.

PR #421, opened 2020-08-04
Allow Frame Profiler to use system libraries.

lpy:
Bug: b/148970509

### DIFF
--- a/gapii/client/adb.go
+++ b/gapii/client/adb.go
@@ -72,21 +72,13 @@ func Start(ctx context.Context, p *android.InstalledPackage, a *android.Activity
 		abi = p.Device.Instance().GetConfiguration().PreferredABI(nil)
 	}
 
-	driver, err := d.GraphicsDriver(ctx)
-	if err != nil {
-		return nil, nil, log.Err(ctx, err, "Failed to locate pre-release driver package")
+	if a == nil || a.Package == nil {
+		return nil, nil, log.Err(ctx, nil, "Invalid activity information.")
 	}
 
-	cleanup := app.Cleanup(func(ctx context.Context) {})
-	// Force the traced app to use the pre-release driver, or fail early
-	if driver.Package != "" && a != nil && a.Package != nil {
-		nextCleanup, err := adb.SetupPrereleaseDriver(ctx, d, a.Package)
-		cleanup = cleanup.Then(nextCleanup)
-		if err != nil {
-			return nil, cleanup.Invoke(ctx), err
-		}
-	} else {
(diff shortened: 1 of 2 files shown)


gapii/client/adb.go
```diff
@@ -72,21 +72,9 @@ func Start(ctx context.Context, p *android.InstalledPackage, a *android.Activity
...
-	// Force the traced app to use the pre-release driver, or fail early
-	if driver.Package != "" && a != nil && a.Package != nil {
-		nextCleanup, err := adb.SetupPrereleaseDriver(ctx, d, a.Package)
-		cleanup = cleanup.Then(nextCleanup)
-		if err != nil {
-			return nil, cleanup.Invoke(ctx), err
-		}
-	} else {
-		return nil, cleanup.Invoke(ctx), log.Err(ctx, nil, "Failed to locate pre-release driver package")
+	supported, _, cleanup, err := d.PrepareGpuProfiling(ctx, a.Package)
```
hevrard, line 75: I think we should keep on checking `a != nil` and `a.Package != nil`, as the old code did.

lpy:
ptal

gapii/client/adb.go
```diff
@@ -72,21 +72,11 @@ func Start(ctx context.Context, p *android.InstalledPackage, a *android.Activity
...
-	// Force the traced app to use the pre-release driver, or fail early
-	if driver.Package != "" && a != nil && a.Package != nil {
-		nextCleanup, err := adb.SetupPrereleaseDriver(ctx, d, a.Package)
-		cleanup = cleanup.Then(nextCleanup)
-		if err != nil {
-			return nil, cleanup.Invoke(ctx), err
+	if a != nil && a.Package != nil {
+		supported, _, cleanup, err := d.PrepareGpuProfiling(ctx, a.Package)
+		if err != nil || !supported {
+			return nil, cleanup.Invoke(ctx), log.Err(ctx, err, "GPU profiling is not supported.")
```
hevrard, line 78: Compilation fails because `cleanup` is now undefined later in this function, so it looks like you should still declare it outside of this block.

Also, to mimic the previous code, I think you should return an error if `a` or `a.Package` is nil.

So maybe you can refactor as:

```
if a == nil || a.Package == nil {
  return nil, nil, log.Err(ctx, nil, "Invalid activity") // maybe find a better error message...
}
supported, _, cleanup, err := d.PrepareGpuProfiling(ctx, a.Package)
if err != nil || !supported {
  return nil, cleanup.Invoke(ctx), log.Err(ctx, err, "GPU profiling is not supported.")
}
```

lpy:
ptal

PR #423, opened 2020-08-04
Force AGI to launch data producer when setup Perfetto tracing.

lpy:
Bug: b/148420473

### DIFF
--- a/gapidapk/gapidapk.go
+++ b/gapidapk/gapidapk.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/google/gapid/core/app/crash"
 	"github.com/google/gapid/core/app/layout"
-	"github.com/google/gapid/core/event/task"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/os/android"
 	"github.com/google/gapid/core/os/android/adb"
@@ -217,7 +216,9 @@ func LayerName(vulkan bool) string {
 	}
 }
 
-func ensurePerfettoProducerLaunched(ctx context.Context, d adb.Device) error {
+// EnsurePerfettoProducerLaunched kills the existing launch_producer and starts
+// a new one to launch the perfetto data producer.
+func EnsurePerfettoProducerLaunched(ctx context.Context, d adb.Device) error {
 	// Always kill the exisiting perfetto producer launcher, otherwise perfetto client
 	// may end up quering back wrong data source information.
 	d.Shell("killall", perfettoProducerLauncher).Run(ctx)
@@ -226,21 +227,17 @@ func ensurePerfettoProducerLaunched(ctx context.Context, d adb.Device) error {
 		log.W(ctx, "Failed to query developer driver: %v, assuming no developer driver found.", err)
 	}
 
-	startSignal, startFunc := task.NewSignal()
-	startFunc = task.Once(startFunc)
+	launchErr := make(chan error)
 	crash.Go(func() {
-		err := launchPerfettoProducer(ctx, d, driver, startFunc)
+		err := launchPerfettoProducer(ctx, d, driver, launchErr)
 		if err != nil {
-			log.E(ctx, "[ensurePerfettoProducerLaunched] error: %v", err)
+			log.E(ctx, "[EnsurePerfettoProducerLaunched] error: %v", err)
 		}
-
-		// Ensure the start signal is fired on failure/immediate return.
-		startFunc(ctx)
 	})
-	startSignal.Wait(ctx)
+	err = <-launchErr
 	// TODO(b/148420473): Data sources are queried before gpu.counters is registered.
 	time.Sleep(1 * time.Second)
-	return nil
+	return err
 }
 
 func preparePerfettoProducerLauncherFromApk(ctx context.Context, d adb.Device) error {
@@ -260,9 +257,10 @@ func preparePerfettoProducerLauncherFromApk(ctx context.Context, d adb.Device) e
 	return nil
 }
 
-func launchPerfettoProducer(ctx context.Context, d adb.Device, driver adb.Driver, startFunc task.Task) error {
+func launchPerfettoProducer(ctx context.Context, d adb.Device, driver adb.Driver, launchErr chan<- error) error {
 	// Extract the producer launcher from the APK.
 	if err := preparePerfettoProducerLauncherFromApk(ctx, d); err != nil {
+		launchErr <- err
 		return err
 	}
 
@@ -281,6 +279,7 @@ func launchPerfettoProducer(ctx context.Context, d adb.Device, driver adb.Driver
 	fail := make(chan error, 1)
 	crash.Go(func() {
 		buf := bufio.NewReader(reader)
+		start := false
 		for {
 			line, e := buf.ReadString('\n')
 			switch e {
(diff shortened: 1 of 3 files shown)


gapis/perfetto/android/trace.go
```diff
@@ -172,6 +172,16 @@ func Start(ctx context.Context, d adb.Device, a *android.ActivityAction, opts *s
 		}
 	}
 
+	// Enforce the AGI to launch the perfetto data producer, this will kill the
+	// running perfetto data producer and re-launch it if it's already running.
+	// And hence, any sequential update after perfetto data sources are queried
+	// will likely lead to undefined behaviour.
+	// If there's an error happen, likely the perfetto data producer or
+	// the device has issues, log the error, ignore it and keep proceeding.
```
hevrard, line 180: The current implementation of EnsurePerfettoProducerLaunched seems to never return an error.

It feels like EnsurePerfettoProducerLaunched may silence some errors, but should still propagate errors from launchPerfettoProducer(). And here we should fail if an error is reported.

What do you think?

lpy: I am familiar with golang, but I think since `launchPerfettoProducer` is within its own function we can't propagate errors from it.

lpy:
ptal

lpy:
ping

gapis/perfetto/android/trace.go
hevrard: You can propagate the error outside of the crash.Go lambda via a channel. It's not obvious, I've seen that approach in used in the `gapir/client` package.

Here, this should do the trick:

```
--- a/gapidapk/gapidapk.go
+++ b/gapidapk/gapidapk.go
@@ -228,19 +228,22 @@ func ensurePerfettoProducerLaunched(ctx context.Context, d adb.Device) error {
 
        startSignal, startFunc := task.NewSignal()
        startFunc = task.Once(startFunc)
+       launchErr := make(chan error)
        crash.Go(func() {
                err := launchPerfettoProducer(ctx, d, driver, startFunc)
                if err != nil {
                        log.E(ctx, "[ensurePerfettoProducerLaunched] error: %v", err)
                }
+               launchErr <- err
 
                // Ensure the start signal is fired on failure/immediate return.
                startFunc(ctx)
        })
        startSignal.Wait(ctx)
+       err = <-launchErr
        // TODO(b/148420473): Data sources are queried before gpu.counters is registered.
        time.Sleep(1 * time.Second)
-       return nil
+       return err
 }
```

lpy: that would block `EnsurePerfettoProducerLaunched` from returning, right? `launchPerfettoProducer` may not return immediately.

hevrard: If I understand the task framework correctly, you're already waiting for launchPerfettoProducer() to terminate with `startSignal.Wait(ctx)`? And so, by the time `startSignal.Wait(ctx)` returns, you're guaranteed to have something in launchErr to read?

lpy: no, `startSignal.Wait(ctx)` either waits for `launchPerfettoProducer` to signal where in that case, it won't return so that means the `crash.Go` is still stuck at `launchPerfettoProducer`, or `launchPerfettoProducer` to fail and the `startFunc(ctx)` is called. So `err = <- launchErr` will stuck if everything goes well.

I suppose we should just pass `launchErr` in probably.

hevrard: Oh yeah, my naive approach here would deadlock, sorry! Passing launchErr to `launchPerfettoProducer` sounds good.

lpy:
ptal

gapidapk/gapidapk.go
```diff
@@ -228,19 +230,22 @@ func ensurePerfettoProducerLaunched(ctx context.Context, d adb.Device) error {
 
 	startSignal, startFunc := task.NewSignal()
 	startFunc = task.Once(startFunc)
+	launchErr := make(chan error)
```
pmuetschard, line 233: You are always sending 2 messages on this channel, but only ever read one. Please clean it up so that it doesn't send it twice. You can also get rid of the start signal, since it's waited on/signaled at the same time as a message is sent on the channel. Having both could lead to a deadlock.

lpy:
ptal!

gapis/perfetto/android/trace.go
```diff
@@ -172,6 +172,16 @@ func Start(ctx context.Context, d adb.Device, a *android.ActivityAction, opts *s
...
 
+	// Enforce the AGI to launch the perfetto data producer, this will kill the
+	// running perfetto data producer and re-launch it if it's already running.
+	// And hence, any sequential update after perfetto data sources are queried
+	// will likely lead to undefined behaviour.
+	// If an error happens, likely the perfetto data producer or the device has
+	// issues, hence log the error, ignore it and keep proceeding.
+	if err := gapidapk.EnsurePerfettoProducerLaunched(ctx, d); err != nil {
+		log.E(ctx, "Failed to enforce agi_launch_producer to start: %v", err)
+	}
```
hevrard, line 183: Now that we get proper errors back, I think we should propagate them up here?

lpy:
ptal

I still log the error because I think it would be useful if the UI doesn't surface it

gapis/perfetto/android/trace.go
```diff
@@ -172,6 +172,15 @@ func Start(ctx context.Context, d adb.Device, a *android.ActivityAction, opts *s
 		}
 	}
 
+	// Enforce the AGI to launch the perfetto data producer, this will kill the
+	// running perfetto data producer and re-launch it if it's already running.
+	// And hence, any sequential update of the perfetto data producer or driver
+	// will likely lead to undefined behaviour.
```
pmuetschard, line 178: This comment is a bit confusing. It sounds like restarting (which is what we are doing) will cause undefined behavior. I'm assuming you are trying to say that restarting avoids the undefined behavior?

lpy: ah. I am trying to say restarting might cause undefined behaviour. The scenario, which might be very rare, in my mind is:

1) A device is plugged in, device info is collected
2) the user starts tracing and in the same time, the user updates the driver package

Then AGI will start the data producer from the new package, where counters might be different. It shouldn't crash if I understand correctly, so I it's undefined.

Maybe I should say `any sequential update of the perfetto data producer or driver after device info is queried`

lpy: Is the new comment better?

gapis/perfetto/android/trace.go
```diff
@@ -172,6 +172,16 @@ func Start(ctx context.Context, d adb.Device, a *android.ActivityAction, opts *s
 		}
 	}
 
+	// Enforce the AGI to launch the perfetto data producer, this will kill the
+	// running perfetto data producer and re-launch it if it's already running.
+	// And hence, any sequential update of the perfetto data producer or driver
+	// after device info is queried will likely lead to undefined behaviour due
+	// to potential driver behaviour change or counters set difference.
```
pmuetschard, line 179: Ok, got it. Below is how I would phrase it, feel free to use as is or inspiration.

```suggestion
	// Forcefully re-launch the Perfetto data producer in case it has gotten into
	// a bad state. This will kill any running producer and re-launch it.
	// Note that this could lead to undefined behavior due to driver behavior
	// changes or counters set differences (both are rare), if the driver was updated
	// since the producer was last started (e.g. at device discovery time). This should
	// be a very rare circumstance and can be fixed by restarting AGI.
```

gapis/perfetto/android/trace.go
lpy: merged your suggested comment. Thanks!

PR #424, opened 2020-08-04
Update GPU slice selection panel to not be vulkan specific

apbodnar:
Bug: b/161817447

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
@@ -50,7 +50,7 @@ import org.eclipse.swt.widgets.Composite;
  * Displays information about a list of selected slices.
  */
 public class SlicesSelectionView extends Composite {
-  private static final int PROPERTIES_PER_PANEL = 8;
+  private static final int PROPERTIES_PER_PANEL = 9;
   private static final int PANEL_INDENT = 25;
 
   public SlicesSelectionView(Composite parent, State state, SliceTrack.Slices slices) {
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
```diff
@@ -50,7 +50,7 @@
  * Displays information about a list of selected slices.
  */
 public class SlicesSelectionView extends Composite {
-  private static final int PROPERTIES_PER_PANEL = 8;
+  private static final int PROPERTIES_PER_PANEL = 9;
```
stellama0208, line 53: I think it's ok to keep it as 8, this constant is to set how many rows of properties displayed for each column, since we are not changing the info box's default height, maybe we should also keep this constant unchanged?

gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
```diff
@@ -104,37 +104,47 @@ private void setSingleSliceView(State state, SliceTrack.Slices slice) {
...
-          propsBuilder.put("VkCommandBuffer:", String.format("0x%08X", renderStageInfo.commandBufferHandle));
+          vkPropsBuilder.put("VkCommandBuffer:", String.format("0x%08X", renderStageInfo.commandBufferHandle));
         } else {
-          propsBuilder.put("VkCommandBuffer:", renderStageInfo.commandBufferName + " <" + String.format("0x%08X", renderStageInfo.commandBufferHandle) + ">");
+          vkPropsBuilder.put("VkCommandBuffer:", renderStageInfo.commandBufferName + " <" + String.format("0x%08X", renderStageInfo.commandBufferHandle) + ">");
         }
       }
 
       if (renderStageInfo.submissionId != 0) {
         propsBuilder.put("SubmissionId:", Long.toString(renderStageInfo.submissionId));
```
stellama0208, line 133: Question: I'm not super familiar with CG APIs, but I'm wondering whether this should also be vulkan specific info? Since it looks like submit looks like something vulkan to me.

gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
```diff
@@ -104,37 +104,47 @@ private void setSingleSliceView(State state, SliceTrack.Slices slice) {
...
 
+      ImmutableMap<String, String> vkProps = vkPropsBuilder.build();
+      if (!vkProps.isEmpty()) {
+        withLayoutData(createBoldLabel(main, "Vulkan Info:"),
+            withSpans(new GridData(), 2, 1));
+        vkProps.forEach((key, value) -> {
+          createSelectableLabel(main, key);
+          createSelectableLabel(main, value);
+        });
+      }
```
stellama0208, line 144: Nit: Just personal opinion, would we like to render general info first and vk info the next? It looks more straight forward to me to see general things first.

gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
apbodnar: I made it 9 to accommodate the "General Info" label and avoid touching your custom layout logic for the next column. Currently there's a finite amount of handles we handle so we can safely avoid more complex logic for now.  Adding more logic to make it overflow nicely into the next column seems non-trivial, but can be done if you think it's important.

gapic/src/main/com/google/gapid/perfetto/views/SlicesSelectionView.java
apbodnar: iirc, GL will also have an (internal?) event homologous with a Vk queue submission.

lpy: Right Adam, eventually GL GPU slices should expose similar submission id with a bunch of explicit submission commands. We have clarified this to all GPU vendors, so in short `submission_id` is graphics API agnostics.

PR #425, opened 2020-08-05
Use the sampling period from GPU counter spec descriptor.

lpy:
Previously AGI hard coded the min sampling period to 1ms and max sampling
period to 1000ms. This patch changes that to use the sampling period from the
GPU counter spec descriptor and modifies the time unit to microsecond in order
to allow higher sampling rate.

Bug: b/154149534

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
@@ -32,7 +32,9 @@ import static com.google.gapid.widgets.Widgets.scheduleIfNotDisposed;
 import static com.google.gapid.widgets.Widgets.withIndents;
 import static com.google.gapid.widgets.Widgets.withLayoutData;
 import static com.google.gapid.widgets.Widgets.withMargin;
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.logging.Level.WARNING;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
@@ -285,7 +287,7 @@ public class TraceConfigDialog extends DialogBase {
             .getConfigBuilder()
                 .setName("gpu.counters")
                 .getGpuCounterConfigBuilder()
-                    .setCounterPeriodNs(MILLISECONDS.toNanos(gpu.getCounterRate()));
+                    .setCounterPeriodNs(MICROSECONDS.toNanos(gpu.getCounterRate()));
         counters.addAllCounterIds(gpu.getCounterIdsList());
       }
       if (caps.getHasFrameLifecycle() && gpu.getSurfaceFlinger()) {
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -492,8 +494,26 @@ public BasicInputArea(Composite parent, Settings settings, Theme theme,
               withIndents(new GridData(), GROUP_INDENT, 0));
           gpuCountersLabels = new Label[3];
           gpuCountersLabels[1] = createLabel(counterGroup, "Poll Rate:");
-          gpuCountersRate = createSpinner(counterGroup, sGpu.getCounterRate(), 1, 1000);
-          gpuCountersLabels[2] = createLabel(counterGroup, "ms");
+
+          long minSamplingPeriod = NANOSECONDS.toMicros(gpuCaps.getGpuCounterDescriptor().getMinSamplingPeriodNs());
+          long maxSamplingPeriod = gpuCaps.getGpuCounterDescriptor().getMaxSamplingPeriodNs();
```
hevrard, line 499: Shouldn't you also have NANOSECONDS.toMicros(...) around ...getMaxSamplingPeriod**Ns**() ?

lpy:
ptal

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -492,8 +494,26 @@ public BasicInputArea(Composite parent, Settings settings, Theme theme,
               withIndents(new GridData(), GROUP_INDENT, 0));
           gpuCountersLabels = new Label[3];
           gpuCountersLabels[1] = createLabel(counterGroup, "Poll Rate:");
-          gpuCountersRate = createSpinner(counterGroup, sGpu.getCounterRate(), 1, 1000);
-          gpuCountersLabels[2] = createLabel(counterGroup, "ms");
+
+          long minSamplingPeriod = NANOSECONDS.toMicros(gpuCaps.getGpuCounterDescriptor().getMinSamplingPeriodNs());
+          long maxSamplingPeriod = NANOSECONDS.toMicros(gpuCaps.getGpuCounterDescriptor().getMaxSamplingPeriodNs());
+          maxSamplingPeriod = maxSamplingPeriod > 0 ? maxSamplingPeriod : minSamplingPeriod * 1000;
```
pmuetschard, line 500: what if the min sampling period is also 0? we should first set the min sampling period to 1ms if it's not specified.

lpy: You are right. I plan to add validation and then notify vendors although I think they all at least advertise min sampling period. In that case do we still need to set the min sampling period to a default value?

lpy: ping for feedback

silence-do-good: Maybe we can have a min value for now and add a todo for when the validation part is completed. When it has become a requirement that they should set min value, we can remove the hack then.

pmuetschard: I wouldn't call it a hack. We have a default for the max, it makes sense to have a default for the min. I would prefer a default we expect to never use over bad behavior we expect to never happen.

lpy:
Thanks for the feedback, ptal!

PR #428, opened 2020-08-08
Trim down invalid ftrace events to the time all data producers ack start.

lpy:
The ftrace can occasionally be buggy and may lead to events being included in
the trace from well before the tracing started. Until now we've just included
these events in the trace even though they really shouldn't be.

Now that Perfetto has introduced parsing of the tracing_started
TraceServiceEvent which gives the timestamp where tracing started (before any
producer is informed of tracing being ready) as well as the timestamp of when
all data producers ack the start. This patch modifies our trace processor
wrapper to use this option to filter out all events before it.

Note that such filtering is restricted to ftrace only as it's the only data
source fully under Perfetto's control, hence it's possible that other data
producers write data from before trace start.

Bug: b/154156099

### DIFF
--- a/gapis/perfetto/cc/processor.cpp
+++ b/gapis/perfetto/cc/processor.cpp
@@ -27,6 +27,10 @@ namespace ptp = perfetto::trace_processor;
 
 processor new_processor() {
   ptp::Config config;
+  // TODO(b/154156099): In Android S, AGI should use the time when trace starts
+  // instead of the time when all data sources ack the start.
+  config.drop_ftrace_data_before =
+      ptp::DropFtraceDataBefore::kAllDataSourcesStarted;
   return ptp::TraceProcessor::CreateInstance(config).release();
 }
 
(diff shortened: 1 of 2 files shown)


lpy:
ptal

gapis/perfetto/cc/processor.cpp
```diff
@@ -27,6 +27,10 @@ namespace ptp = perfetto::trace_processor;
 
 processor new_processor() {
   ptp::Config config;
+  // TODO(b/154156099): In Android S, AGI should use the time when trace starts
```
hevrard, line 30: Maybe this should be tracked in a new bug, since this PR will arguably close b/154156099? (I'm not sure about bug tracking best practices, it's a more a question than a request)

lpy: I think we should always link to the old bug so that anyone that picks this up can have the whole picture without having to dig into a lot of logs. And then the one that picks this up can figure out whether a new bug should be filed or just work on the old bug.

hevrard: :ok_hand:

hevrard:
@lpy typo in commit message title: Tirm -> Trim

lpy:
oh no! Changing the PR title doesn't change the commit message :(

PR #429, opened 2020-08-08
Move GPU performance calculation logic from gapic to gapis.

stellama0208:
- This PR defines proto, builds up data flow framework, and moves
 the GPU performance metric's calculation from gapic to gapis.
 - The performance calculation for renderpass command node is done
 in the same way as what used to be in the client side.
 - The performance calculation for upper command nodes which contains
 renderpasses is new. And it is done by 1. derive all the parent nodes of 
 renderpass nodes that's also possible to have a performance based on 
 command index, 2. for each parent node, do a further aggregation step 
 based on all the renderpass nodes it contains.
 - Bug: b/159729267.


stellama0208:
The PR is now ready for review, thanks! :)

stellama0208:
The above force push was only to rebase to agi tot, and resolve a merge conflict with PR #466.

gapis/trace/android/gpu_performance/gpu_performance.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// Evaluate and return the appropriate aggregation method for a GPU counter.
+func getCounterAggregationMethod(counter *service.ProfilingData_Counter) service.ProfilingData_GpuPerformance_Metric_AggregationOperator {
+	// TODO: Use time-weighted average to aggregate all counters for now. May need vendor's support. Bug tracked with b/158057709.
+	return service.ProfilingData_GpuPerformance_Metric_TimeWeightedAvg
+}
+
+// Compare and return the smaller one for two uint64 numbers.
+func min(a, b uint64) uint64 {
```
pmuetschard, line 256: We have `min` and `max in `core/math/u64`.

pmuetschard (commented):
Looks good, only naming and minor comments.

stellama0208:
The above suggestions have been addressed with the new commit, PTAL thanks! :)

pmuetschard (approved):

PR #431, opened 2020-08-10
Support GpuMemTotal datasource for initial counter values

zhangyiwei:
This PR includes the following changes:

1. Add the GPU memory total datasource to the trace config if the producer is available.
2. The raw counter name of GPU Memory total has been updated, so the special mappings have been removed.

Bug: 157152869

unit
tooltip_proc
tooltip_global
initial_counter_from_producer

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -45,6 +45,7 @@ const (
 
 	maxRootAttempts                         = 5
 	gpuRenderStagesDataSourceDescriptorName = "gpu.renderstages"
+	gpuMemTotalDataSourceDescriptorName     = "android.gpu.memory"
 
 	perfettoPort = NamedFileSystemSocket("/dev/socket/traced_consumer")
 
@@ -353,6 +354,7 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
 // that support the GPU profiling functionailities, it includes:
 //     1) gpu.counters
 //     2) gpu.renderstages
+//     3) android.gpu.memory
 func (b *binding) QueryPerfettoGpuProfilingDataSources(ctx context.Context) (*device.GPUProfiling, error) {
 	gpu := &device.GPUProfiling{}
 	encoded, err := b.Shell("perfetto", "--query-raw", "|", "base64").Call(ctx)
@@ -369,6 +371,10 @@ func (b *binding) QueryPerfettoGpuProfilingDataSources(ctx context.Context) (*de
 		desc := ds.GetDsDescriptor()
 		if desc.GetName() == gpuRenderStagesDataSourceDescriptorName {
 			gpu.HasRenderStage = true
+			continue
+		}
+		if desc.GetName() == gpuMemTotalDataSourceDescriptorName {
+			gpu.HasGpuMemTotal = true
 			continue
 		}
 		counters := desc.GetGpuCounterDescriptor().GetSpecs()
--- a/core/os/device/device.proto
+++ b/core/os/device/device.proto
@@ -195,6 +195,7 @@ message GPUProfiling {
   GpuCounterDescriptor gpu_counter_descriptor = 2;
   reserved 3;
   bool has_render_stage_producer_layer = 4;
+  bool has_gpu_mem_total = 5;
 }
 
 // Instance represents a physical device.
--- a/gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
@@ -137,10 +137,13 @@ public class Tracks {
         .filter(c -> c.count > 0)
         .collect(toMap(c -> c.id, c -> c));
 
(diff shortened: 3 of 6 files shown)


zhangyiwei:
@lpy @silence-do-good @stellama0208 @pmuetschard Ptal, thanks!

core/os/android/adb/commands.go
```diff
@@ -371,6 +372,10 @@ func (b *binding) QueryPerfettoGpuProfilingDataSources(ctx context.Context) (*de
 			gpu.HasRenderStage = true
 			continue
 		}
+		if desc.GetName() == gpuMemTotalDataSourceDescriptorName {
```
lpy, line 375: update the comment of this function. We should probably add a comment to set `HasGpuMemTotal` to true always once it's enforced by CDD

zhangyiwei: Comment for this function has been updated. As for the comment for cdd enforcement, for this kernel tracking system, GMS will most likely be the proper vehicle. Do we still want to mention this? unlike Frame Life Cycle, we would always need to back support any volunteered R devices.

lpy: ACK

core/os/device/device.proto
```diff
@@ -193,7 +193,7 @@ message GPUProfiling {
   bool hasRenderStage = 1;
   // The GPU performance counters.
   GpuCounterDescriptor gpu_counter_descriptor = 2;
-  reserved 3;
```
pmuetschard, line 196: You shouldn't re-use reserved ones. They are reserved because they used to be used for something else. This could cause deserialization of old protos fail. While this isn't technically an issue with this proto, we shouldn't re-use reserved ones.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -137,10 +137,11 @@ private Tracks() {
...
 
-    List<CounterInfo> gpuMemGlobalCounter = data.getCounters().values().stream()
-        .filter(c -> c.type == CounterInfo.Type.Global && c.ref == 0 /* pid - 0 */
-        && c.count > 0 && c.name.equals("gpumemtotal.global"))
-        .collect(toList());
+    List<CounterInfo> gpuMemGlobalCounter =
+        data.getCounters().values().stream().filter(c -> c.type == CounterInfo.Type.Global &&
+                                                    c.ref == 0 /* pid - 0 */ &&
+                                                    c.count > 0 &&
+                                                    c.name.equals("GPU Memory")).collect(toList());
```
pmuetschard, line 144:
```suggestion
    List<CounterInfo> gpuMemGlobalCounter =
        data.getCounters().values().stream()
            .filter(c -> c.type == CounterInfo.Type.Global &&
                c.ref == 0 /* pid - 0 */ &&
                c.count > 0 &&
                c.name.equals("GPU Memory"))
            .collect(toList());
```

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -295,6 +295,12 @@ public static String getConfigSummary(Settings settings, Device.PerfettoCapabili
             .getConfigBuilder()
                 .setName("android.surfaceflinger.frame");
       }
+      // Skip UI checkbox for GpuMemTotal producer to align with its ftrace events
```
pmuetschard, line 298: I don't understand this comment. Are you trying to explain why we don't allow the user to disable this? If so, can you phrase it like this: "Always enable GPU memory, if available, because [insert reason here]"

tools/build/workspace.bzl
```diff
@@ -185,8 +185,8 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
         name = "perfetto",
         locals = locals,
         remote = "https://android.googlesource.com/platform/external/perfetto",
-        commit = "bbba7b0b884545401dbf5272819823752812412b",
-        shallow_since = "1595504601 +0100",
+        commit = "1dd60f85b97ef5068e9a39a4165087ccbdf6b09f",
+        shallow_since = "1596837015 -0700",
```
pmuetschard, line 189: #428 just updated this, can you make sure you are not downgrading what it updated to?

core/os/device/device.proto
zhangyiwei: Done.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
zhangyiwei: Done.

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
zhangyiwei: Done.

tools/build/workspace.bzl
zhangyiwei: It's an upgrade. Don't worry.

PR #432, opened 2020-08-11
Emit warnings when replay instance creation fails

hevrard:
On Android, the replay happens in gapid APK, where any layer/extension
provided by a library in the original APK will not be present. So if
the original app depends on such layers, the instance creation will
fail. Still, we cannot simply return false upon creation failure here,
as the app may also have tried to create an instance and failed, and
will subsequently re-try to create another instance with different
parameters. Yet, instance creation is likely to be expected to be
successful most of the time, we do emit a debug message upon failure.

Bug: b/158671749

### DIFF
--- a/gapir/cc/context.cpp
+++ b/gapir/cc/context.cpp
@@ -240,6 +240,33 @@ void Context::registerCallbacks(Interpreter* interpreter) {
                 "debug report extension, drop them and try again");
             if (api->replayCreateVkInstanceImpl(stack, pCreateInfo, pAllocator,
                                                 pInstance, true, &result)) {
+              // On Android, the replay happens in gapid APK, where any
+              // layer/extension provided by a library in the original APK will
+              // not be present. So if the original app depends on such layers,
+              // the instance creation will fail. Still, we cannot simply return
+              // false upon creation failure here, as the app may also have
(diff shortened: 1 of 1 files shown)


hevrard:
> Update: I'm wrong, it returns true and forwards the result if callVkCreateDevice was called. Is it intended not to output a debug message if the function returns false?

replayCreateVkInstanceImpl returns two things:
- a boolean to indicate if calling the Vulkan function was achieved without hitting a replayer-level error
- the actual result of the vulkan call, stored in `result`, which may or may not be VK_SUCCESS -- and I think it might be expected to not be VK_SUCCESS, since the original app may have tried to create an instance and have it failed, we would still have recorded this attempt.

In the lambda that defines Builtins::ReplayCreateVkInstance, if the second call to replayCreateVkInstanceImpl() returns false, then the built-in also returns false, with a Fatal debug message. If I understand your question correctly, this is the case your were asking about, so it looks like we do emit a message and propagate the `false` in that case.

My changes add the following: if the second call to replayCreateVkInstanceImpl() returns true (== the subsequent vulkan call was made without errors), then we peek at the result and emit a warning if it is different from VK_SUCCESS.

PR #436, opened 2020-08-25
Sixth part of the new transform system

yalcinmelihyasin:
This is a cleanup to remove the remnants of the old transform system.

For Q&A:
go/agi-transform-review


yalcinmelihyasin:
This is the last part of the transform system refactor.

After deciding the fate of existing frame looping code, we can merge this.

Bringing back DCE and frame looping will be done later, depending on their priorities.

yalcinmelihyasin:
This is mostly removing things and renaming. But the interesting changes that should be given more attention has happened in transform_chain, sync and transform_vulkan_terminator..

gapis/api/controlFlowGenerator/linear_control_flow_generator.go
```diff
@@ -16,28 +16,68 @@ package controlFlowGenerator
...
+	var id transform.CommandID
+	defer recoverFromPanic(id)
+
+	subctx := keys.Clone(context.Background(), ctx)
 	for !cf.chain.IsEndOfCommands() {
-		if err := cf.chain.GetNextTransformedCommands(ctx); err != nil {
+		id = cf.chain.GetCurrentCommandID()
+		if id.GetCommandType() == transform.TransformCommand {
+			cmdID := uint64(id.GetID())
+			if cmdID%100 == 99 {
```
purvisa-at-google-com, line 52: nit: == 0 is nicer for these kind of comparisons.

gapis/api/commandGenerator/command_generator.go
```diff
@@ -32,4 +32,7 @@ type CommandGenerator interface {
 	// IsEndOfCommands should return true when there is no more command
 	// to be produced.
 	IsEndOfCommands() bool
+
+	// GetNumberOfCommands should return total number of commands to be generated
+	GetNumberOfCommands() uint64
```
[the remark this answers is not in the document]
yalcinmelihyasin: 👍

gapis/api/transform/transform.go
```diff
@@ -33,7 +33,7 @@ type StateMutator func(cmds []api.Cmd) error
 // they aim to do and emit a list of commands to pass it to the next transform.
 type Transform interface {
 	// BeginTransform is called before transforming any command.
-	BeginTransform(ctx context.Context, inputCommands []api.Cmd, inputState *api.GlobalState) ([]api.Cmd, error)
+	BeginTransform(ctx context.Context, inputState *api.GlobalState) error
```
purvisa-at-google-com, line 36: We've agreed with Pascal to keep this change. But for the record on this code review: I still think this function should have a symmetrical signature with EndTransform() so that commands can be inserted at the start of a replay as well as the end.

purvisa-at-google-com (approved):

PR #440, opened 2020-08-31
Add new ANGLE package name

tobine:
Chromium-built ANGLE package name is being updated to
org.chromium.angle so add that in and favor it as the first custom-
built ANGLE apk name.

Bug: b/167223999

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -406,15 +406,16 @@ func (b *binding) QueryAnglePackageName(ctx context.Context) (string, error) {
 	}
 	// ANGLE supported, so check for installed ANGLE package
 	// Favor custom installed package first, followed by default system package
-	custom := "com.chromium.angle"
+	//  org.chromium.angle is first because it's the most recent & preferred name
+	custom := [2]string{"org.chromium.angle", "com.chromium.angle"}
 	system := "com.google.android.angle"
 	// Check installed packages for ANGLE package
 	packages, _ := b.InstalledPackages(ctx)
 	if pkg := packages.FindByName(custom); pkg != nil {
-		return custom, nil
+		return pkg.Name, nil
 	}
 	if pkg := packages.FindByName(system); pkg != nil {
-		return system, nil
+		return pkg.Name, nil
 	}
 	return "", fmt.Errorf("No ANGLE packages installed on this device")
 }


core/os/android/adb/commands.go
```diff
@@ -406,15 +406,16 @@ func (b *binding) QueryAnglePackageName(ctx context.Context) (string, error) {
...
-	custom := "com.chromium.angle"
+	//  org.chromium.angle is first because it's the most recent & preferred name
+	custom := [2]string{"org.chromium.angle", "com.chromium.angle"}
 	system := "com.google.android.angle"
 	// Check installed packages for ANGLE package
 	packages, _ := b.InstalledPackages(ctx)
 	if pkg := packages.FindByName(custom); pkg != nil {
-		return custom, nil
+		return pkg.Name, nil
 	}
```
pmuetschard, line 416: You need a loop here now that `custom` is a slice, no?

core/os/android/adb/commands.go
```diff
@@ -406,15 +406,16 @@ func (b *binding) QueryAnglePackageName(ctx context.Context) (string, error) {
...
 	// ANGLE supported, so check for installed ANGLE package
 	// Favor custom installed package first, followed by default system package
-	custom := "com.chromium.angle"
+	//  org.chromium.angle is first because it's the most recent & preferred name
+	custom := [2]string{"org.chromium.angle", "com.chromium.angle"}
```
lpy, line 410: Is ANGLE going to be `org.chromium.angle` moving forward? If yes, why can't everyone switch to it?

core/os/android/adb/commands.go
tobine: Yes, I misread FindByName thinking that it took a slice and processed those strings in a loop, but it only takes a string and processes the package names in a loop.
I think the easier and cleaner thing to do in the long-term is just switch to the org.chromium.angle name only going fwd. I'm pushing that now so the com.chromium.angle name should die quickly and better to not keep the incorrect name alive.

core/os/android/adb/commands.go
tobine: Yes, I agree that just switching is cleaner going fwd. Updating to that.

tobine:
Closing this in favor of simpler https://github.com/google/agi/pull/442 which just moves to org package only.
I was thinking of keeping both to facilitate a transition period, but as this is an experimental feature that almost no one is currently using and the org.chromium.angle package name will be the correct one going forward I think it's better (as well as simpler) to just let the com.chromium.angle package name die.

PR #442, opened 2020-09-01
Add new ANGLE package name

tobine:
Chromium-built ANGLE package name is being updated to
org.chromium.angle so use that the first custom-built ANGLE apk name.

Bug: b/167223999


tobine:
@hevrard how did we make sure the ANGLE apk is installed on the test systems? We need to update that to a new APK with the org.chromium.angle package name as well.

hevrard:
@tobine uploading the ANGLE APK to swarming devices was done manually as a one-time process. Let's chat offline about the best way to streamline this from now on.

PR #443, opened 2020-09-02
Fix for destroying the last instance on windows.

hevrard:
If a game destroys that last instance, some static data
will fail to get cleaned up causing a crash. Prevent
libgapii from being unloaded on windows in this case.

Cherry-pick of: https://github.com/google/gapid/pull/3843

### DIFF
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -16,6 +16,12 @@
 
 #include "spy.h"
 
+#include <cstdlib>
+#include <memory>
+#include <sstream>
+#include <thread>
+#include <vector>
+
 #include "connection_header.h"
 #include "connection_stream.h"
 #include "protocol.h"
@@ -35,11 +41,10 @@
 #include "gapis/capture/capture.pb.h"
 #include "gapis/memory/memory_pb/memory.pb.h"
 
-#include <cstdlib>
-#include <memory>
-#include <sstream>
-#include <thread>
-#include <vector>
+#if TARGET_OS == GAPID_OS_WINDOWS
+#include <windows.h>
+#include "windows/wgl.h"
+#endif  //  TARGET_OS == GAPID_OS_WINDOWS
 
 #if TARGET_OS == GAPID_OS_ANDROID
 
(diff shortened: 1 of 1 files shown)


gapii/cc/spy.cpp
```diff
@@ -35,11 +41,10 @@
 #include "gapis/capture/capture.pb.h"
 #include "gapis/memory/memory_pb/memory.pb.h"
 
-#include <cstdlib>
-#include <memory>
-#include <sstream>
-#include <thread>
-#include <vector>
+#if TARGET_OS == GAPID_OS_WINDOWS
+#include <windows.h>
```
johnplate, line 45: I propose to add the following before including windows.h to reduce compiling time:
#define WIN32_LEAN_AND_MEAN 1

johnplate (approved):

gapii/cc/spy.cpp
hevrard: Could we do this as another PR, throughout the code base?

`windows.h` is included in several other places:

```
$ git grep 'windows\.h' | grep '#include'
cmd/vulkan_sample/main.cpp:#include <windows.h>
core/cc/dl_loader.cpp:#include <windows.h>
core/cc/timer.cpp:#include "windows.h"
core/cc/windows/crash_handler.cpp:#include <windows.h>
core/cc/windows/debugger.cpp:#include <windows.h>
core/cc/windows/get_vulkan_proc_address.cpp:#include <windows.h>
core/cc/windows/thread.cpp:#include <windows.h>
core/memory/arena/cc/arena.cpp:#include <windows.h>
core/os/device/deviceinfo/cc/windows/query.cpp:#include <windows.h>
core/vulkan/layer_helpers/threading.h:#include <windows.h>
core/vulkan/vk_virtual_swapchain/cc/threading.h:#include <windows.h>
gapii/cc/call_observer.cpp:#include <windows.h>
tools/build/mingw_toolchain/file_collector.cpp:#include <windows.h>
tools/build/mingw_toolchain/gcc_wrapper.cpp.tpl:#include <windows.h>
tools/build/third_party/breakpad/dump_syms_pe.cc:#include <windows.h>
```

WIN32_LEAN_AND_MEAN seems to be used once:

```
$ git grep WIN32_LEAN_AND_MEAN
core/memory/arena/cc/arena.cpp:#define WIN32_LEAN_AND_MEAN
core/memory/arena/cc/arena.cpp:#undef WIN32_LEAN_AND_MEAN
```

So I think it would be better to do a full pass on the code to add it where possible, and also measure the compilation time impact.

pmuetschard (approved):

PR #452, opened 2020-09-10
Fix compilation-time apic warning

hevrard:
Delete useless imports that were triggering a warning.

Bug: b/168111325


pmuetschard (approved):
Thanks! I've been wanting to fix this.... since switching to bazel years ago.... 🙄

yalcinmelihyasin (approved):

PR #453, opened 2020-09-10
Bump the major version to version 1.0.0.

pmuetschard:


lpy:
I think such major version change needs to be reviewed by Jay and documented down in the plan doc. I couldn't remember what is the versioning decision we made a while back.

lpy:
Checked with Jay offline and Jay informed me this is the plan. I have asked Jay to document this down to the plan doc.

PR #455, opened 2020-09-11
Fix GAPIS termination deadlock

hevrard:
Make the error channel `c` be asynchronous to prevent a deadlock in
the termination sequence of GAPIS. The deadlock is related to the
Status() RPC, where a write to the synchronous `c` channel is blocking
since the corresponding read cannot be reached before this write has
terminated.

In gapis/server/grpc.go:grpc.Status(), the synchronous channel `c` is
created and used in the `f` lambda, which writes into it. This `f` is
passed to the call to s.handler.Status(), and this call must terminate
with a non-nil error to have `c` being read from.

It turns out the s.handler.Status() call leads to a blocking write to
`c`. Looking at its implementation in
gapis/server/server.go:server.Status(), note the status.Finish()
deferred function registered at the beginning, then the registration
of a new statusListener that embeds `f`. When the context is
cancelled, this Status() function wants to terminate, of the defered
status.Finish() is called, triggering statusListener.OnTaskFinish(),
which itself calls `f`, which tries a stream.Send(). If this
stream.Send() fails, `f` blocks on a write to `c`, effectively
preventing the termination of server.Status(). However, sever.Status()
must return for grpc.Status() to have a chance to read from `c`.

Making `c` asynchronous avoid `f` to block on the write, thus letting
sever.Status() terminate such that grpc.Server() can also terminate.

Bug: b/150773653
Test: manual, make sure AGI always terminates properly, the deadlock
is almost systematic.

### DIFF
--- a/gapis/server/grpc.go
+++ b/gapis/server/grpc.go
@@ -373,7 +373,12 @@ func (s *grpcServer) Status(req *service.ServerStatusRequest, stream service.Gap
 	ctx, cancel := task.WithCancel(stream.Context())
 	defer s.addInterrupter(cancel)()
 
-	c := make(chan error)
+	// This channel MUST be asynchronous to prevent deadlocks: upon
+	// termination, statusListener.OnTaskFinish() calls f() which may write
+	// to this channel. If this write is blocking, it prevents the
+	// s.handler.Status() call below to terminate, thus preventing also the
+	// read of channel `c`, effectively leading to a deadlock.
+	c := make(chan error, 1)
 	f := func(t *service.TaskUpdate) {
 		if err := stream.Send(&service.ServerStatusResponse{
 			Res: &service.ServerStatusResponse_Task{t},


pmuetschard (changes requested):
I'm not sure this is the correct solution. I think the correct solution is to make sure that the channel send and the channel read are run on parallel goroutines. Let me take a closer look at this.

hevrard:
@pmuetschard yeah I'm not sure it's the right fix either, at least I clarified the reason of the deadlock. I'll let you clarify a relevant cancellation / error propagation flow such that there's no risk of deadlock anymore.

silence-do-good:
This doesn't seem to fix the issue. Tested it on gLinux and gapis still doesnt exit.
`[server.ChildProcess-gapis][server.ChildProcess.onExit] The gapis process exited with a non-zero exit value: -1`

pmuetschard:
I can confirm that this doesn't fix the underlying issue. I've also tried it with separate goroutines and verified that it indeed doesn't deadlock here, but the server still hangs. I'm not saying this is not part of the issue, just that there is more going on here.

hevrard:
Woops, I declared victory too soon. This fix unblocked GAPIS termination in the following use-case that I used for debugging:

- start gapis with ./gapis -enable-local-files -rpc localhost:8888

- start gapic with ./agi --gapis localhost:8888

- close the UI

- try to Ctrl-C the terminal where you started gapis, it doesn't stop the process unless you have this fix

pmuetschard:
This doesn't fix the deadlock and is also the wrong solution for the channel issue.

PR #457, opened 2020-09-11
Use grpc's forceful stop instead of the graceful stop.

pmuetschard:
The graceful stop seems to deadlock when there are streaming RPCs.


pmuetschard:
This is a cherry-pick of #456 into the release branch.

PR #463, opened 2020-09-15
Update DEVDOC.md

yalcinmelihyasin:
New directives for debugging in VSCode with Go Plugin.

### DIFF
--- a/DEVDOC.md
+++ b/DEVDOC.md
@@ -137,7 +137,13 @@ Follow these steps to use the delve debugger for Go with VSCode to debug `gapis`
 
 2. Create a `launch.json` file under the workspace directory with `Ctrl + Shift + P` and `Debug: Open launch.json`
 
-3. Paste the following as one of the launch configurations. This will ensure that there is a launch configuration for attaching to Delve.
+3. There will be two settings file, one is for global the other one is local to the project.
+	- Add this line to your global settings to ensure you have a stable tools directory:
+	`"go.toolsGopath": "<path-to-go-plugin-tools-folder>",`
+	- Add this line to your local settings to be able to search source code in AGI:
+	`"go.gopath": "<path-to-agi-gofuse>"`,
+
+4. Paste the following as one of the launch configurations. This will ensure that there is a launch configuration for attaching to Delve.
 ```
 {
     ...
@@ -149,8 +155,8 @@ Follow these steps to use the delve debugger for Go with VSCode to debug `gapis`
             "request": "attach",
             "mode": "remote",
             "apiVersion": 2,
-            "remotePath": "gapis/",
-            "cwd": "${workspaceFolder}/src/github.com/google/agi/gapis",
+            "remotePath": "",
+            "cwd": "`<path-to-agi-gofuse>`",
             "dlvLoadConfig": {
                 "followPointers": true,
                 "maxVariableRecurse": 1,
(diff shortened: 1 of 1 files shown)


DEVDOC.md
```diff
@@ -137,7 +137,13 @@ Follow these steps to use the delve debugger for Go with VSCode to debug `gapis`
 
 2. Create a `launch.json` file under the workspace directory with `Ctrl + Shift + P` and `Debug: Open launch.json`
 
-3. Paste the following as one of the launch configurations. This will ensure that there is a launch configuration for attaching to Delve.
+3. There will be two settings file, one is for global the other one is local to the project.
```
hevrard, line 140: typo: ...two settings file**s**

I would recommend the present tense: `There are to settings files...`

DEVDOC.md
```diff
@@ -137,7 +137,13 @@ Follow these steps to use the delve debugger for Go with VSCode to debug `gapis`
...
+	- Add this line to your global settings to ensure you have a stable tools directory:
+	`"go.toolsGopath": "<path-to-go-plugin-tools-folder>",`
```
hevrard, line 142: Do you have an example of such a plugin tools folder path? (It's the first time I hear of such a thing, is it VS-Code specific?)

DEVDOC.md
```diff
@@ -181,6 +187,9 @@ dlv exec --headless --listen=127.0.0.1:1234 --api-version 2 ./bazel-bin/pkg/gapi
 
 6. Now VSCode can interact with Delve and can be used for debugging `gapis` in VSCode UI instead of the command line. Enjoy your debugging :)
 
+This will allow you to put breakpoint in any lines in Go standard library, generated or handwritten AGI Go source code. For the generated file,
```
hevrard, line 190: Present tense, and remove a few "the" and some typos:

```
This allows you to put breakpoints at any line in the Go standard library, or in
generated or handwritten AGI Go source code. For generated files, you can put
breakpoints under the  `bazel-bin/` or `bazel-out/` folder and the debugger will
still find the relevant files in the gofuse directory. The only downside is that you
may have two versions of the same file, but this is the only working workaround
until VSCode Go Plugin properly supports Bazel-generated files.
```

(Feel free to adapt as you see fit)

DEVDOC.md
yalcinmelihyasin: I am not sure if this is a VSCode specific thing or does it have other versions in other IDEs but it is the place that the plugin downloads tools into. Setting it globally is not mandatory(otherwise it uses $GOPATH) but it helps to be able to change $GOPATH per project.

hevrard: OK, thanks for the clarification. The plugin here is the VSCode go plugin.

I think go.toolsGopath and go.gopath are not in `launch.json`? If so, these new instructions should not be in the middle of the list related to `launch.json`

hevrard:
I can confirm this lets me put breakpoints in e.g. the generated `.../agi-gofuse/src/github.com/google/gapid/gapis/api/vulkan/api_state.go`.

DEVDOC.md
yalcinmelihyasin: Oh I missed that. Yes, I will move this to somewhere earlier as a pre-step. With including information about setting files.

DEVDOC.md
```diff
@@ -177,9 +185,11 @@ The command below will allow using port `1234` (or any other preferred port) to
 dlv exec --headless --listen=127.0.0.1:1234 --api-version 2 ./bazel-bin/pkg/gapis -- -persist -rpc localhost:8888
 ```
 
-5. Start debugging with `Debug->Start Debugging` (on Linux with `F5`) and make sure `Attach to Delve` is selected as the launch configuration.
+6. Start debugging with `Debug->Start Debugging` (on Linux with `F5`) and make sure `Attach to Delve` is selected as the launch configuration.
+
+7. Now VSCode can interact with Delve and can be used for debugging `gapis` in VSCode UI instead of the command line. Enjoy your debugging :)
 
-6. Now VSCode can interact with Delve and can be used for debugging `gapis` in VSCode UI instead of the command line. Enjoy your debugging :)
+This allows you to put breakpoint at any line in AGI Go source code regardless if they are handwritten, generated or in Go Standard Library. For the generated file, you can put the breakpoints under the `bazel-bin/` or `bazel-out/` folder and the debugger will still find it under the fuse directory during debugging and open it. The only downside is you will have two version of the same file but this is the only working workaround until Go Plugin supports bazel-generated files.
```
hevrard, line 192: ... you will have two version**S** of the same file ...

hevrard (approved):
:warning: I approved, but there is one last typo to fix before merging!

Thanks for this update, that will help :)

PR #465, opened 2020-09-16
Fill slices in memory struct view

hliatis:
Bug: b/165833962

Filters out duplicate entries in the slices, will probably make a separate bug to get rid of them in the first place.

### DIFF
--- a/gapis/resolve/memory.go
+++ b/gapis/resolve/memory.go
@@ -49,7 +49,10 @@ func filterTypedRanges(ranges []*service.TypedMemoryRange) []*service.TypedMemor
 	sort.Slice(ranges, func(i, j int) bool {
 		return (ranges[i].Root < ranges[j].Root ||
 			ranges[i].Root == ranges[j].Root &&
-				ranges[i].Type.TypeIndex < ranges[j].Type.TypeIndex)
+				ranges[i].Type.TypeIndex < ranges[j].Type.TypeIndex ||
+			ranges[i].Root == ranges[j].Root &&
+				ranges[i].Type.TypeIndex == ranges[j].Type.TypeIndex &&
+				ranges[i].Range.Base < ranges[j].Range.Base)
 	})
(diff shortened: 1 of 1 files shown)


gapis/resolve/memory.go
```diff
@@ -49,7 +49,10 @@ func filterTypedRanges(ranges []*service.TypedMemoryRange) []*service.TypedMemor
 	sort.Slice(ranges, func(i, j int) bool {
 		return (ranges[i].Root < ranges[j].Root ||
 			ranges[i].Root == ranges[j].Root &&
-				ranges[i].Type.TypeIndex < ranges[j].Type.TypeIndex)
+				ranges[i].Type.TypeIndex < ranges[j].Type.TypeIndex ||
+			ranges[i].Root == ranges[j].Root &&
+				ranges[i].Type.TypeIndex == ranges[j].Type.TypeIndex &&
+				ranges[i].Range.Base < ranges[j].Range.Base)
```
hliatis, line 55: This added sort criteria makes sure we see entries in order. Otherwise, later-entries that are encountered earlier may skip earlier-entries encountered later.

gapis/resolve/memory.go
```diff
@@ -65,6 +68,16 @@ func filterTypedRanges(ranges []*service.TypedMemoryRange) []*service.TypedMemor
 				end = ranges[i].Range.Base + ranges[i].Range.Size
 			}
 
+			if newRanges[last].Range.Size != end-start {
```
hliatis, line 71: This condition is being used to filter duplicates. Once we get rid of the duplication, we can get rid of this and the above sort criterion.

hliatis: By the way, items that trigger both the OnRead and OnWrite methods (e.g. VkFramebuffer in VkCreateFramebuffer) will inherently be duplicated, so we probably can't get rid of this unless we make a distinction between the two kinds of ranges. However, this doesn't account for the duplication of entries in an array.

pmuetschard (approved):

PR #466, opened 2020-09-17
Fix a NPE triggered by querying unloaded profile.

stellama0208:
- Sometimes a profile loading will fail in practice, e.g. when desktop replay profiling
 is unsupported. At performance tab's column initialization, check whether there's a
 valid profile data first before the attempt to fetch the profile's counter data.
 - Bug: b/168807497.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/PerformanceView.java
+++ b/gapic/src/main/com/google/gapid/views/PerformanceView.java
@@ -101,6 +101,10 @@ public class PerformanceView extends Composite
 
   @Override
   public void onProfileLoaded(Loadable.Message error) {
+    if (error != null) {
+      loading.showMessage(error);
+      return;
+    }
     // Create columns for counters after profile get loaded, because we need to know counter numbers.
     for (Service.ProfilingData.Counter counter : models.profile.getData().getCounters()) {
       tree.addColumnForCounter(counter);


stellama0208:
@johnplate Hey John, I assume this PR would fix the issue you described in the ticket, since I lack the environment to reproduce the bug, would appreciate it if you could verify whether it would make an actual fix, thanks! :)

gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -101,6 +101,10 @@ public void onCommandsLoaded() {
 
   @Override
   public void onProfileLoaded(Loadable.Message error) {
+    if (error != null || models.profile.getData() == null) {
+      loading.showMessage(Error, Messages.PROFILE_LOAD_FAILURE);
```
pmuetschard, line 105: Why not just `loading.showMessage(error)`?

stellama0208: When I was writing I was thinking 
1. respect the convention of how we handled `onCaptureLoaded` failure above (I thought it was written that way because we don't want to show that detail in UI?), 
2. still show something when this if clause is triggered by the second predicate `models.profile.getData() == null` (so far it is sure that if data is null after loading, there should definitely be an error instance accompanied by, but I'm worried about possible corner cases in the future)

I can append the actual error message if you think it's necessary!

johnplate (approved):
It works on Windows, thank you!

gapic/src/main/com/google/gapid/views/PerformanceView.java
pmuetschard: I think the capture error is handled that way, because the error is displayed somewhere else, because the capture model isn't really this view's primary model. However, the profile is this view's primary model, so we should show the error from the model, rather than a generic message.
Assuming that the model is loaded properly (i.e. not null) if the event is fired and `error` is null, is fine. We do that all over the place.

stellama0208: Great! Updated the PR to do simply the error check, and reflect the original error message on UI, thanks for explaining!

gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -101,6 +101,10 @@ public void onCommandsLoaded() {
 
   @Override
   public void onProfileLoaded(Loadable.Message error) {
+    if (error != null) {
+      loading.showMessage(error.type, error.text);
```
pmuetschard, line 105:
```suggestion
      loading.showMessage(error);
```

pmuetschard (approved):

PR #474, opened 2020-09-24
Avoid recording invalid command buffer

hevrard:
Some of the command buffers contained in the initial state cannot be
recorded due to some lacking resources. This fix adds a check to avoid
recording a command buffer which has a vkCmdExecuteCommands command
with as an argument a list of secondary command buffers where one of
the secondary command buffer recording has not been completed.

Also, fix a variable name typo: use `cb` to refer to a command
buffer (`qp` probably came from a copy-paste of the loop on QueryPools
just above).

Bug: b/161071991

Test: manual, with a trace that had a vkCmdExecuteCommands with a
secondary command buffer whose recording was not completed.

### DIFF
--- a/gapis/api/vulkan/command_buffer_rebuilder.go
+++ b/gapis/api/vulkan/command_buffer_rebuilder.go
@@ -921,6 +921,9 @@ func rebuildVkCmdExecuteCommands(
 		if !GetState(s).CommandBuffers().Contains(buf) {
 			return nil, nil, fmt.Errorf("Cannot find CommandBuffer %v", buf)
 		}
+		if GetState(s).CommandBuffers().Get(buf).Recording() != RecordingState_COMPLETED {
+			return nil, nil, fmt.Errorf("vkCmdExecuteCommands: secondary command buffer %v has not completed its recording", buf)
+		}
 	}
 
 	commandBufferData, commandBufferCount := unpackMap(ctx, s, d.CommandBuffers())
--- a/gapis/api/vulkan/state_rebuilder.go
+++ b/gapis/api/vulkan/state_rebuilder.go
@@ -268,14 +268,14 @@ func (API) RebuildState(ctx context.Context, oldState *api.GlobalState) ([]api.C
 		sb.createQueryPool(s.QueryPools().Get(qp))
 	}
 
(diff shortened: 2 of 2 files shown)


gapis/api/vulkan/command_buffer_rebuilder.go
```diff
@@ -921,6 +921,9 @@ func rebuildVkCmdExecuteCommands(
 		if !GetState(s).CommandBuffers().Contains(buf) {
 			return nil, nil, fmt.Errorf("Cannot find CommandBuffer %v", buf)
 		}
+		if GetState(s).CommandBuffers().Get(buf).Recording() != RecordingState_COMPLETED {
```
hevrard, line 924: To reviewers: I guess a question is: is it OK to keep on generating the replay if we detect such an issue, or should we abort the replay entirely?

In the test trace I used to debug this (from a Unity demo), the replay works fine with this fix.

hevrard: After a chat with @AWoloszyn , he agreed this is a relevant fix: it's OK to not recreate the primary commandbuffer, which would be invalid to submit anyway.

purvisa-at-google-com (approved):
This is better than a crash or undefined behaviour, but I wonder if the "true" fix is to reorder the recording of command buffers to prevent this dependency problem?

hevrard:
>  I wonder if the "true" fix is to reorder the recording of command buffers to prevent this dependency problem?

I had similar thoughts while debugging, so the commit message may need a bit of clarification: the secondary CB is in NOT_STARTED state in the trace's initial state, hence we won't even bother recreating it, ever.

The CB re-creation ordering is already relevant: we first re-create all secondary CB, then primary ones, see :
https://github.com/google/agi/blob/a76b8df2dc133c46ea8ce40f9fc868b79d469161/gapis/api/vulkan/state_rebuilder.go#L271-L279

And recordCommandBuffer() has an early return in the case the CB is NOT_STARTED, see:

https://github.com/google/agi/blob/a76b8df2dc133c46ea8ce40f9fc868b79d469161/gapis/api/vulkan/state_rebuilder.go#L3020-L3021

hevrard:
Merged with a clarified commit message.

PR #481, opened 2020-10-02
Don't store subslices in the database. 

pmuetschard:
Instead of storing the subslices in the database, store a function that can compute the data instead. This reduces memory usage a lot with very minimal added CPU performance cost.


AWoloszyn (approved):
Looks reasonable to me

PR #483, opened 2020-10-03
Add GPU counter confidence range feature in perftab.

stellama0208:
- When we get raw profile data from perfetto, there's an uncertainty
 in how GPU counter samples should be attributed to GPU slices,
 diffrent attribution strategy will lead to different performance
 aggregation result. Now in AGI, in addition to the previous best-guessing
 performance number, we have another confidence range showing what
 the performance will be in other two extreme scenarios: when minimal or
 maximal counter samples are attributed to a renderpass.
 - Bug: b/158055962.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Profile.java
+++ b/gapic/src/main/com/google/gapid/models/Profile.java
@@ -205,7 +205,7 @@ public class Profile
     public Service.ProfilingData.GpuCounters.Perf getGpuPerformance(List<Long> commandIndex, int metricId) {
       CommandIndex indexStr = new CommandIndex(commandIndex);
       Map<Integer, Service.ProfilingData.GpuCounters.Perf> perfs = perfLookup.get(indexStr);
-      return (perfs == null) ? null : perfs.getOrDefault(metricId, null);
+      return (perfs == null) ? null : perfs.get(metricId);
     }
 
     public Duration getDuration(Path.Commands range) {
--- a/gapic/src/main/com/google/gapid/models/Profile.java
+++ b/gapic/src/main/com/google/gapid/models/Profile.java
@@ -135,7 +135,7 @@ public class Profile
     public final Service.ProfilingData profile;
     private final Map<Integer, List<TimeSpan>> spansByGroup;
     private final List<Service.ProfilingData.GpuSlices.Group> groups;
-    private final Map<CommandIndex, Map<Integer, Double>> perfLookup; // commandIndex -> {metricId -> performanceValue}
+    private final Map<CommandIndex, Map<Integer, Service.ProfilingData.GpuCounters.Perf>> perfLookup; // commandIndex -> {metricId -> performanceValue}
 
     public Data(Path.Device device, Service.ProfilingData profile) {
       super(device);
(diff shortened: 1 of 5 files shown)


gapis/trace/android/profile/profile.go
```diff
@@ -134,77 +143,117 @@ func setGpuCounterMetrics(ctx context.Context, groupToSlices map[int32][]*servic
...
-				weightedSum += float64(ct) * vs[i]
-			} else if cStart < sEnd { // Sample wraps GPU slice's span, or sample's earlier part overlaps with slice.
-				ct := sEnd - u64.Max(sStart, cStart)
-				ctSum += ct
-				weightedSum += float64(ct) * vs[i]
+			} else if cStart > sEnd { // Sample later than GPU slice's span.
 				break
+			} else if cStart > sStart && cEnd < sEnd { // Sample is contained inside GPU slice's span.
+				estimateSet[i] = 1 * concurrencyWeight
+				minSet[i] = 1 * concurrencyWeight
```
pmuetschard, line 210: I think the minimum should only include this value, if there are no concurrent slices. If there are concurrent slices, there is a possibility that the value is entirely attributable to one of the other concurrent slices.

stellama0208:
The above force push solves merge conflict (related to #503 and #517), other than that there's no functional changes.

gapic/src/main/com/google/gapid/models/Profile.java
```diff
@@ -202,10 +202,10 @@ public TimeSpan getSlicesTimeSpan() {
       return new TimeSpan(start, end);
     }
 
-    public Double getGpuPerformance(List<Long> commandIndex, int metricId) {
+    public Service.ProfilingData.GpuCounters.Perf getGpuPerformance(List<Long> commandIndex, int metricId) {
       CommandIndex indexStr = new CommandIndex(commandIndex);
       if (!perfLookup.containsKey(indexStr)) {
-        return Double.NaN;
+        return null;
```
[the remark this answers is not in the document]
stellama0208: Thanks for reminding! The NPE resulted from a second consecutive map query is also done for this PR now.

gapis/trace/android/profile/profile.go
stellama0208: Good catch! Done in a new commit.

stellama0208:
Addressed the above comments, ready for another look, thanks!

gapic/src/main/com/google/gapid/models/Profile.java
```diff
@@ -202,10 +202,10 @@ public TimeSpan getSlicesTimeSpan() {
       return new TimeSpan(start, end);
     }
 
-    public Double getGpuPerformance(List<Long> commandIndex, int metricId) {
+    public Service.ProfilingData.GpuCounters.Perf getGpuPerformance(List<Long> commandIndex, int metricId) {
       CommandIndex indexStr = new CommandIndex(commandIndex);
-      Map<Integer, Double> metrics = perfLookup.get(indexStr);
-      return (metrics == null) ? Double.NaN : metrics.getOrDefault(metricId, Double.NaN);
+      Map<Integer, Service.ProfilingData.GpuCounters.Perf> perfs = perfLookup.get(indexStr);
+      return (perfs == null) ? null : perfs.getOrDefault(metricId, null);
```
pmuetschard, line 208: `null` is already the default, so using `getOrDefault` here only adds additional overhead if `null` is returned.

pmuetschard (approved):

gapic/src/main/com/google/gapid/models/Profile.java
stellama0208: Indeed! Updated accordingly.

PR #485, opened 2020-10-05
Disable commands in secondary command buffers

yalcinmelihyasin:
Support for disabling commands in secondary command buffers and disabling VkCmdExecuteCommands

b/169673918

### DIFF
--- a/gapis/api/vulkan/transform_command_disabler.go
+++ b/gapis/api/vulkan/transform_command_disabler.go
@@ -58,10 +58,6 @@ func (disablerTransform *commandDisabler) remove(ctx context.Context, id api.Sub
 		return fmt.Errorf("Requested id is empty")
 	}
 
-	if len(id) > 4 {
-		return fmt.Errorf("Drawcall removal from secondary command buffer is not implemented")
-	}
-
 	id[0] = id[0] + disablerTransform.cmdsOffset
 
 	disablerTransform.disabledCommands = append(disablerTransform.disabledCommands,
@@ -147,7 +143,8 @@ func (disablerTransform *commandDisabler) TransformCommand(ctx context.Context, 
 	return nil, nil
 }
 
-func (disablerTransform *commandDisabler) removeCommandFromVkQueueSubmit(ctx context.Context, idx api.SubCmdIdx, cmd *VkQueueSubmit, inputState *api.GlobalState) error {
+func (disablerTransform *commandDisabler) removeCommandFromVkQueueSubmit(ctx context.Context,
+	idx api.SubCmdIdx, cmd *VkQueueSubmit, inputState *api.GlobalState) error {
 	layout := inputState.MemoryLayout
 	cb := CommandBuilder{Thread: cmd.Thread(), Arena: inputState.Arena}
 	cmd.Extras().Observations().ApplyReads(inputState.Memory.ApplicationPool())
@@ -193,7 +190,8 @@ func (disablerTransform *commandDisabler) removeCommandFromVkQueueSubmit(ctx con
 	return nil
 }
 
-func (disablerTransform *commandDisabler) removeCommandFromSubmit(ctx context.Context, idx api.SubCmdIdx, submitInfo VkSubmitInfo, cmd *VkQueueSubmit, inputState *api.GlobalState) (VkSubmitInfo, error) {
+func (disablerTransform *commandDisabler) removeCommandFromSubmit(ctx context.Context,
+	idx api.SubCmdIdx, submitInfo VkSubmitInfo, cmd *VkQueueSubmit, inputState *api.GlobalState) (VkSubmitInfo, error) {
 	layout := inputState.MemoryLayout
 	// pCommandBuffers
 	commandBuffers := submitInfo.PCommandBuffers().Slice(0, uint64(submitInfo.CommandBufferCount()), layout).MustRead(ctx, cmd, inputState, nil)
@@ -235,11 +233,12 @@ func (disablerTransform *commandDisabler) removeCommandFromSubmit(ctx context.Co
 	return newSubmitInfo, nil
 }
 
-func (disablerTransform *commandDisabler) rewriteCommandBuffer(ctx context.Context, idx api.SubCmdIdx, existingCommandBuffer CommandBufferObjectʳ, cmd *VkQueueSubmit, inputState *api.GlobalState) (VkCommandBuffer, error) {
+func (disablerTransform *commandDisabler) rewriteCommandBuffer(ctx context.Context, idx api.SubCmdIdx,
+	existingCommandBuffer CommandBufferObjectʳ, cmd *VkQueueSubmit, inputState *api.GlobalState) (VkCommandBuffer, error) {
 	cb := CommandBuilder{Thread: cmd.Thread(), Arena: inputState.Arena}
 	st := GetState(inputState)
 
-	newCmdBuffer, err := disablerTransform.getNewCommandBufferAndBegin(ctx, cmd, inputState)
+	newCmdBuffer, err := disablerTransform.getNewCommandBufferAndBegin(ctx, existingCommandBuffer, cmd, inputState)
 	if err != nil {
 		log.E(ctx, "Command buffer could not be created : %v", err)
 		return VkCommandBuffer(0), err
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/transform_command_disabler.go
```diff
@@ -314,10 +370,10 @@ func (disablerTransform *commandDisabler) getNewCommandBufferAndBegin(ctx contex
 	}
 
 	beginInfo := NewVkCommandBufferBeginInfo(inputState.Arena,
-		VkStructureType_VK_STRUCTURE_TYPE_COMMAND_BUFFER_BEGIN_INFO, // sType
-		NewVoidᶜᵖ(memory.Nullptr),                                   // pNext
-		0,                                                           // flags
-		NewVkCommandBufferInheritanceInfoᶜᵖ(memory.Nullptr), // pInheritanceInfo
+		VkStructureType_VK_STRUCTURE_TYPE_COMMAND_BUFFER_BEGIN_INFO,                                         // sType
+		NewVoidᶜᵖ(memory.Nullptr),                                                                           // pNext
+		VkCommandBufferUsageFlags(VkCommandBufferUsageFlagBits_VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT), // flags
```
yalcinmelihyasin, line 375: @AWoloszyn @pmuetschard Does this affect performance characteristics?

AWoloszyn: Yes it can. PROBABLY not gpu performance, but it  maybe.

gapis/api/vulkan/transform_command_disabler.go
```diff
@@ -251,25 +250,48 @@ func (disablerTransform *commandDisabler) rewriteCommandBuffer(ctx context.Conte
...
+			executeCommandsArgs, ok := args.(VkCmdExecuteCommandsArgsʳ)
+			if !ok {
+				return VkCommandBuffer(0), fmt.Errorf("VkExecuteCommands could not found %v: ", args)
+			}
+
+			newCmd, err := disablerTransform.rewriteExecuteSecondaryCommandBuffer(ctx,
+				currentSubCmdID, newCmdBuffer, executeCommandsArgs, cmd, inputState)
 			if err != nil {
-				panic(fmt.Errorf("Invalid command-buffer detected %+v", err))
+				return VkCommandBuffer(0), fmt.Errorf("Error during rewriting secondary command buffer")
```
hevrard, line 275: Here you're losing the info of `err`, you could propagate the error up:

```suggestion
				return VkCommandBuffer(0), err
```

or

```suggestion
				return VkCommandBuffer(0), log.Err(ctx, err, "Cannot rewrite secondary command buffer")
```

gapis/api/vulkan/transform_command_disabler.go
```diff
@@ -251,25 +250,48 @@ func (disablerTransform *commandDisabler) rewriteCommandBuffer(ctx context.Conte
...
+			log.I(ctx, "Secondary Command Buffer %v updated", currentSubCmdID)
 			continue
 		}
 
-		if !isCmdAllowedToDisable(args) {
-			return VkCommandBuffer(0), fmt.Errorf("Command type is not allowed to be disabled : %v", args)
+		// Copy the unaffected commands to the new command buffer
+		cleanup, newCmd, err := AddCommand(ctx, cb, newCmdBuffer, inputState, inputState, args)
+		if err != nil {
+			panic(fmt.Errorf("Invalid command-buffer detected %+v", err))
```
hevrard, line 288: This panic message mentions an invalid command buffer, but should it rather say "Cannot add command: %+v" ?

gapis/api/vulkan/transform_command_disabler.go
```diff
@@ -280,7 +302,41 @@ func (disablerTransform *commandDisabler) rewriteCommandBuffer(ctx context.Conte
...
+func (disablerTransform *commandDisabler) rewriteExecuteSecondaryCommandBuffer(ctx context.Context,
+	idx api.SubCmdIdx, primaryCommandBuffer VkCommandBuffer, args VkCmdExecuteCommandsArgsʳ,
+	cmd *VkQueueSubmit, inputState *api.GlobalState) (api.Cmd, error) {
+
+	existingCmdBufferCount := uint32(args.CommandBuffers().Len())
+	newCommandBuffers := make([]VkCommandBuffer, 0, existingCmdBufferCount)
+
+	for i := uint32(0); i < existingCmdBufferCount; i++ {
+		currentSubCmdID := append(idx, uint64(i))
+		existingCommandBuffer := args.CommandBuffers().Get(uint32(i))
```
hevrard, line 314: The type cast `uint32(i)` is probably not needed?

gapis/api/vulkan/transform_command_disabler.go
```diff
@@ -280,7 +302,41 @@ func (disablerTransform *commandDisabler) rewriteCommandBuffer(ctx context.Conte
...
+		if !disablerTransform.doesContainDisabledCmd(currentSubCmdID) {
+			newCommandBuffers = append(newCommandBuffers, existingCommandBuffer)
+			continue
+		}
+
+		existingCommandBufferObject := GetState(inputState).CommandBuffers().Get(existingCommandBuffer)
+		newCommandBuffer, err := disablerTransform.rewriteCommandBuffer(ctx, currentSubCmdID, existingCommandBufferObject, cmd, inputState)
+		if err != nil {
+			return nil, fmt.Errorf("Error during writing secondary command buffer : %v", err)
```
hevrard, line 323: When there is a context and you want to propagate an error up with an additional message, prefer using log.Err when possible, e.g.: `return nil, log.Err(ctx, err, "Failed to rewrite secondary command buffer")`

gapis/api/vulkan/transform_command_disabler.go
hevrard: Shouldn't this be `existingCommandBuffer.Flags()` ?

gapis/api/vulkan/transform_command_disabler.go
```diff
@@ -314,10 +370,10 @@ func (disablerTransform *commandDisabler) getNewCommandBufferAndBegin(ctx contex
...
 
 	beginInfo := NewVkCommandBufferBeginInfo(inputState.Arena,
-		VkStructureType_VK_STRUCTURE_TYPE_COMMAND_BUFFER_BEGIN_INFO, // sType
-		NewVoidᶜᵖ(memory.Nullptr),                                   // pNext
-		0,                                                           // flags
-		NewVkCommandBufferInheritanceInfoᶜᵖ(memory.Nullptr), // pInheritanceInfo
+		VkStructureType_VK_STRUCTURE_TYPE_COMMAND_BUFFER_BEGIN_INFO,                                         // sType
+		NewVoidᶜᵖ(memory.Nullptr),                                                                           // pNext
+		VkCommandBufferUsageFlags(VkCommandBufferUsageFlagBits_VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT), // flags
+		NewVkCommandBufferInheritanceInfoᶜᵖ(memory.Nullptr),                                                 // pInheritanceInfo
```
hevrard, line 376: For secondary command buffers, this pInheritanceInfo cannot be NULL. Shouldn't this be the inheritance info of `existingCommandBuffer` ?

hevrard (changes requested):
Main logic looks good.

gapis/api/vulkan/transform_command_disabler.go
yalcinmelihyasin: Good catch

gapis/api/vulkan/transform_command_disabler.go
yalcinmelihyasin: Makes sense.

gapis/api/vulkan/transform_command_disabler.go
yalcinmelihyasin: Yes this was the case I was mentioning in private chat. I am looking into this. This logic might be incorrect in multiple places outside of this transform.

yalcinmelihyasin: I have fixed this. Actually we have to check the other places that this is missing e.g. Vulkan terminator

hevrard (approved):

PR #486, opened 2020-10-06
Fix gapis crash when perfetto profile data is empty.

stellama0208:
- This bug is triggered when the returned perfetto profile is empty, and AGI
 tries to read the first byte in the buffer when creating a perfetto processor.
 - Instead of crashing AGI entirely, throw an error from server and mirror
 it in UI.
 - Bug: b/169426834.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/CommandTree.java
+++ b/gapic/src/main/com/google/gapid/views/CommandTree.java
@@ -245,10 +245,13 @@ public class CommandTree extends Composite
   }
 
   @Override
+  public void onProfileLoadingStart() {
+    tree.profileLoadingError = null;
+  }
+
+  @Override
   public void onProfileLoaded(Loadable.Message error) {
-    if (error != null) {
-      tree.profileLoadingError = error;
-    }
+    tree.profileLoadingError = error;
     tree.refresh();
   }
 
--- a/gapic/src/main/com/google/gapid/views/CommandTree.java
+++ b/gapic/src/main/com/google/gapid/views/CommandTree.java
@@ -246,6 +246,9 @@ public class CommandTree extends Composite
 
   @Override
   public void onProfileLoaded(Loadable.Message error) {
+    if (error != null) {
+      tree.profileLoadingError = error;
+    }
     tree.refresh();
   }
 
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/views/CommandTree.java
```diff
@@ -246,6 +246,9 @@ public void onCommandsSelected(CommandIndex index) {
 
   @Override
   public void onProfileLoaded(Loadable.Message error) {
+    if (error != null) {
+      tree.profileLoadingError = error;
+    }
```
hevrard, line 251: Here, once tree.profileLoadingError is set to an error, there is no way to turn it back to null. Would it make sense to just have:

```suggestion
    tree.profileLoadingError = error;
```

such that one can retry to profile and, if the second attempt to profile works, get rid of the error?

gapis/trace/android/trace.go
```diff
@@ -133,7 +133,9 @@ func (t *androidTracer) ProcessProfilingData(ctx context.Context, buffer *bytes.
 		return nil, log.Err(ctx, err, "Failed to read trace buffer")
 	}
 	processor, err := perfetto.NewProcessor(ctx, rawData)
-	defer processor.Close()
+	if processor != nil {
+		defer processor.Close()
+	}
```
pmuetschard, line 138: To simplify this, just have the `processor.Close` function handle a nil receiver object.

gapis/perfetto/processor.go
```diff
@@ -35,6 +35,11 @@ type Processor struct {
 func NewProcessor(ctx context.Context, data []byte) (*Processor, error) {
 	p := C.new_processor()
 	log.D(ctx, "[perfetto] Parsing %d bytes", len(data))
+	if len(data) == 0 {
+		log.W(ctx, "[perfetto] Empty profile data.")
+		C.delete_processor(p)
+		return nil, log.Errf(ctx, nil, "profile data is empty.")
+	}
```
pmuetschard, line 42: Put this check above the `C.new_processor()` call, so we can avoid allocating and de-allocating the processor in the error case.

gapic/src/main/com/google/gapid/views/CommandTree.java
stellama0208: Nice catch! Done with the new commit!

gapis/trace/android/trace.go
stellama0208: Thanks for suggestion, done! (Finally I see a nice thing with golang :))

hevrard (approved):

pmuetschard (approved):

PR #488, opened 2020-10-06
Create a "gapit commands" option that only shows draw calls

hliatis:
This option creates a command tree of just draw calls and their parent queue submits. Needed to create a screenshot swarming test. Gapit also filters out the queue submits to only show the draw calls. 

Bug: b/161536617

### DIFF
--- a/cmd/gapit/commands.go
+++ b/cmd/gapit/commands.go
@@ -59,6 +59,7 @@ func (verb *commandsVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 	if err != nil {
 		return log.Err(ctx, err, "Failed to build the CommandFilter")
 	}
+	filter.OnlyExecutedDraws = verb.OnlyExecutedDraws
 
 	treePath := capture.CommandTree(filter)
 	treePath.GroupByApi = verb.GroupByAPI
@@ -102,10 +103,16 @@ func (verb *commandsVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 	}
 
 	return traverseCommandTree(ctx, client, tree.Root, func(n *service.CommandTreeNode, prefix string) error {
-		fmt.Fprintf(os.Stdout, prefix)
-		if n.Group != "" {
-			fmt.Fprintln(os.Stdout, n.Group)
-			return nil
+		if verb.OnlyExecutedDraws {
+			if n.Group != "" || n.NumChildren > 0 {
+				return nil
+			}
+		} else {
+			fmt.Fprintf(os.Stdout, prefix)
+			if n.Group != "" {
+				fmt.Fprintln(os.Stdout, n.Group)
+				return nil
+			}
 		}
 		return getAndPrintCommand(ctx, client, n.Commands.First(), verb.Observations)
 	}, "", true)
--- a/cmd/gapit/flags.go
+++ b/cmd/gapit/flags.go
@@ -254,6 +254,7 @@ type (
 		GroupByUserMarkers   bool   `help:"Group commands by user markers"`
 		GroupBySubmission    bool   `help:"Group commands by submissions"`
 		AllowIncompleteFrame bool   `help:"_Make a group for incomplete frames"`
+		OnlyExecutedDraws    bool   `help:"Only show executed draw calls from within command buffers"`
 		Observations         ObservationFlags
 		CommandFilterFlags
 		CaptureFileFlags
--- a/gapis/api/cmd_id_group.go
+++ b/gapis/api/cmd_id_group.go
@@ -438,6 +438,21 @@ func (c *SubCmdRoot) Insert(r []uint64, nameLookUp *SubCmdIdxTrie) {
 	}
 }
 
+func (c *SubCmdRoot) InsertWithFilter(r []uint64, nameLookUp *SubCmdIdxTrie, filter func(CmdID) bool) {
+	childRoot := c.newChildSubCmdRoots(r[0:len(r)-1], nameLookUp)
+	// Add subcommands one-by-one to the SubCmdRoot and its subgroups/child
+	// SubCmdRoots
+	id := r[len(r)-1]
+	if CmdID(id) > childRoot.SubGroup.Range.End {
+		childRoot.SubGroup.Range.End = CmdID(id + 1)
+	}
+	for i := CmdID(0); i < CmdID(id); i++ {
+		if filter(i) {
+			childRoot.SubGroup.AddCommand(i)
+		}
+	}
+}
+
 // AddSubCmdMarkerGroups adds the given groups to the target SubCmdRoot
 // with the relative hierarchy specified in r. If the groups are not added as
 // immediate children of the target SubCmdRoot (r is not empty), child
--- a/gapis/resolve/command_tree.go
+++ b/gapis/resolve/command_tree.go
@@ -232,7 +232,7 @@ func (r *CommandTreeResolvable) Resolve(ctx context.Context) (interface{}, error
 		if err := cmd.Mutate(ctx, id, s, nil, nil); err != nil {
 			return fmt.Errorf("Fail to mutate command %v: %v", cmd, err)
 		}
-		if filter(id, cmd, s) {
+		if filter(id, cmd, s, api.SubCmdIdx([]uint64{uint64(id)})) {
 			for _, g := range groupers {
 				g.Process(ctx, id, cmd, s)
 			}
@@ -299,12 +299,12 @@ func (r *CommandTreeResolvable) Resolve(ctx context.Context) (interface{}, error
 			return fmt.Errorf("Fail to mutate command %v: %v", cmd, err)
 		}
 
-		if !filter(id, cmd, s) {
+		if !filter(id, cmd, s, api.SubCmdIdx([]uint64{uint64(id)})) {
 			return nil
 		}
 
 		if v, ok := snc.SubcommandGroups[id]; ok {
-			r := out.root.AddRoot([]uint64{uint64(id)}, snc.SubcommandNames)
+			subr := out.root.AddRoot([]uint64{uint64(id)}, snc.SubcommandNames)
 			// subcommands are added before nesting SubCmdRoots.
 			cv := append([]api.SubCmdIdx{}, v...)
 			sort.SliceStable(cv, func(i, j int) bool { return len(cv[i]) < len(cv[j]) })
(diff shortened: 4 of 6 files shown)


gapis/service/path/path.proto
```diff
@@ -327,12 +327,14 @@ message CommandTree {
...
+  // If true, only shows draw calls from inside a command buffer
+  bool only_executed_draws = 13;
   // If positive, synthetic sub-nodes are created for nodes with more than
   // this many children.
-  int32 max_children = 13;
+  int32 max_children = 14;
   // If positive, synthetic sub-nodes are created for long spans of commands
   // between groups. This ensures the groups do not get lost in the noise.
-  int32 max_neighbours = 14;
+  int32 max_neighbours = 15;
```
hevrard, line 337: protobuf IDs are meant to never be changed once they have been used. So here `only_executed_draws` needs to be a new ID (16 i guess?), and leave the other IDs untouched.

hevrard (commented):
Presubmit is failing.

test/swarming/bot-scripts/video.py
```diff
@@ -84,6 +86,45 @@ def main():
...
+    ]
+    p = botutil.runcapture(cmd)
+
+    if p.returncode != 0:
+        return 1
+
+    commands = []
+    for line in p.stdout.splitlines():
+        if line.startswith('['):
+            commands.append(re.search(r'^\[.*\]', line).group(0).replace(' ', ', '))
```
hevrard, line 109: Anytime you use regexps, I highly encourage to document what they are expected to match:

```suggestion
        # Lines with draw calls commands looks like: "[1 2 3 4] vkCmdDraw..."
        # Here we select those lines, extract the command ID and add commas
        # between the indexes to get: "[1, 2, 3, 4]"
        if line.startswith('['):
            commands.append(re.search(r'^\[.*\]', line).group(0).replace(' ', ', '))
```

test/swarming/bot-scripts/botutil.py
```diff
@@ -32,6 +32,10 @@ def runcmd(cmd):
     '''Run a command, redirecting output to the system stdout and stderr.'''
     return subprocess.run(cmd, stdout=sys.stdout, stderr=sys.stderr)
 
+def runcapture(cmd):
+    '''Run a command, capturing output.'''
+    return subprocess.run(cmd, capture_output=True, text=True)
```
hevrard, line 37: I would prefer to rename to `runSaveOutput`, because "capture" already means something on AGI lingo.

hevrard (commented):
We need to not be random in the draw calls we select.

Also, the regexp approach is brittle, as we discussed. Would it be possible to bake this directly into a `gapit screenshot` flag, e.g. `gapit screenshot -at-mid-frame-draws 5` which would compute the list of draw call indexes and then request the screenshots at these indexes, without the need for regexps matching on some text interface with regexpes?

test/swarming/bot-scripts/botutil.py
hliatis: I think I will just modify runcmd() with a default named parameter save_output set to false

gapis/service/path/path.proto
```diff
@@ -333,6 +333,8 @@ message CommandTree {
   // If positive, synthetic sub-nodes are created for long spans of commands
   // between groups. This ensures the groups do not get lost in the noise.
   int32 max_neighbours = 14;
+  // If true, only shows draw calls from inside a command buffer
+  bool only_executed_draws = 15;
```
pmuetschard, line 337: This should go into the `CommandFilter` struct above.

cmd/gapit/commands.go
```diff
@@ -102,10 +103,16 @@ func (verb *commandsVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 	}
 
 	return traverseCommandTree(ctx, client, tree.Root, func(n *service.CommandTreeNode, prefix string) error {
-		fmt.Fprintf(os.Stdout, prefix)
-		if n.Group != "" {
-			fmt.Fprintln(os.Stdout, n.Group)
-			return nil
+		if !verb.OnlyExecutedDraws {
```
pmuetschard, line 106: Please remove the `!` from the condition and swap the then and else clauses.

gapis/service/path/path.proto
hliatis: I need to look into how much refactoring is needed to make filters useful. The way it works now, the filter would pass for recording draw calls (Host Coordination), but failt for VkQueueSubmits that house the actual draw calls we want.

gapis/resolve/command_tree.go
```diff
@@ -313,22 +313,35 @@ func (r *CommandTreeResolvable) Resolve(ctx context.Context) (interface{}, error
...
+						return filter(id, subCmd, s, api.SubCmdIdx(append(parentIdx, uint64(id))))
+					}
+
+					return false
+				})
 			}
 			return nil
 		}
 
+		//if !p.OnlyExecutedDraws {
```
hevrard, line 336: remove commented code

gapis/resolve/command_tree.go
```diff
@@ -313,22 +313,35 @@ func (r *CommandTreeResolvable) Resolve(ctx context.Context) (interface{}, error
...
 		out.root.AddCommand(id)
 
 		if flags := cmd.CmdFlags(); flags.IsDrawCall() || flags.IsClear() {
 			drawOrClearCmds = append(drawOrClearCmds, &api.CmdIDRange{
 				Start: id, End: id + 1,
 			})
 		}
+		//}
```
hevrard, line 344: to remove.

gapis/api/cmd_id_group.go
```diff
@@ -438,6 +438,21 @@ func (c *SubCmdRoot) Insert(r []uint64, nameLookUp *SubCmdIdxTrie) {
 	}
 }
 
+func (c *SubCmdRoot) InsertWithFilter(r []uint64, nameLookUp *SubCmdIdxTrie, f func(CmdID) bool) {
```
hevrard, line 441: It would be clearer to rename `f` by `filter` here.

cmd/gapit/flags.go
```diff
@@ -366,15 +366,16 @@ type (
...
-		Out        string           `help:"output image file (default 'screenshot.png')"`
-		NoOpt      bool             `help:"disables optimization of the replay stream"`
-		Attachment string           `help:"the attachment to show (0-3 for color, d for depth, s for stencil)"`
-		Overdraw   bool             `help:"renders the overdraw instead of the color framebuffer"`
+		Gapis          GapisFlags
+		Gapir          GapirFlags
+		At             []flags.U64Slice `help:"command/subcommand index (e.g. '[123, 0, 0, 4]') for the screenshot (repeatable)"`
+		Frame          []int            `help:"frame index for the screenshot (repeatable). Empty for last"`
+		Draws          bool             `help:"create a screenshot of every draw call in the requested frame(s) (only honored if using -frame)"`
+		ExecutedDraws  int				`help:"create a screenshot of evenly spaced executed draw calls. Number provided is the maximum amount. (not used with -frame)"`
```
hevrard, line 374: Broken indentation.

I think we may have a help text maybe a bit more explicit:

```
create a screenshot at every (total-draw-calls / this-flag-value) drawcall index in the frame, e.g. for a frame with 100 draw calls, '-executed-draws 4' takes a screenshot at every 25 draw calls.
```

cmd/gapit/screenshot.go
```diff
@@ -75,6 +76,11 @@ func (verb *screenshotVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 		for _, at := range verb.At {
 			commands = append(commands, capture.Command(at[0], at[1:]...))
 		}
+	} else if verb.ExecutedDraws > 0 {
+		commands, err = verb.executedDrawCommands(ctx, capture, client, verb.ExecutedDraws)
```
hevrard, line 80: Check that `verb.ExecutedDraws > 0`, return an error otherwise

cmd/gapit/screenshot.go
```diff
@@ -248,6 +254,45 @@ func (verb *screenshotVerb) frameCommands(ctx context.Context, capture *path.Cap
...
+		return nil
+	}, "", true)
+
+	if len(allDrawCommands) > maxAmount {
+		commands := make([]*path.Command, maxAmount)
+		step := float32(len(allDrawCommands) - 1) / float32(maxAmount - 1)
+	
+		for i := 0; i < len(commands); i++ {
+			commands[i] = allDrawCommands[int32(float32(i) * step)]
+		}
```
hevrard, line 288: Why the `-1`s  here? Also, this can lead to a division by zero if `maxAmount == 1`?

I think a simpler solution would be:

```
commands := []*path.Command{}
step := len(allDrawCommands) / maxAmount
if step == 0 {
  step = 1
}
for i := 0; i < len(allDrawCommands); i += step {
  commands = append(commands, allDrawCommands[i])
}
```

hevrard (commented):
The main logic looks good, but several things to cleanup.

cmd/gapit/screenshot.go
hliatis: -1 helps with making sure our evenly spaced indices ends with the last index. Otherwise, it will stop early. For example, 30/10 will stop at 27, but 29/19 will stop at 29. It's not a big deal. I just felt that it should be spaced between the first and last item.

hliatis: So. I did a quick test of your method. If len(allDrawCommands) == 5 and maxAmount == 3, then step == 1. Meaning, we would end up appending 5 commands, not 3.

cmd/gapit/screenshot.go
hevrard: Woops, you do check this on the line above, I read too fast. We could return an error when verb.ExecutedDraws < 0, but it's OK to be silent here since this is a developer-oriented option.

test/swarming/bot-scripts/video.py
```diff
@@ -85,6 +85,18 @@ def main():
         gfxtrace
     ]
     p = botutil.runcmd(cmd)
+    if p.returncode != 0:
+        return p.returncode
+    
+    #### Screenshot test
```
hevrard, line 91: Be more explicit: `#### Screenshot test to retrieve mid-frame resources`

cmd/gapit/commands.go
```diff
@@ -102,10 +103,17 @@ func (verb *commandsVerb) Run(ctx context.Context, flags flag.FlagSet) error {
...
 	return traverseCommandTree(ctx, client, tree.Root, func(n *service.CommandTreeNode, prefix string) error {
-		fmt.Fprintf(os.Stdout, prefix)
-		if n.Group != "" {
-			fmt.Fprintln(os.Stdout, n.Group)
-			return nil
+		if verb.OnlyExecutedDraws {
+			if n.Group != "" || n.NumChildren > 0 {
+				return nil
+			}
+
```
hevrard, line 110: remove empty line here.

cmd/gapit/screenshot.go
hevrard: As discussed offline, let's keep the float step approach but remove the `-1`'s.

cmd/gapit/screenshot.go
```diff
@@ -248,6 +254,46 @@ func (verb *screenshotVerb) frameCommands(ctx context.Context, capture *path.Cap
...
+		if n.Group != "" || n.NumChildren > 0 {
+			return nil
+		}
+		allDrawCommands = append(allDrawCommands, n.Commands.First())
+		return nil
+	}, "", true)
+
+	if len(allDrawCommands) > maxAmount {
+		commands := make([]*path.Command, maxAmount)
+		// We use a float step so that we have an fairly even distribution when ration < 2 (e.g. 5/3)
```
hevrard, line 284: typo: `an fairly` -> `a fairly`
typo: `ration` -> `ratio`

hevrard (approved):

PR #491, opened 2020-10-09
Align performance tab to command tree tab.

stellama0208:
- Add two features that's already in command tab (link with profile tab and
 resource loading at selection), in this way we could reduce the need of
 switching between the experimental performance tab and command tab.
 - Bug: b/169854984.


hevrard (approved):
LGTM. I tested it and it works like a charm!

PR #495, opened 2020-10-12
Adds the capability of disabling multiple commands

yalcinmelihyasin:
for command disabler transform

### DIFF
--- a/gapis/api/vulkan/transform_command_disabler.go
+++ b/gapis/api/vulkan/transform_command_disabler.go
@@ -485,6 +485,19 @@ func (disablerTransform *commandDisabler) observeAndWriteCommand(cmd api.Cmd) er
 }
 
 func (disablerTransform *commandDisabler) sortAndMergeDisabledCmds(ctx context.Context) {
+	/*
+		This functions ensures these conditions:
+			- All disabled commands are sorted so we can guarantee that any command to be
+			disabled is not touched by command disabler, therefore can be assumed to have
+			correct id.
+
+			- Ensures for this input set:
+			"[3893.0.6.3.0.13], [3893.0.6.3.0.17], [3893.0.6.3.0.22], [3893.0.6.3]"
+			All the subcommands of [3893.0.6.3] are eliminated as they are not necessary.
+
+			- Removes all duplicates.
+	*/
+
 	api.SortSubCmdIDs(disablerTransform.disabledCommands)
 
 	size := len(disablerTransform.disabledCommands)
--- a/gapis/api/vulkan/transform_command_disabler.go
+++ b/gapis/api/vulkan/transform_command_disabler.go
@@ -486,13 +486,12 @@ func (disablerTransform *commandDisabler) observeAndWriteCommand(cmd api.Cmd) er
 
 func (disablerTransform *commandDisabler) sortAndMergeDisabledCmds(ctx context.Context) {
 	/*
-		This functions ensures these conditions:
-			- All disabled commands are sorted so we can guarantee that any command to be
-			disabled is not touched by command disabler, therefore can be assumed to have
-			correct id.
+		This function ensures these conditions:
+			- To guarantee the command indices to be valid, all operations has to be done to the
+			original command buffers in order and one pass. In that way, we can guarantee
+			that the command order in command buffers are untouched and as it is in the trace.
 
-			- Ensures for this input set:
-			"[3893.0.6.3.0.13], [3893.0.6.3.0.17], [3893.0.6.3.0.22], [3893.0.6.3]"
+			- For this input set: "[3893.0.6.3.0.13], [3893.0.6.3.0.17], [3893.0.6.3.0.22], [3893.0.6.3]"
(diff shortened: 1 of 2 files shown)


hevrard (commented):
Main logic looks good, but I think we may simplify a piece.

gapis/api/vulkan/transform_command_disabler.go
```diff
@@ -476,6 +483,25 @@ func (disablerTransform *commandDisabler) observeAndWriteCommand(cmd api.Cmd) er
...
+
+	// Remove the duplicates and clean the subcommands of deleted subcommands
+	for i := int64(size - 2); i >= 0; i-- {
+		if ids[len(ids)-1].Contains(disablerTransform.disabledCommands[i]) {
+			continue
+		}
+		ids = append(ids, disablerTransform.disabledCommands[i])
+	}
+
+	api.ReverseSubCmdIDs(ids)
```
hevrard, line 501: I don't see why you loop in reverse order and then reverse the whole `ids` list? I think you can achieve the same while looping one, in order:

```suggestion
	// Remove the duplicates and clean the subcommands of deleted subcommands
	for _, cmd := range disablerTransform.disabledCommands {
		if len(ids) == 0 || !cmd.Contains(ids[len(ids)-1]) {
			ids = append(ids, cmd)
		} else {
			// Here cmd does contain the last element of ids, hence
			// disabling cmd will effectively disable ids[len(ids)-1],
			// so we replace ids[len(ids)-1] by cmd
			ids[len(ids)-1] = cmd
		}
	}
```

What do you think?

yalcinmelihyasin: I think you have a bit missing information here. The intention is: 

For this input list:

`[3893,0,6,3,0,22], [3893,0,6,3],[3893,0,6,3,0,13], [3893,0,6,3,0,28], [3893,0,11,19], [3893,0,6,3,0,17]`

sort them into:

`[3893,0,6,3,0,13],  [3893,0,6,3,0,17], [3893,0,6,3,0,22], [3893,0,6,3,0,28], [3893,0,6,3], [3893,0,11,19],`

then later deduct the unnecessary ones:

`[3893,0,6,3], [3893,0,11,19],`

If you run your code by hand, you can see that your output is different.

The problem is: the code you suggest only checks if the last command is contained by the current one but in our case there are multiple command has been contained by one.

I think any further iteration after here to eliminate reversing will make the code more complicated rather than simple. Reversing will not have any performance impact as we expect a small set of input commands for the usual case and semantically simpler than any trick to remove duplicates or deducting the subcommands of a disabled command.

hevrard: OK, thanks for the clarification. We could avoid the list reversal by sorting appropriately, e.g. in a way that leads to:

`[3893,0,6,3], [3893,0,6,3,0,13], [3893,0,6,3,0,17], [3893,0,6,3,0,22], [3893,0,6,3,0,28], [3893,0,11,19]`

such that we can then cleanup with this code:
```
	for _, cmd := range disablerTransform.disabledCommands {
		if len(ids) == 0 || !ids[len(ids)-1].Contains() {
			ids = append(ids, cmd)
		}
	}
```

How hard would it be to get the sorting as above?

If you just want to keep your current solution, I think you need to comment what the sorting leads to, so the removal procedure makes sense, e.g.:

```
// sorting leads to container commands being after their containee commands, e.g.:
// [1, 0, 2, 3], [1, 0, 2, 4], [1, 0, 2], ...
```

yalcinmelihyasin: That would require the change all the code that uses the sort logic which I don't know how much is used and how much is necessary.

I think a comment would be enough

gapis/api/vulkan/transform_command_disabler.go
```diff
@@ -476,35 +484,70 @@ func (disablerTransform *commandDisabler) observeAndWriteCommand(cmd api.Cmd) er
...
+		This functions ensures these conditions:
+			- All disabled commands are sorted so we can guarantee that any command to be
+			disabled is not touched by command disabler, therefore can be assumed to have
+			correct id.
+
+			- Ensures for this input set:
+			"[3893.0.6.3.0.13], [3893.0.6.3.0.17], [3893.0.6.3.0.22], [3893.0.6.3]"
+			All the subcommands of [3893.0.6.3] are eliminated as they are not necessary.
+
+			- Removes all duplicates.
```
hevrard, line 498: typo: This function (no "s") ensures ...

Suggested rephrasing:
```
- To-be-disabled commands are first sorted and duplicates are removed
  such that the actual command disabling can happen in a single sweep
  of the original command buffer, with the relevant command IDs.

- In the list of commands to be disabled, commands that are subcommands of others are removed.
  E.g., "[1.0.2.3], [1.0.2.4], [1.0.2]" is cleaned up to "[1.0.2]", since disabling this command also disables
  its subcommands.
```

hevrard (approved):
LGMT, just proposal to clarify the comments.

PR #497, opened 2020-10-13
Preload dependency graph only on graphics captures

hevrard:
This avoids an error message when dealing with perfetto traces.

Bug: b/169062298


johnplate (approved):
LGTM

PR #499, opened 2020-10-14
Fix the crash due to fast interaction with command tree

yalcinmelihyasin:
b/170321382

### DIFF
--- a/gapic/src/main/com/google/gapid/widgets/LinkifiedTree.java
+++ b/gapic/src/main/com/google/gapid/widgets/LinkifiedTree.java
@@ -276,7 +276,7 @@ public abstract class LinkifiedTree<T, F> extends Composite {
 
     @Override
     public Object[] getChildren(Object parent) {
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/widgets/LinkifiedTree.java
```diff
@@ -276,7 +276,7 @@ protected T getElement(Widget item) {
 
     @Override
     public Object[] getChildren(Object parent) {
-      return getChildNodes(cast(parent));
+      return !hasChildren(parent) ? null : getChildNodes(cast(parent));
```
hevrard, line 279: I would avoid the negative test (!):

```suggestion
      return hasChildren(parent) ? getChildNodes(cast(parent)) : null;
```

hevrard (approved):
The fix sounds good, but I would avoid negative conditionals when possible.

PR #500, opened 2020-10-15
Fix the Windows build

pmuetschard:
1) Go back to getting packages from the main MSYS repo, except...
2) ...for patch, get it off of a separate mirror that still hosts the old version of patch.

The newer version of patch is built with a different MSYS.dll, and so fails to run on our old version of MSYS.

### DIFF
--- a/kokoro/windows/build.bat
+++ b/kokoro/windows/build.bat
@@ -33,6 +33,7 @@ REM Install Android SDK platform, build tools and NDK
 setlocal
 call %ANDROID_HOME%\tools\bin\sdkmanager.bat platforms;android-26 build-tools;29.0.2
 endlocal
+echo on
 
 wget -q https://dl.google.com/android/repository/android-ndk-r21d-windows-x86_64.zip
 unzip -q android-ndk-r21d-windows-x86_64.zip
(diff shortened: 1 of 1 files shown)


kokoro/windows/build.bat
```diff
@@ -33,6 +33,7 @@ REM Install Android SDK platform, build tools and NDK
 setlocal
 call %ANDROID_HOME%\tools\bin\sdkmanager.bat platforms;android-26 build-tools;29.0.2
 endlocal
+echo on
```
hevrard, line 36: Maybe remove this before merging?

hevrard (approved):
Thanks for this fix!

kokoro/windows/build.bat
pmuetschard: no, why?

stellama0208 (approved):
Thanks for finding a working version! Windows build is now green for the PR!

kokoro/windows/build.bat
```diff
@@ -50,17 +51,17 @@ REM supports only ".pkg.tar.xz" package archives, it does not support
 REM the more recent ".pkg.tar.zst" packages archives.
 c:\tools\msys64\usr\bin\bash --login -c "pacman -R --noconfirm catgets libcatgets"
 REM Use an old version of patch known to work with the msys runtime
-REM version that comes on Kokoro.
-set REPO_MSYS2=https://sourceforge.net/projects/msys2/files/REPOS/MSYS2/x86_64
-set REPO_MINGW=https://sourceforge.net/projects/msys2/files/REPOS/MINGW/x86_64
```
stellama0208, line 55: nit: looks like we could still keep the variables here and only replace the link .

kokoro/windows/build.bat
hevrard: :+1:

PR #501, opened 2020-10-16
 Image conversion crash fix

yalcinmelihyasin:
This commit fixes the crash caused by the bug. And also
     - Prints more information when image conversion is failed
     - Renames the function that reads min and max value from Float src
     - It does reading in the conversion that needs it.

b/161683278

### DIFF
--- a/core/stream/convert.go
+++ b/core/stream/convert.go
@@ -87,13 +87,21 @@ func Convert(dst, src *Format, data []byte) ([]byte, error) {
 	// Some components can be implicitly added (alpha, Y, Z, W).
 	mappings = resolveImplicitMappings(count, mappings, src, data)
 
+	// Calculate min/max if floats
+	min, max := float64(math.MaxFloat64), -float64(math.MaxFloat64)
+	for _, m := range mappings {
+		if m.dst.component.DataType.IsInteger() && m.src.component.DataType.IsFloat() && !m.dst.component.DataType.Signed {
+			readMinMax(count, m.src, &min, &max)
+		}
+	}
+
 	// Do the conversion work.
 	for _, m := range mappings {
 		if m.src.component == nil {
 			return nil, fmt.Errorf("Channel %v not found in source format: %v",
 				m.dst.component.Channel, src)
 		}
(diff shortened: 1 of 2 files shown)


core/stream/convert.go
```diff
@@ -195,9 +188,10 @@ func (m *mapping) conv(count int, min, max float64) error {
 		if d.DataType.Signed {
 			return ftos(count, m.dst, m.src)
 		}
-		return ftou(count, m.dst, m.src, min, max)
+		return ftou(count, m.dst, m.src)
+	default:
+		return fmt.Errorf("Cannot convert from Unknown %v to Unknown %v", s, d)
```
yalcinmelihyasin, line 193: I guess this should never happen. Maybe we can change it with panic?

pmuetschard (changes requested):
The min/max should be calculated across all channels not channel-by-channel.

When fixing a bug, don't add additional changes in at the same time. Make them separate commits. It's hard to see what the actual bug fix is here due to the additional noisy changes. The changes inside the `ftouMinMax` function should be enough to fix the bug, no?

core/stream/convert.go
```diff
@@ -396,7 +398,7 @@ func writeUintClamped(bs *binary.BitStream, bits uint64, count uint32) {
 	bs.Write(u64.Min(bits, limit), count)
 }
 
-func ftouMinMax(count int, dst, src buf, min, max *float64) error {
+func readMinMax(count int, src buf, min, max *float64) error {
```
pmuetschard, line 401: why the rename? I don't see how the new name is better than the old.

hliatis: I guess because he got rid of dst?

yalcinmelihyasin: remove the dst as it is not used.

yalcinmelihyasin:
What is the requested change? I could not understand. Removing the other things except the crash fix?

yalcinmelihyasin:
Also, I have reverted the change in the logic for calculating min/max. Therefore the only change about it is changing the bits. Maybe you have missed that.

pmuetschard (approved):

PR #503, opened 2020-10-21
Fix a NPE in Performance tab when returning a Double object.

stellama0208:
Check data availability and return Double.NAN when possible, avoid
 returning null.

 Bug: N/A.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Profile.java
+++ b/gapic/src/main/com/google/gapid/models/Profile.java
@@ -204,10 +204,8 @@ public class Profile
 
     public Double getGpuPerformance(List<Long> commandIndex, int metricId) {
       CommandIndex indexStr = new CommandIndex(commandIndex);
-      if (!perfLookup.containsKey(indexStr)) {
-        return Double.NaN;
-      }
-      return perfLookup.get(indexStr).get(metricId);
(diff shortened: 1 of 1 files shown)


hevrard (approved):

gapic/src/main/com/google/gapid/models/Profile.java
```diff
@@ -204,7 +204,7 @@ public TimeSpan getSlicesTimeSpan() {
 
     public Double getGpuPerformance(List<Long> commandIndex, int metricId) {
       CommandIndex indexStr = new CommandIndex(commandIndex);
-      if (!perfLookup.containsKey(indexStr)) {
+      if (!perfLookup.containsKey(indexStr) || !perfLookup.get(indexStr).containsKey(metricId)) {
```
pmuetschard, line 207: I would prefer
```
Map<Integer, Double> metrics = perfLookup.get(indexStr);
return (metrics == null) ? Double.NaN : metrics.getOrDefault(metricId, Double.NaN);
```

pmuetschard (approved):

gapic/src/main/com/google/gapid/models/Profile.java
stellama0208: Thanks for suggestion, make sense to improve the efficiency here since it would be run frequently, done with the updated branch!

PR #505, opened 2020-10-21
Bug report

yalcinmelihyasin:
We should rename GAPID to AGI


hevrard (approved):

pmuetschard (commented):
There is... little reason to do this. Also, after you submit it, any previously released AGI will now point to a non-existing template. Because of that, I would say, let's not do this.

yalcinmelihyasin:
@pmuetschard I see. That's a big problem. But on the other hand, we should not let the name GAPID sticks with us eternally. What is our policy for this situations? Do we have a grace period for backward compatibility? For example, keeping two versions of file for a time periods(let's say 6 month as an example) and except users to switch to a newer version?

pmuetschard:
We'll probably be stuck with "GAPID" in certain places forever. In this specific case, it's not even a problem as besides the URL, it's not visible to the user.

lpy:
Would it work if we just keep the old template?

lpy:
Also just driven-by comment: both the PR description and summary can describe the change better. `Bug report` as a title is very confusing...

yalcinmelihyasin:
Decided to not proceed as it creates more problems than it solves.

yalcinmelihyasin:
@lpy has a suggestion to keep both old and new template together and deprecate the old one after a few releases.(In a few month) as they should be already replaced.

yalcinmelihyasin:
@pmuetschard what do you think about @lpy 's idea?

lpy:
Before we close, I would like to understand whether we can deprecate the old template by keeping it for a while. After a few releases the old releases should likely be upgraded. Then we can remove the template.

pmuetschard:
This is a lot of churn for a non-issue. Why bother?

PR #510, opened 2020-10-22
Re-label dispatches as ExecutedDispatches instead of ExecutedDraws

hliatis:
Added IsExecutedDispatch test:
gapis/resolve/resource_data.go

Did not add test:
gapis/api/vulkan/draw_call_mesh.go
gapis/resolve/filter.go
gapis/resolve/stats.go

Bug: b/161536617

### DIFF
--- a/gapis/api/cmd_flags.go
+++ b/gapis/api/cmd_flags.go
@@ -27,6 +27,7 @@ const (
 	PopUserMarker
 	UserMarker
 	ExecutedDraw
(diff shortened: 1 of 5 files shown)


hliatis:
Let me know if I should add the test to those files.

hevrard (approved):
LGTM, and I think it's OK to only and keep using only IsExecutedDraw in all places but resource_data.go.

PR #512, opened 2020-10-23
Translate the counter unit into a readible format in perftab.

stellama0208:
Bug: b/169453133.

### DIFF
--- a/gapis/trace/android/mali/profiling_data.go
+++ b/gapis/trace/android/mali/profiling_data.go
@@ -27,6 +27,7 @@ import (
 	"github.com/google/gapid/gapis/service"
 	"github.com/google/gapid/gapis/service/path"
 	"github.com/google/gapid/gapis/trace/android/profile"
+	"github.com/google/gapid/gapis/trace/android/utils"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


gapis/trace/android/mali/profiling_data.go
```diff
@@ -252,6 +253,13 @@ func processGpuSlices(ctx context.Context, processor *perfetto.Processor, captur
 }
 
 func processCounters(ctx context.Context, processor *perfetto.Processor, desc *device.GpuCounterDescriptor) ([]*service.ProfilingData_Counter, error) {
+	// Use counter name, rather than counter id as key here, since the counter id
+	// is different from the counter track id that we would query.
+	specMapping := map[string]*device.GpuCounterDescriptor_GpuCounterSpec{}
+	for _, spec := range desc.GetSpecs() {
+		specMapping[spec.GetName()] = spec
+	}
```
hevrard, line 261: Could this mapping be built once-for-all rather than every time processCounters() is called?

It looks like this mapping only depends on `desc`, which itself only depends on the device configuration, see how `desc` is obtained in gapis/trace/android/trace.go:ProcessProfilingData(). If so, we could create this mapping once and somehow attach it to the device binding information, to avoid re-creating it everytime here. What do you think?

hevrard (commented):
The logic looks good. I'm not sure FormatUnit is best located in a new trace/android/utils package, could this function maybe land in an already existing package?

pmuetschard (changes requested):
This should not live on the server. We have code in the client to already do this and we'll need the specific unit details in order to properly format the values on the client. See https://github.com/google/agi/blob/master/gapic/src/main/com/google/gapid/perfetto/Unit.java

stellama0208:
A new PR is created to address the unit formatting issue. #517

PR #513, opened 2020-10-26
Add scaffolding for the framegraph feature

hevrard:
This adds the scaffolding code to start working on the framegraph
feature: a gapit framegraph verb, a GetFramegraph RPC, and a
framegraph module.

The code here always returns an empty framegraph: this is on purpose,
the actual creation of a meaningful framegraph will be added in
subsequent changes, once this scaffolding is in place.

Bug: b/171694399

### DIFF
--- a/cmd/gapit/BUILD.bazel
+++ b/cmd/gapit/BUILD.bazel
@@ -31,6 +31,7 @@ go_library(
         "dump_shaders.go",
         "export_replay.go",
         "flags.go",
+        "framegraph.go",
         "inputs.go",
         "main.go",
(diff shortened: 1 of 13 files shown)


hevrard:
Some comments from a chat on another channels:

- make the framegraph accessible via a Get() request
- define a `GetFramegraph` to the `api` interface
- the main logic will be in the Vulkan implement of this `GetFramegraph` rather than in a separate module like the current proposal.

hevrard:
Superseded by https://github.com/google/agi/pull/526

PR #514, opened 2020-10-26
Add a mean to retrieve subCmd in MutateSubcommands

hevrard:
This change propagates a reference to the subcommand upon which
preSubCmdCallback() and postSubCmdCallback() are called.

Each of these two callbacks, used in MutateSubcommands() and
MutateWithSubcommands(), currently receives as argument:

- the complete api.SubCmdIdx of the subcommand

- the api.Cmd corresponding to the **parent** command

While there are ways to retrieve the actual subcommand from these two
arguments, it would be often a lot of work.

However, these callbacks are actually called in a context where we do
have a vulkan.CommandReferenceʳ to the subcommand, see
e.g. gapis/vulkan/api/externs.go:onPostSubcommand(). This reference is
being passed as an interface{} because it is API-specific, while
MutateSubcommands() is API-agnostic. This change makes sure that this
command reference is made accessible to the callbacks.

Currently, the only user of MutateWithSubcommands() is
FramebufferChangesResolvable.Resolve(), and it does not need this
command reference, but the upcoming framegraph code will require it.

Bug: b/171719615

### DIFF
--- a/gapis/api/sync/sync.go
+++ b/gapis/api/sync/sync.go
@@ -58,9 +58,10 @@ type SynchronizedAPI interface {
 	// MutateSubcommands mutates the given Cmd and calls callbacks for subcommands
 	// attached to that Cmd. preSubCmdCallback and postSubCmdCallback will be
 	// called before and after executing each subcommand callback.
-	// Both preSubCmdCallback() and postSubCmdCallback() receive the top command as a
-	// api.Cmd, the complete subcommand index, and an interface that enables to
-	// retrieve the actual subcommand via API-specific primitives.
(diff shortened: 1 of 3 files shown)


AWoloszyn (commented):
This seems reasonable to me. I generally dislike the interface{} but there is precedence, and I cant think of a nicer way.

yalcinmelihyasin:
My only (optional) comment is should we need to keep the lambdas doing the actual work? Or would it be clear to read, if the logic would move to a function and lambda's would call it by adding needed captured values to the incoming parameters and calling it.

Otherwise, LGTM.

yalcinmelihyasin (approved):

pmuetschard (approved):

PR #516, opened 2020-10-27
Disabling Anisotropic Filtering Experiment

yalcinmelihyasin:
Adds the capability of disabling anisotropic filtering for all samplers

b/171811131

### DIFF
--- a/gapis/api/vulkan/transform_af_disabler.go
+++ b/gapis/api/vulkan/transform_af_disabler.go
@@ -1,0 +1,89 @@
+// Copyright (C) 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vulkan
+
+import (
+	"context"
+
+	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/api/transform"
+	"github.com/google/gapid/gapis/memory"
+)
+
+type afDisablerTransform struct {
+	allocations *allocationTracker
+}
+
+func newAfDisablerTransform() *afDisablerTransform {
+	return &afDisablerTransform{
+		allocations: nil,
+	}
+}
+
+func (afDisabler *afDisablerTransform) RequiresAccurateState() bool {
+	return false
+}
+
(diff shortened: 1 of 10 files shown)


gapis/api/vulkan/transform_af_disabler.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (afDisabler *afDisablerTransform) modifySamplerCreation(ctx context.Context, cmd *VkCreateSampler, inputState *api.GlobalState) api.Cmd {
+	cmd.Extras().Observations().ApplyReads(inputState.Memory.ApplicationPool())
+
+	pAlloc := memory.Pointer(cmd.PAllocator())
+	pSampler := memory.Pointer(cmd.PSampler())
+
+	pInfo := cmd.PCreateInfo()
+	info := pInfo.MustRead(ctx, cmd, inputState, nil)
+	info.SetAnisotropyEnable(VkBool32(0))
```
hevrard, line 78: Add a comment here to highlight the meaningful change of this transform, e.g.:

```suggestion
	// Disable anisotropic filtering
	info.SetAnisotropyEnable(VkBool32(0))
```

gapis/api/vulkan/transform_af_disabler.go
```diff
@@ -0,0 +1,89 @@
...
+	newInfo := afDisabler.allocations.AllocDataOrPanic(ctx, info)
+
+	cb := CommandBuilder{Thread: cmd.Thread(), Arena: inputState.Arena}
+	newCmd := cb.VkCreateSampler(cmd.Device(), newInfo.Ptr(), pAlloc, pSampler, cmd.Result())
+	newCmd.AddRead(newInfo.Data())
+	for _, w := range cmd.Extras().Observations().Writes {
+		newCmd.AddWrite(w.Range, w.ID)
+	}
```
hevrard, line 86: Should you also loop on all the `cmd.Extras().Observations().Reads`, or are we sure there is only one read corresponding to `newInfo`?

hevrard (commented):
The logic looks good to me!

I have two high-level comments:

- This PR does 2 changes that I think could be separated in 2 PRs, in this order:
  1. create the `message ProfileExperiments` proto and move the disable command experiment in it, cascading all changes related to "disable commands".
  2. introduce the new transform to disable anisotropic filtering

   At the very least, if this PR isn't split in two, the commit message should clarify that there are two changes happening here.

- Avoid using "af" as a shortcut for "anisotropic filtering" in type/function/field names (variable names is ok): out of context, "af" is not a recognizable acronym, and when in the future some developer wants to search over the code base to find code related to anisotropic filtering, if they have to search for "af", this will match a lot of unrelated code. In practice, I suggest the following renamings:
  - `DisableAF` -> `DisableAnisotropicFiltering`
  - `transform_af_disabler.go` -> `transform_anisotropic_filtering_disabler.go`
  - `type afDisablerTransform` -> `type anisotropicFilteringDisablerTransform`
  - `func newAfDisablerTransform` -> `func newAnisotropicFilteringDisablerTransform`
  - it's OK to keep e.g. `afDisabler` as a variable name.

PR #517, opened 2020-10-27
Format counter unit and counter values in perftab.

stellama0208:
Bug: b/169453133.

### DIFF
--- a/gapis/trace/android/profile/profile.go
+++ b/gapis/trace/android/profile/profile.go
@@ -16,12 +16,14 @@ package profile
 
 import (
 	"context"
+	"fmt"
 	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/math/u64"
+	"github.com/google/gapid/core/os/device"
 	"github.com/google/gapid/gapis/service"
 )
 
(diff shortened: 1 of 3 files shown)


stellama0208:
Override the old PR #512

gapis/trace/android/profile/profile.go
```diff
@@ -82,13 +84,13 @@ func setTimeMetrics(groupToSlices map[int32][]*service.ProfilingData_GpuSlices_S
 	*metrics = append(*metrics, &service.ProfilingData_GpuCounters_Metric{
 		Id:   gpuTimeMetricId,
 		Name: "GPU Time",
-		Unit: "ns",
+		Unit: fmt.Sprint(device.GpuCounterDescriptor_MeasureUnit_value[device.GpuCounterDescriptor_NANOSECOND.String()]),
```
pmuetschard, line 87: You don't need to go through all these mappings (i.e. now it maps int->name->int->string). Just
`strconv.Itoa(device.GpuCounterDescriptor_NANOSECOND)` will do :)

stellama0208: Yeah the converting is painful! I actually had a try on the `strconv.Itoa(device.GpuCounterDescriptor_NANOSECOND)` but would get an error of `couldn't use type device.GpuCounterDescriptor_MeasureUnit) as type int`, it seems to be not seeing it as a golang enum. But now I add a smaller wrapper of `int` it seems to work!

pmuetschard (approved):

PR #518, opened 2020-10-27
Limit percentage counter's value to 2 decimal in perftab.

stellama0208:
Bug: N/A.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/Unit.java
+++ b/gapic/src/main/com/google/gapid/perfetto/Unit.java
@@ -195,13 +195,19 @@ public class Unit {
     }
 
     public static class Percent extends Simple {
+      private static final double CUT_OFF = 0.05;
+
       public Percent() {
         super("%");
       }
 
       @Override
       public String format(double value) {
-        return (value == 100) ? super.format(100) : super.format(value);
+        if (value == 100) {
+          return super.format(100);
+        } else {
+          return value < CUT_OFF ? String.format("%,g%%", value) : String.format("%,.2f%%", value);
+        }
       }
     }
 


gapic/src/main/com/google/gapid/perfetto/Unit.java
```diff
@@ -201,7 +201,7 @@ public Percent() {
 
       @Override
       public String format(double value) {
-        return (value == 100) ? super.format(100) : super.format(value);
+        return (value == 100) ? super.format(100) : String.format("%.2f%s", value, "%");
```
pmuetschard, line 204: `String.format("%,.2f%%", value)`

gapic/src/main/com/google/gapid/perfetto/Unit.java
```diff
@@ -195,13 +195,18 @@ public String format(double value) {
     }
 
     public static class Percent extends Simple {
+      private static final Double CUT_OFF = 0.05;
```
pmuetschard, line 198: This should not be a boxed double. Also, add a blank line between the constant and the constructor.

gapic/src/main/com/google/gapid/perfetto/Unit.java
```diff
@@ -195,13 +195,18 @@ public String format(double value) {
...
         super("%");
       }
 
       @Override
       public String format(double value) {
-        return (value == 100) ? super.format(100) : super.format(value);
+        if (value == 100) {
+          return super.format(100);
+        } else {
+          return Math.abs(value) < CUT_OFF ? String.format("%,.2g%%", value) : String.format("%,.2f%%", value);
```
pmuetschard, line 208: You don't need to do the abs here, as it's done above - the formatter should only be getting positive values.
Also, the format should be `"%,g%%"`.

stellama0208: Thanks for being patient here! Just done it! 👍

pmuetschard (approved):

PR #519, opened 2020-10-28
Link a Profile tab's GPU slice to corresponding renderpass command node.

stellama0208:
- Instead of linking it to the vkCmdBeginRenderPass command.
 - Provide a "preferGroup" option when resolving path.CommandTreeNodeForCommand,
 if two CommandTreeNodes share the same command index, but one is a CmdIDGroup and one
 is a command, by this toggle, we could control which CommandTreeNode should be
 returned by the server.
 - Bug: b/170310168.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/CommandStream.java
+++ b/gapic/src/main/com/google/gapid/models/CommandStream.java
@@ -18,6 +18,7 @@ package com.google.gapid.models;
 import static com.google.gapid.proto.service.memory.Memory.PoolNames.Application_VALUE;
 import static com.google.gapid.util.Paths.command;
 import static com.google.gapid.util.Paths.commandTree;
+import static com.google.gapid.util.Paths.commandTreeNodeForCommand;
 import static com.google.gapid.util.Paths.lastCommand;
 import static com.google.gapid.util.Paths.observationsAfter;
 import static com.google.gapid.widgets.Widgets.submitIfNotDisposed;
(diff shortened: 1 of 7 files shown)


gapic/src/main/com/google/gapid/models/CommandStream.java
```diff
@@ -233,19 +234,6 @@ protected void onUiThread(Path.CommandTreeNode result) {
...
-        while (path.getSegmentCount() > 0) {
-          CommandStream.Node node = (CommandStream.Node)path.getLastSegment();
-          if (!node.isLastChild()) {
-            break;
-          }
-          path = path.getParentPath();
-        }
-        return path;
-      });
-    }
```
stellama0208, line 248: Trying to delete it since 1. This part of code is not being used: before this PR there's no longer `CommandIndex` initialized with the `group` value set to be true. 2. `Find the deepest group/node in the path that is not the last child of its parent.` doesn't seem to handle the aim of finding the group, especially this will do wrong to a corner case: e.g. the path passed in is already a valid group but happens to be the last child, then the path be wrongly transformed into one of its parents(seen this bug in testing).

The `group` variable will be turned into use to config how gapis resolve a command index into a real tree node.

stellama0208:
Mark it as WIP since found a behavior to improve on in this PR: should look into why when selecting the renderpass node returned from server, the status in other tabs are for the first command in the renderpass. (By comparison, selecting by mouse the status is for the last command)

stellama0208:
Updated the PR to solve the previous existing bug, now the slice will link to the renderpass grouping command, and load other resources tabs based on the "representation" command in the renderpass.
Ready for formal review now, thanks for patience!

gapic/src/main/com/google/gapid/models/CommandStream.java
stellama0208: A corner case: 
--Submit 0
----FrameBuffer0                      <- (c)
------Renderpass0
------Renderpass1
------Renderpass2                    <- (b)
--------vkBeginRenderPass.    <- (a)
----FrameBuffer1

Although our intention towards group handling here is to jump back from (a) to (b), but we would actually jump back to (c), since it happens to be the last renderpass.

hevrard (approved):
LGTM but I'm not very familiar with this part of the code.

PR #522, opened 2020-10-28
Set the screen on flag in gapir.

pmuetschard:


stellama0208 (approved):
Verified it will keep my Note10-975 awake even when I don't turn on the "Stay Awake" developer option.

PR #524, opened 2020-10-29
Avoid using nil source component.

lpy:
When resolving the mapping, there's no guarantee a non-nil source component is
properly created, and hence do a nil check before we use it, skip subsequent
operation.

Bug: b/171928591

### DIFF
--- a/core/stream/convert.go
+++ b/core/stream/convert.go
@@ -90,6 +90,10 @@ func Convert(dst, src *Format, data []byte) ([]byte, error) {
 	// Calculate min/max if floats
 	min, max := float64(math.MaxFloat64), -float64(math.MaxFloat64)
 	for _, m := range mappings {
+		if m.src.component == nil {
+			return nil, fmt.Errorf("Channel %v not found in source format: %v",
+				m.dst.component.Channel, src)
+		}
 		if m.dst.component.DataType.IsInteger() && m.src.component.DataType.IsFloat() && !m.dst.component.DataType.Signed {
 			readMinMax(count, m.src, &min, &max)
 		}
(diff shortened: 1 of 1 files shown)


lpy:
My very naive attempted fix

yalcinmelihyasin:
No harm of doing that but the callstack in you bug does not seems to verify that this will fix. Just after the check we are reading `src.component` and I would expect the crash to be happen there but instead it happens in Convert function in image.go either during store operation or reading the fields of info.

yalcinmelihyasin:
Also the particular loop is to read min and max data from the channels. In the next loop which actually does the conversion, there is already a check.

What I suggest would be merging this, as it is already preventing another crash(but maybe logging a warning etc.) but looking for the other reasons that crash can occur.

lpy:
The `if` condition reads `m.src.component.DataType` but `m.src.component` can be nil. In the particular trace I attached in the bug, the source format only has the stencil channel, destination channel is red, so no `src.component` is created.

core/stream/convert.go
```diff
@@ -90,6 +90,9 @@ func Convert(dst, src *Format, data []byte) ([]byte, error) {
 	// Calculate min/max if floats
 	min, max := float64(math.MaxFloat64), -float64(math.MaxFloat64)
 	for _, m := range mappings {
+		if m.src.component == nil {
+			continue
+		}
```
pmuetschard, line 95: Instead of doing the continue, move the error return from the below loop here (and adjust the comment). That makes it a proper early exit on error, so we won't ever convert any channels if we end up returning an error for a missing mapping anyway, so an overall improvement, too.

lpy:
Thanks for the suggestion! PTAL!

core/stream/convert.go
```diff
@@ -90,6 +90,10 @@ func Convert(dst, src *Format, data []byte) ([]byte, error) {
 	// Calculate min/max if floats
 	min, max := float64(math.MaxFloat64), -float64(math.MaxFloat64)
 	for _, m := range mappings {
+		if m.src.component == nil {
+			return nil, fmt.Errorf("Channel %v not found in source format: %v",
+				m.dst.component.Channel, src)
+		}
```
pmuetschard, line 96: Now that the check is here, remove it from the loop below.

lpy:
😂 
ptal!

pmuetschard (approved):

PR #525, opened 2020-10-29
Split texture tab in list and view

johnplate:
This splits the texture tab into a separate texture list and texture view.

First PR for bug: b/170298289

### DIFF
--- a/gapic/src/main/com/google/gapid/GraphicsTraceView.java
+++ b/gapic/src/main/com/google/gapid/GraphicsTraceView.java
@@ -44,6 +44,7 @@ import com.google.gapid.views.ReportView;
 import com.google.gapid.views.ShaderView;
 import com.google.gapid.views.StateView;
 import com.google.gapid.views.Tab;
+import com.google.gapid.views.TextureList;
 import com.google.gapid.views.TextureView;
 import com.google.gapid.widgets.TabArea;
 import com.google.gapid.widgets.TabArea.FolderInfo;
@@ -137,7 +138,7 @@ public class GraphicsTraceView extends Composite implements MainWindow.MainView,
 
   @Override
   public void onTextureFollowed(Service.Resource resource) {
-    tabs.showTab(MainTab.Type.Textures);
+    tabs.showTab(MainTab.Type.TextureView);
   }
 
   @Override
@@ -370,13 +371,14 @@ public class GraphicsTraceView extends Composite implements MainWindow.MainView,
 
       Framebuffer(View.Framebuffer, "Framebuffer", DefaultPosition.Center, FramebufferView::new),
       Pipeline(View.Pipeline, "Pipeline", DefaultPosition.Center, PipelineView::new),
-      Textures(View.Textures, "Textures", DefaultPosition.Center, TextureView::new),
+      TextureList(View.TextureList, "Textures", DefaultPosition.Center, TextureList::new),
       Geometry(View.Geometry, "Geometry", DefaultPosition.Center, GeometryView::new),
       Shaders(View.Shaders, "Shaders", DefaultPosition.Center, ShaderView::new),
       Performance(View.Performance, "Performance(Experimental)", DefaultPosition.Center, PerformanceView::new),
       Report(View.Report, "Report", DefaultPosition.Center, ReportView::new),
       Log(View.Log, "Log", DefaultPosition.Center, (p, m, w) -> new LogView(p, w)),
 
+      TextureView(View.TextureView, "Texture", DefaultPosition.Right, TextureView::new),
       ApiState(View.State, "State", DefaultPosition.Right, StateView::new),
       Memory(View.Memory, "Memory", DefaultPosition.Right, MemoryView::new);
 
--- a/gapic/src/main/com/google/gapid/GraphicsTraceView.java
+++ b/gapic/src/main/com/google/gapid/GraphicsTraceView.java
@@ -25,6 +25,7 @@ import com.google.common.collect.Sets;
 import com.google.gapid.models.Analytics.View;
 import com.google.gapid.models.Follower;
 import com.google.gapid.models.Models;
+import com.google.gapid.models.Resources;
 import com.google.gapid.models.Settings;
 import com.google.gapid.proto.SettingsProto;
 import com.google.gapid.proto.service.Service;
@@ -71,7 +72,8 @@ import java.util.function.Function;
 /**
  * Main view shown when a graphics trace is loaded.
  */
-public class GraphicsTraceView extends Composite implements MainWindow.MainView, Follower.Listener {
+public class GraphicsTraceView extends Composite
+    implements MainWindow.MainView, Resources.Listener, Follower.Listener {
   private final Models models;
   private final Widgets widgets;
   protected final Set<MainTab.Type> hiddenTabs;
@@ -102,8 +104,10 @@ public class GraphicsTraceView extends Composite implements MainWindow.MainView,
 
     tabs.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
 
+    models.resources.addListener(this);
     models.follower.addListener(this);
     addListener(SWT.Dispose, e -> {
+      models.resources.removeListener(this);
       models.follower.removeListener(this);
     });
   }
@@ -127,6 +131,13 @@ public class GraphicsTraceView extends Composite implements MainWindow.MainView,
   }
 
   @Override
+  public void onTextureSelected(Service.Resource texture) {
+    if (texture != null) {
+      tabs.showTab(MainTab.Type.TextureView);
+    }
+  }
+
+  @Override
   public void onMemoryFollowed(Path.Memory path) {
     tabs.showTab(MainTab.Type.Memory);
   }
@@ -134,11 +145,6 @@ public class GraphicsTraceView extends Composite implements MainWindow.MainView,
   @Override
   public void onStateFollowed(Path.Any path) {
     tabs.showTab(MainTab.Type.ApiState);
-  }
-
-  @Override
-  public void onTextureFollowed(Service.Resource resource) {
-    tabs.showTab(MainTab.Type.TextureView);
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


gapic/src/main/com/google/gapid/GraphicsTraceView.java
```diff
@@ -370,13 +371,14 @@ private static void flatten(
...
       Pipeline(View.Pipeline, "Pipeline", DefaultPosition.Center, PipelineView::new),
-      Textures(View.Textures, "Textures", DefaultPosition.Center, TextureView::new),
+      TextureList(View.TextureList, "Textures", DefaultPosition.Center, TextureList::new),
       Geometry(View.Geometry, "Geometry", DefaultPosition.Center, GeometryView::new),
       Shaders(View.Shaders, "Shaders", DefaultPosition.Center, ShaderView::new),
       Performance(View.Performance, "Performance(Experimental)", DefaultPosition.Center, PerformanceView::new),
       Report(View.Report, "Report", DefaultPosition.Center, ReportView::new),
       Log(View.Log, "Log", DefaultPosition.Center, (p, m, w) -> new LogView(p, w)),
 
+      TextureView(View.TextureView, "Texture", DefaultPosition.Right, TextureView::new),
```
stellama0208, line 381: nit: The names of these two tabs seem to be too close?

stellama0208 (approved):
LGTM! Also verified the function.

gapic/src/main/com/google/gapid/views/TextureList.java
```diff
@@ -0,0 +1,551 @@
+/*
+ * Copyright (C) 2017 Google Inc.
```
pmuetschard, line 2: wrong year

gapic/src/main/com/google/gapid/views/TextureList.java
```diff
@@ -0,0 +1,551 @@
...
+
+  private final Models models;
+  private final TableViewer textureTable;
+  private final ImageProvider imageProvider;
+
+  public TextureList(Composite parent, Models models, Widgets widgets) {
+    super(parent, SWT.NONE);
+    this.models = models;
+
+    setLayout(new FillLayout(SWT.VERTICAL));
```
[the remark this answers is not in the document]
pmuetschard: to me this is not a nit. you should do this.

gapic/src/main/com/google/gapid/views/TextureList.java
```diff
@@ -0,0 +1,551 @@
...
+      if (d.info.getID().equals(resource.getID())) {
+        textureTable.getTable().setSelection(items[i]);
+        break;
+      }
+    }
+  }
+
+  private void updateTextures(boolean resourcesChanged) {
+    if (!models.resources.isLoaded()) {
+      // TODO: show capture loading message
```
pmuetschard, line 207: This needs to be done before it's submitted.

gapic/src/main/com/google/gapid/views/TextureList.java
```diff
@@ -0,0 +1,551 @@
...
+          });
+
+      textureTable.setInput(textures);
+      packColumns(textureTable.getTable());
+
+      if (!resourcesChanged && selection >= 0 && selection < textures.size()) {
+        textureTable.getTable().select(selection);
+      }
+      textureTable.setComparator(comparator);
+      updateSelection();
```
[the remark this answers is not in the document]
pmuetschard: Also, not a TODO, it should be part of this PR.

gapic/src/main/com/google/gapid/views/TextureList.java
```diff
@@ -0,0 +1,551 @@
...
+    textureTable.getTable().requestLayout();
+    imageProvider.reset();
+  }
+
+  private void updateSelection() {
+    int selection = textureTable.getTable().getSelectionIndex();
+    if (selection >= 0) {
+      Data data = (Data)textureTable.getElementAt(selection);
+      Path.ResourceData path = models.resources.getResourcePath(data.info);
+      models.follower.onFollow(Path.Any.newBuilder().setResourceData(path).build());
```
pmuetschard, line 251: I wonder if it makes more sense to manage this via the resource model - i.e. have something like `Resources.selectTexture`, etc. but maybe for now this is ok.

gapic/src/main/com/google/gapid/views/TextureList.java
```diff
@@ -0,0 +1,551 @@
+/*
```
pmuetschard, line 1: It'd be nice if git tracked this as a copy of `TextureView.java`.

gapic/src/main/com/google/gapid/views/TextureView.java
```diff
@@ -119,99 +80,42 @@ public TextureView(Composite parent, Models models, Widgets widgets) {
...
-
-    textureTable.getTable().addListener(SWT.Selection, e -> updateSelection());
-    showDeleted.addListener(SWT.Selection, e -> {
-      if (showDeleted.getSelection()) {
-        textureTable.removeFilter(filterDeleted);
-      } else {
-        textureTable.addFilter(filterDeleted);
-      }
     });
+    reinitialize();
```
pmuetschard, line 100: We should not do this. What are you relying on to happen in `reinitialize` for the constructor?

gapic/src/main/com/google/gapid/views/TextureView.java
```diff
@@ -119,99 +80,42 @@ public TextureView(Composite parent, Models models, Widgets widgets) {
...
 
   @Override
   public void reinitialize() {
-    onCaptureLoadingStart(false);
-    updateTextures(true);
+    if (!models.capture.isLoaded()) {
+      onCaptureLoadingStart(false);
+    } else {
+      onCaptureLoaded(null);
+    }
```
pmuetschard, line 118: FYI: This doesn't restore the currently selected texture, which is why I think that should be tracked in the model, rather than via the follow.

pmuetschard (changes requested):

gapic/src/main/com/google/gapid/views/TextureList.java
johnplate: I agree, but unfortunately [git doesn't support](https://stackoverflow.com/questions/1043388/record-file-copy-operation-with-git) tracking of copies and known workarounds don't work anymore with recent git versions.

gapic/src/main/com/google/gapid/GraphicsTraceView.java
johnplate: Yes, I was considering to rename the "Textures" tab to "Texture List", but I preferred to keep the former to avoid the additional space requirement in the tab bar if we consistently would add "List" to all resource list tabs. It might not be a problem if users prefer to have the texture list and the views in separate UI folders.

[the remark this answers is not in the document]
johnplate: As discussed I will keep the TextureView a main tab for now as its customized position should be persistent. Only the TextureView which shows the currently selected texture will be a main tab. I a later PR I will introduce pinned texture views which are not main tabs.

gapic/src/main/com/google/gapid/views/TextureList.java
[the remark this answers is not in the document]
johnplate: Thanks Stella, good catch! But for the LoadablePanel I need a single Control, so I will use the existing tableAndOption Composite.

gapic/src/main/com/google/gapid/views/TextureList.java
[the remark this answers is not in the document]
johnplate: I added LoadablePanel to show NO_TEXTURES, but we can't show SELECT_TEXTURE here as it would hide the table and it wouldn't be possible to select a texture.

gapic/src/main/com/google/gapid/GraphicsTraceView.java
```diff
@@ -370,13 +371,14 @@ private static void flatten(
 
       Framebuffer(View.Framebuffer, "Framebuffer", DefaultPosition.Center, FramebufferView::new),
       Pipeline(View.Pipeline, "Pipeline", DefaultPosition.Center, PipelineView::new),
-      Textures(View.Textures, "Textures", DefaultPosition.Center, TextureView::new),
+      TextureList(View.TextureList, "Textures", DefaultPosition.Center, TextureList::new),
```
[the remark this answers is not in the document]
johnplate: Fixed

gapic/src/main/com/google/gapid/models/Analytics.java
```diff
@@ -34,7 +34,7 @@
     Main, LeftTabs, RightTabs,
     About, Help, GotoCommand, GotoMemory, Licenses, Settings, Trace, Welcome,
     // See MainWindow.MainTab.Type
-    FilmStrip, Profile, Commands, Framebuffer, Pipeline, Textures, Geometry, Shaders, Performance, Report, Log, State, Memory,
+    FilmStrip, Profile, Commands, Framebuffer, Pipeline, TextureList, TextureView, Geometry, Shaders, Performance, Report, Log, State, Memory,
```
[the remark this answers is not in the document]
johnplate: Fixed

gapic/src/main/com/google/gapid/views/TextureList.java
johnplate: Fixed

gapic/src/main/com/google/gapid/views/TextureList.java
johnplate: Done

gapic/src/main/com/google/gapid/views/TextureList.java
johnplate: I agree, implemented as discussed.

gapic/src/main/com/google/gapid/views/TextureView.java
johnplate: Fixed as discussed.

gapic/src/main/com/google/gapid/views/TextureView.java
johnplate: Tracked in the model as discussed.

gapic/src/main/com/google/gapid/models/Resources.java
```diff
@@ -177,6 +185,17 @@ protected void onUiThread(Path.Capture result) {
     });
   }
 
+  public Service.Resource getSelectedTexture() {
+    return selectedTexture;
+  }
+
+  public void selectTexture(Service.Resource texture) {
+    if (selectedTexture != texture) {
```
pmuetschard, line 193: this is a pointer comparison and is unlikely to work properly for anything other than `null`s.

gapic/src/main/com/google/gapid/models/Follower.java
```diff
@@ -69,6 +69,8 @@ public Follower(Shell shell, Client client, Resources resources) {
     this.shell = shell;
     this.client = client;
     this.resources = resources;
+
+    addListener(resources);
```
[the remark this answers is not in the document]
johnplate: Thanks, done. That solution makes me more happy as well!

gapic/src/main/com/google/gapid/models/Resources.java
johnplate: Thank you, fixed

pmuetschard (approved):

PR #526, opened 2020-10-29
Add scaffolding for the framegraph feature

hevrard:
This adds the scaffolding code to start working on the framegraph
feature: a gapit framegraph verb and a Framegraph path to be resolved
by a Get RPC.

The code here always returns an empty framegraph: this is on purpose,
the actual creation of a meaningful framegraph will be added in
subsequent changes, once this scaffolding is in place.

Bug: b/171694399

### DIFF
--- a/cmd/gapit/framegraph.go
+++ b/cmd/gapit/framegraph.go
@@ -1,0 +1,103 @@
+// Copyright (C) 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/google/gapid/core/app"
+	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/gapis/api"
+)
+
+// Default output file name.
+const defaultOutFilename = "framegraph.dot"
+
+type framegraphVerb struct{ FramegraphFlags }
+
+func init() {
+	verb := &framegraphVerb{}
+	app.AddVerb(&app.Verb{
+		Name:      "framegraph",
+		ShortHelp: "Create frame graph (in DOT format) from capture",
+		Action:    verb,
+	})
+}
+
+// Run is the main logic for the 'gapit framegraph' command.
+func (verb *framegraphVerb) Run(ctx context.Context, flags flag.FlagSet) error {
+	if flags.NArg() != 1 {
+		app.Usage(ctx, "Exactly one gfx trace file expected, got %d", flags.NArg())
+		return nil
+	}
+
+	captureFilename := flags.Arg(0)
+
+	client, capture, err := getGapisAndLoadCapture(ctx, verb.Gapis, GapirFlags{}, captureFilename, CaptureFileFlags{})
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	boxedFramegraph, err := client.Get(ctx, capture.Framegraph().Path(), nil)
+	if err != nil {
+		return err
+	}
+	framegraph := boxedFramegraph.(*api.Framegraph)
+
+	dot := framegraph2dot(framegraph, captureFilename)
+
+	// Write the DOT representation of the framegraph into a file
+	filePath := verb.Out
+	if filePath == "" {
+		filePath = defaultOutFilename
+	}
+	file, err := os.Create(filePath)
+	if err != nil {
+		return log.Errf(ctx, err, "Creating file (%v)", filePath)
+	}
+	defer file.Close()
+	bytesWritten, err := fmt.Fprint(file, dot)
+	if err != nil {
+		return log.Errf(ctx, err, "Error after writing %d bytes to file", bytesWritten)
+	}
+
+	return nil
+}
+
+// framegraph2dot formats a framegraph in the Graphviz DOT format.
+// https://graphviz.org/doc/info/lang.html
+func framegraph2dot(framegraph *api.Framegraph, captureFilename string) string {
+	s := "digraph agiFramegraph {\n"
+	// Graph title: use capture filename, on top
+	s += "label = \"" + captureFilename + "\";\n"
+	s += "labelloc = \"t\";\n"
+	// Use monospace font everywhere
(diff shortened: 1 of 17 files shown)


johnplate (changes requested):
The kokoro test failed:
API does not implement api.API (missing GetFramegraph method)

hevrard:
OK, I fixed the tests.

gapis/service/path/path.proto
```diff
@@ -44,6 +44,7 @@ message Any {
     FramebufferAttachments framebuffer_attachments = 18;
     FramebufferAttachment framebuffer_attachment = 19;
     Field field = 20;
+    Framegraph framegraph = 44;
```
yalcinmelihyasin, line 47: Style: I am not sure about this id. We update client and server at the same time. Therefore changing the number wouldn't be an issue(except the unnecessary change of 20 lines).

But outside of that we have a unusual usage of protobuff field numbers here. We put them in alphabetical order and chronological order(the numbers). As we easily think that we cannot guarantee that every new field will be continued to named alphabetically, just as in here, breaking this is inevitable. But, in my opinion, the more important one of those is the numbers instead of alphabetical order. So rather than breaking the number order, I would break the alphabetical order. Because every time we have to add a new field here, we would need to check all the numbers and ensure we don't have the same number. Because they have to be unique. But keeping the alphabetical order does not give anything to us.

gapis/api/vulkan/framegraph.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/service/path"
+)
+
+// GetFramegraph creates the framegraph of the given capture.
+func (API) GetFramegraph(ctx context.Context, p *path.Capture) (*api.Framegraph, error) {
+	log.W(ctx, "TODO: implement framegraph creation")
+	return &api.Framegraph{}, nil
```
yalcinmelihyasin, line 28: nit: I may choose returning nil instead of an empty struct.

gapis/service/path/path.proto
hevrard: Yes, thanks for spotting this! I was trying to at the same time not rewrite IDs and keep the alphabetical order, but as you say that approach would lead to something where it's difficult to know the current max ID.

I've moved this to the bottom of the list.

gapis/api/vulkan/framegraph.go
hevrard: Done

johnplate (approved):
LGTM, but wait for at least one more approval.

cmd/gapit/framegraph.go
```diff
@@ -0,0 +1,103 @@
...
+// framegraph2dot formats a framegraph in the Graphviz DOT format.
+// https://graphviz.org/doc/info/lang.html
+func framegraph2dot(framegraph *api.Framegraph, captureFilename string) string {
+	s := "digraph agiFramegraph {\n"
+	// Graph title: use capture filename, on top
+	s += "label = \"" + captureFilename + "\";\n"
+	s += "labelloc = \"t\";\n"
+	// Use monospace font everywhere
+	s += "node [fontname = \"Monospace\"];\n"
+	s += "\n"
```
yalcinmelihyasin, line 92: nit: Can this be reduced into a string format? e.g `fmt.sprintf("...`

yalcinmelihyasin: I think that it would be much readable.

cmd/gapit/framegraph.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return log.Errf(ctx, err, "Error after writing %d bytes to file", bytesWritten)
+	}
+
+	return nil
+}
+
+// framegraph2dot formats a framegraph in the Graphviz DOT format.
+// https://graphviz.org/doc/info/lang.html
+func framegraph2dot(framegraph *api.Framegraph, captureFilename string) string {
```
yalcinmelihyasin, line 85: A general feedback about this: 

Golang strings are immutable and every time there is a an add operation, it may trigger a new allocation. So, this does not seem to be a hot path call, it's probably a call that runs once per query so not a big issue but if you plan to do more work on this in the future, it may be good to look at if this can be converted to something easier to read, more efficient to run with either string formatting or other alternatives.

yalcinmelihyasin (approved):

cmd/gapit/framegraph.go
hevrard: Thanks for the feedback!

I am aware of the perf impact of the string manipulations here, I used it for prototyping and knowing this won't be noticeable given that we deal with not-very-big files here. (To give an idea, generating the dependency graph is in the order of several seconds, so in comparison producing the framegraph (which is much smaller than the dependency graph) dot representation is negligible.)

I'll keep your remarks in mind and improve once we have a working framegraph!

cmd/gapit/framegraph.go
hevrard: WIll adress in future PRs once we have a working framegraph MVP.

gapis/api/api.go
```diff
@@ -57,6 +58,9 @@ type API interface {
 	// The segments of memory that were used to create these commands are
 	// returned in the rangeList.
 	RebuildState(ctx context.Context, s *GlobalState) ([]Cmd, interval.U64RangeList)
+
+	// GetFramegraph returns the framegraph of the capture.
+	GetFramegraph(ctx context.Context, p *path.Capture) (*Framegraph, error)
```
pmuetschard, line 63: Since you are resolving the capture in order to get the API instance, passing in the capture path here is useless, please remove it.

gapis/resolve/framegraph.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/gapis/capture"
+	"github.com/google/gapid/gapis/database"
+	"github.com/google/gapid/gapis/service/path"
+)
+
+// Framegraph retrieves the framegraph from the database, creating it if need be.
+func Framegraph(ctx context.Context, p *path.Framegraph, r *path.ResolveConfig) (interface{}, error) {
+	obj, err := database.Build(ctx, &FramegraphResolvable{Capture: p.Capture})
```
pmuetschard, line 28: This is not necessary. Anything that is returned via the `Get` API is already stored in the database. Also, since you are calling `database.Build`, it's run in-place, so just put the code from the below `Resolve` here.

gapis/api/api.go
hevrard: The path is needed in the implementation of `GetFramegraph` for two things:
- create the dependency graph
- sync.MutateWithSubcommands

Do you prefer to pass the capture and re-create its path inside `GetFramegraph`?

gapis/resolve/framegraph.go
hevrard: Fixed with: https://github.com/google/agi/pull/542

Note that I was mirroring `gapis/resolve/resources.go`, which has this double-function layout .

PR #530, opened 2020-11-02
Framegraph: record renderpasses and their dependencies

hevrard:
Add the structure to go over subcommands, record renderpasses, and
establish their dependencies based on the dependency graph. For now,
renderpasses have very little information attached to them: this is on
purpose, to keep this commit small.

The main part of this commit is the inter-renderpass dependencies
logic, in updateDependencies().

The change in MutateWithSubcommands() is to allow passing a nil
postCmdCb to it.

Bug: b/172218304
Test: manually, on a few traces.

### DIFF
--- a/gapis/api/vulkan/framegraph.go
+++ b/gapis/api/vulkan/framegraph.go
@@ -16,14 +16,153 @@ package vulkan
 
 import (
 	"context"
+	"fmt"
 
-	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/api/sync"
+	"github.com/google/gapid/gapis/capture"
+	"github.com/google/gapid/gapis/resolve/dependencygraph2"
 	"github.com/google/gapid/gapis/service/path"
 )
 
+// renderpassInfo stores a renderpass' info relevant for the framegraph.
+type renderpassInfo struct {
+	id       uint64
+	beginIdx api.SubCmdIdx
+	endIdx   api.SubCmdIdx
+	nodes    []dependencygraph2.NodeID
+	deps     map[uint64]struct{} // set of renderpasses this renderpass depends on
+}
+
 // GetFramegraph creates the framegraph of the given capture.
 func (API) GetFramegraph(ctx context.Context, p *path.Capture) (*api.Framegraph, error) {
-	log.W(ctx, "TODO: implement framegraph creation")
-	return nil, nil
+	config := dependencygraph2.DependencyGraphConfig{
+		SaveNodeAccesses:    true,
+		ReverseDependencies: true,
+	}
+	dependencyGraph, err := dependencygraph2.GetDependencyGraph(ctx, p, config)
+	if err != nil {
+		return nil, err
+	}
+
+	rpInfos := []*renderpassInfo{}
+	var rpInfo *renderpassInfo
+	currRpId := uint64(0)
+
+	// postSubCmdCb effectively processes each subcommand to extract renderpass info.
+	postSubCmdCb := func(state *api.GlobalState, subCmdIdx api.SubCmdIdx, cmd api.Cmd, i interface{}) {
+		vkState := GetState(state)
+		cmdRef := i.(CommandReferenceʳ)
+		cmdArgs := GetCommandArgs(ctx, cmdRef, vkState)
+
+		// Beginning of renderpass
+		if _, ok := cmdArgs.(VkCmdBeginRenderPassArgsʳ); ok {
+			if rpInfo != nil {
+				panic("Renderpass starts without having ended")
+			}
+			rpInfo = &renderpassInfo{
+				id:       currRpId,
+				beginIdx: subCmdIdx,
+				nodes:    []dependencygraph2.NodeID{},
+				deps:     make(map[uint64]struct{}),
+			}
+			currRpId++
+		}
+
+		// Process commands that are inside a renderpass
+		if rpInfo != nil {
+			nodeID := dependencyGraph.GetCmdNodeID(api.CmdID(subCmdIdx[0]), subCmdIdx[1:])
+			rpInfo.nodes = append(rpInfo.nodes, nodeID)
+		}
+
+		// Ending of renderpass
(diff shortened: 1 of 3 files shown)


hevrard:
For illustration, this steps let us get this framegraph on Sascha Willems "deferred" demo:

framegraph

johnplate (approved):
LGTM, but wait for another approval

gapis/api/vulkan/framegraph.go
```diff
@@ -16,14 +16,153 @@ package vulkan
...
+	}
+
+	rpInfos := []*renderpassInfo{}
+	var rpInfo *renderpassInfo
+	currRpId := uint64(0)
+
+	// postSubCmdCb effectively processes each subcommand to extract renderpass info.
+	postSubCmdCb := func(state *api.GlobalState, subCmdIdx api.SubCmdIdx, cmd api.Cmd, i interface{}) {
+		vkState := GetState(state)
+		cmdRef := i.(CommandReferenceʳ)
```
yalcinmelihyasin, line 55: Do we know that i has a type of `CommandReferenceʳ` ? If yes, we should replace it with `interface{}`? if no, then we should check if type conversion is valid. Otherwise it's a direct crash.

gapis/api/vulkan/framegraph.go
```diff
@@ -16,14 +16,153 @@ package vulkan
...
+	if err != nil {
+		return nil, err
+	}
+
+	rpInfos := []*renderpassInfo{}
+	var rpInfo *renderpassInfo
+	currRpId := uint64(0)
+
+	// postSubCmdCb effectively processes each subcommand to extract renderpass info.
+	postSubCmdCb := func(state *api.GlobalState, subCmdIdx api.SubCmdIdx, cmd api.Cmd, i interface{}) {
```
yalcinmelihyasin, line 53: I can see that this is a form of closure that modifies the function fields several times in a loop in `MutateWithSubcommands` but this is quite subtle and can be misleading. 

This might be quite common in functional paradigm which I am not familiar with but as the rest of the code is quite non-functional(no puns intended) I think having a small helper struct with one function that actually does the operation and passing a lambda that calls the particular function in the object would be much cleaner.

gapis/api/vulkan/framegraph.go
```diff
@@ -16,14 +16,153 @@ package vulkan
...
+		text := fmt.Sprintf("Renderpass %v\\lbegin:%v\\lend:%v\\l", rpInfo.id, rpInfo.beginIdx, rpInfo.endIdx)
+		nodes[i] = &api.FramegraphNode{
+			Id:   rpInfo.id,
+			Text: text,
+		}
+	}
+
+	edges := []*api.FramegraphEdge{}
+	for _, rpInfo = range rpInfos {
+		for deps := range rpInfo.deps {
```
yalcinmelihyasin, line 113: A notifier: This is a map and in Golang this will iterate differently for every range operation for the same data. If this is considered there is no problem.

gapis/api/vulkan/framegraph.go
```diff
@@ -16,14 +16,153 @@ package vulkan
...
+func updateDependencies(rpInfos []*renderpassInfo, dependencyGraph dependencygraph2.DependencyGraph) {
+	// isInsideRenderpass: node -> renderpass it belongs to.
+	isInsideRenderpass := map[dependencygraph2.NodeID]uint64{}
+	for _, rpInfo := range rpInfos {
+		for _, n := range rpInfo.nodes {
+			isInsideRenderpass[n] = rpInfo.id
+		}
+	}
+	// node2renderpasses: node -> set of renderpasses it depends on.
+	node2renderpasses := map[dependencygraph2.NodeID]map[uint64]struct{}{}
```
yalcinmelihyasin, line 137: Future comment: This does not do anything for now but if you plan to make type assertion on the `struct{}` that's a lot of type assertion there.

gapis/api/vulkan/framegraph.go
```diff
@@ -16,14 +16,153 @@ package vulkan
...
+
+	// For a given renderpass RP, for each of its node, explore the dependency
+	// graph in reverse order to mark all the nodes dependending on RP until we
+	// hit the node of another renderpass, which then depends on RP.
+	for _, rpInfo := range rpInfos {
+		// markNode is recursive, so declare it before initializing it.
+		var markNode func(dependencygraph2.NodeID) error
+		markNode = func(node dependencygraph2.NodeID) error {
```
yalcinmelihyasin, line 145: I don't know how does go compiler handles this(maybe realises and optimizes) but instead of creating a new lambda function for every rpInfo in rpInfos this might be converted into a normal recursive function calls. But not the highest priority I would say especially if it will take long time.

yalcinmelihyasin (changes requested):
I think it is good to change the closure to a struct and a function and having a type assertion for `CommandReferenceʳ`. 

The others are optional.

gapis/api/vulkan/framegraph.go
hevrard: Yep, no worries there.

gapis/api/vulkan/framegraph.go
hevrard: This is a canonical way to represent a set in Go: `map[<type of your set>]struct{}`. Then you can use the map to test for presence in set, etc...

gapis/api/vulkan/framegraph.go
hevrard: There are O(100) renderpasses per frame on realistic workloads, so I don't think we need to worry there. If framegraph creation turns out to be slow, we'll measure and optimize accordingly.

gapis/api/vulkan/framegraph.go
hevrard: Added a type check.

gapis/api/vulkan/framegraph.go
hevrard: I think such closures are fairly common, but I agree with your comment nonetheless. I've extracted this lambda logic into a new `processSubCommand()` function.

yalcinmelihyasin (approved):

PR #534, opened 2020-11-03
fixes and patches for gcc-10 on Linux

johnplate:
This fixes a few missing C++ standard library includes in our code and in LLVM. The latter was fixed with a patch file, as it was much less effort than to update LLVM and to fix all the resulting problems of that.
Also turned off one additional warning in breakpad.

Fixes b/171865830

### DIFF
--- a/tools/build/workspace.bzl
+++ b/tools/build/workspace.bzl
@@ -71,18 +71,18 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
         project = "grpc",
         commit = "c599e6a922a80e40e24a2d3c994a6dd51046796b",  # 1.22.1
         sha256 = "d17ead923510b3c8a03eec623fffe4cba64d43e10b3695f027a1c8f10c03756a",
-	# This patch works around a naming conflict in grpc which leads to
-	# compilation issues in recent gcc/glibc. This issue is fixed on recent
-	# grpc versions (since
-	# https://github.com/grpc/grpc/commit/de6255941a5e1c2fb2d50e57f84e38c09f45023d),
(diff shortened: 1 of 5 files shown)


tools/build/workspace.bzl
```diff
@@ -158,6 +158,11 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
         commit = "e562960fe303c0ffab6f3458fcdb1544b56fd81e",
         build_file = "@gapid//tools/build/third_party:llvm.BUILD",
         sha256 = "3ef3d905849d547b6481b16d8e7b473a84efafbe90131e7bc90a0c6aae4cd8e6",
+	# This patch fixes missing standard library includes which leads to compilation
+	# issues in recent gcc. This issue is fixed on recent llvm versions (since
+	# https://github.com/llvm-mirror/llvm/commit/e0402b5c9813a2458b8dd3f640883110db280395),
+	# but updating our llvm version leads to other errors.
+	patch_file = "@gapid//tools/build/third_party:llvm_fix.patch",
```
pmuetschard, line 165: fix the indent.

Personally, this is a bit of a borderline. We don't really need to update to gcc10, and you are adding a maintenance burden here with the patch. It's a small patch, so I can live with it.

pmuetschard (approved):

tools/build/workspace.bzl
johnplate: I agree, I would also prefer to update LLVM than to use a patch. But we would have to go from llvm-mirror to llvm-project, and that effort would not be justified to support gcc-10.

The indent looked right in the editor, because I copied that part from the grpc entry, which uses tabs. I replaced all tabs in the file with spaces.

PR #535, opened 2020-11-03
Update Mesa driver workaround

johnplate:
The current workaround for a problem with the Mesa driver is limited to LLVM 9, and needs an update for the new LLVM 10.
See https://github.com/google/gapid/issues/1707

Fixes b/172372007

### DIFF
--- a/core/cc/linux/get_vulkan_proc_address.cpp
+++ b/core/cc/linux/get_vulkan_proc_address.cpp
@@ -38,7 +38,7 @@ typedef size_t VkInstance;
 struct MesaLLVMOpener {
   MesaLLVMOpener() {
     char name[512];
-    for (int i = 3; i <= 9; i++) {
+    for (int i = 3; i <= 12; i++) {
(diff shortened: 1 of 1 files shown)


core/cc/linux/get_vulkan_proc_address.cpp
```diff
@@ -38,7 +38,7 @@ typedef size_t VkInstance;
 struct MesaLLVMOpener {
   MesaLLVMOpener() {
     char name[512];
-    for (int i = 3; i <= 9; i++) {
+    for (int i = 3; i <= 12; i++) {
```
pmuetschard, line 41: odd... i have memory of already having done this, since my personal desktop has had LLVM10 for a while now. Can you up the 12 to 20, so we don't have to deal with this again for a while?

pmuetschard (approved):

core/cc/linux/get_vulkan_proc_address.cpp
johnplate: Sure, done.

PR #538, opened 2020-11-05
Don't assume attribute binding description index matches binding ID.

pmuetschard:
Binding descriptors can be in any order and don't need to be contiguous, so assuming that the index of the binding matches the binding ID is invalid.


stellama0208 (approved):
LGTM!

PR #540, opened 2020-11-05
Ignore failure to start Perfetto producer for device info.

pmuetschard:
Failing to start the Perfetto producers shuould not fail device info. Erroring out in the device info means the device will not show up, even when listing unsupported devices.


lpy:
should we log the error instead? That might help understand why a device is identified as no capabilities from the field

pmuetschard:
> should we log the error instead? That might help understand why a device is identified as no capabilities from the field

It actually already is logged by the function itself.

lpy (approved):

hevrard (approved):

PR #541, opened 2020-11-05
Fix a nil pointer server error when req.Experiments is nil.

stellama0208:
This bug happens when AGI tries to do a GPU profile replay
 but the experiment config variable is nil.
 Bug: N/A.


hevrard (approved):

yalcinmelihyasin:
Sorry I have seen this after we merged another fix for the issue. Otherwise, this would be a perfectly fine fix. 

This is the other PR that has already merged.
https://github.com/google/agi/pull/543

PR #544, opened 2020-11-05
Avoid adding null into UGD opt in list.

lpy:
At the beginning of the world there's nothing in the settings global database
and hence when queried updatable_driver_prerelease_opt_in_apps, it returns
null. Appending it back to the opt in list may do no harm but we should avoid
it.

Bug: N/A
Test: `adb shell settings put global updatable_driver_prerelease_opt_in_apps com.android.app` and verify it's appended correctly
Test: `adb shell settings delete global updatable_driver_prerelease_opt_in_apps` and verify no `null` is appended

### DIFF
--- a/core/os/android/adb/device.go
+++ b/core/os/android/adb/device.go
@@ -125,10 +125,18 @@ func SetupPrereleaseDriver(ctx context.Context, d Device, p *android.InstalledPa
 	if err != nil {
 		return nil, log.Err(ctx, err, "Failed to get prerelease driver opt in apps.")
 	}
+	// When key is not in the settings global database, a null will be returned.
+	// Avoid using it.
+	if oldOptinApps == "null" {
+		oldOptinApps = ""
+	}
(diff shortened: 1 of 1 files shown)


hevrard (approved):
LGTM

core/os/android/adb/device.go
```diff
@@ -128,7 +128,10 @@ func SetupPrereleaseDriver(ctx context.Context, d Device, p *android.InstalledPa
 	if strings.Contains(oldOptinApps, p.Name) {
 		return nil, nil
 	}
-	newOptinApps := oldOptinApps + "," + p.Name
+	newOptinApps := p.Name
+	if oldOptinApps != "null" {
+		newOptinApps = oldOptinApps + "," + newOptinApps
+	}
```
pmuetschard, line 134: Below, where we restore it, we should also set it to `""` instead of `"null"`.

lpy:
ptal

pmuetschard (approved):

PR #546, opened 2020-11-06
Framegraph: add renderpass image accesses

hevrard:
List the images accessed by a renderpass.

Refactor a bit the types to extract imageInfo as a standalone type
from attachmentInfo.

Use a lookup table to find which images match a given memory access.

Bug: b/172218645
Test: manual, on several traces

### DIFF
--- a/gapis/api/vulkan/framegraph.go
+++ b/gapis/api/vulkan/framegraph.go
@@ -17,49 +17,79 @@ package vulkan
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/google/gapid/gapis/api"
 	"github.com/google/gapid/gapis/api/sync"
 	"github.com/google/gapid/gapis/capture"
+	"github.com/google/gapid/gapis/memory"
 	"github.com/google/gapid/gapis/resolve/dependencygraph2"
 	"github.com/google/gapid/gapis/service/path"
 )
 
-type attachmentInfo struct {
-	isSwapchainImage bool
-	format           VkFormat
-	imgType          VkImageViewType
-	loadOp           VkAttachmentLoadOp
-	storeOp          VkAttachmentStoreOp
-	usage            VkImageUsageFlags
-	handle           VkImageView
+type imageInfo struct {
+	handle           VkImage
 	width            uint32
 	height           uint32
 	depth            uint32
+	usage            VkImageUsageFlags
+	imgType          VkImageType
+	format           VkFormat
+	isSwapchainImage bool
+}
+
+func (img *imageInfo) String() string {
+	swapchain := ""
+	if img.isSwapchainImage {
+		swapchain = " swapchain"
+	}
+	transient := ""
+	if (img.usage & VkImageUsageFlags(VkImageUsageFlagBits_VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT)) != 0 {
+		transient = " transient"
+	}
+	imgType := strings.TrimPrefix(fmt.Sprintf("%v", img.imgType), "VK_IMAGE_TYPE_")
+	imgFormat := strings.TrimPrefix(fmt.Sprintf("%v", img.format), "VK_FORMAT_")
+	return fmt.Sprintf("[Img %v%s%s %s %s %vx%vx%v]", img.handle, swapchain, transient, imgType, imgFormat, img.width, img.height, img.depth)
+}
+
+func newImageInfo(image *ImageObjectʳ) *imageInfo {
+	return &imageInfo{
+		handle:           image.VulkanHandle(),
+		width:            image.Info().Extent().Width(),
+		height:           image.Info().Extent().Height(),
+		depth:            image.Info().Extent().Depth(),
+		usage:            image.Info().Usage(),
+		imgType:          image.Info().ImageType(),
+		format:           image.Info().Fmt(),
+		isSwapchainImage: image.IsSwapchainImage(),
+	}
+}
+
+type attachmentInfo struct {
+	loadOp        VkAttachmentLoadOp
+	storeOp       VkAttachmentStoreOp
+	imgViewHandle VkImageView
+	imgViewType   VkImageViewType
+	imgViewFormat VkFormat
+	img           *imageInfo
 }
 
 func (a *attachmentInfo) String() string {
 	if a == nil {
 		return "unused"
 	}
-	isSwapChain := ""
-	if a.isSwapchainImage {
-		isSwapChain = " [swapchain]"
-	}
-	transient := ""
-	if (a.usage & VkImageUsageFlags(VkImageUsageFlagBits_VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT)) != 0 {
-		transient = " [transient]"
-	}
+
 	load := strings.TrimPrefix(fmt.Sprintf("%v", a.loadOp), "VK_ATTACHMENT_LOAD_OP_")
 	store := strings.TrimPrefix(fmt.Sprintf("%v", a.storeOp), "VK_ATTACHMENT_STORE_OP_")
-	imgType := strings.TrimPrefix(fmt.Sprintf("%v", a.imgType), "VK_IMAGE_VIEW_TYPE_")
-	format := strings.TrimPrefix(fmt.Sprintf("%v", a.format), "VK_FORMAT_")
-	return fmt.Sprintf("0x%X%s%s %vx%vx%v load:%s store:%s %v %v", a.handle, isSwapChain, transient, a.width, a.height, a.depth, load, store, imgType, format)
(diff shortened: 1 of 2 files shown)


hevrard:
This PR leads to the following framegraph on SW "deferred" demo:

framegraph

johnplate (approved):
LGTM, but again wait for another approval.

gapis/api/vulkan/framegraph.go
```diff
@@ -69,17 +99,14 @@ func newAttachmentInfo(desc VkAttachmentDescription, imgView ImageViewObjectʳ,
 		loadOp = desc.LoadOp()
 		storeOp = desc.StoreOp()
 	}
+	imgObj := imgView.Image()
```
yalcinmelihyasin, line 102: Very nit: You probably don't need the variable.

gapis/api/vulkan/framegraph.go
```diff
@@ -166,6 +212,45 @@ type framegraphInfoHelpers struct {
 	rpInfos  []*renderpassInfo
 	rpInfo   *renderpassInfo
 	currRpId uint64
+	// imageLookup is a lookup table to quickly find images that match a memory
```
yalcinmelihyasin, line 215: Very nit: I think that writing "memoization table" can save a lot of word.

gapis/api/vulkan/framegraph.go
```diff
@@ -166,6 +212,45 @@ type framegraphInfoHelpers struct {
...
+	// updated when a new renderpass is started under a new top-level parent
+	// command: images can only be created/destroyed between top-level commands,
+	// not during the execution of subcommands.
+	imageLookup  map[memory.PoolID]map[memory.Range][]*ImageObjectʳ
+	parentCmdIdx uint64 // used to know when to update imageLookup
+}
+
+// updateImageLookup updates the lookup table to quickly find an image matching a memory observation.
+func (helpers *framegraphInfoHelpers) updateImageLookup(state *State) {
+	helpers.imageLookup = make(map[memory.PoolID]map[memory.Range][]*ImageObjectʳ)
```
yalcinmelihyasin, line 227: This looks like a very deeply accessed information that can cause consecutive cache misses(e.g accessing the elements of two layered map). If there is no alternative, then it's fine but otherwise, for the future, this structure can be reconsidered.

yalcinmelihyasin: This of course depends how it's being read.

gapis/api/vulkan/framegraph.go
```diff
@@ -166,6 +212,45 @@ type framegraphInfoHelpers struct {
...
+				}
+			}
+		}
+	}
+}
+
+// lookupImages returns all the images that contain (pool, memRange).
+func (helpers *framegraphInfoHelpers) lookupImages(pool memory.PoolID, memRange memory.Range) []*ImageObjectʳ {
+	images := []*ImageObjectʳ{}
+	for imgRange := range helpers.imageLookup[pool] {
```
yalcinmelihyasin, line 248: If we are looping over a map to find if a range is included in a memory range, maybe we should store them as array under a pool instead of a map. Because if we are not using the direct access of the map(and I believe these are hash-tables) then we should not pay the cost for non-consecutive memory access. In a large enough data set, it should be much faster to loop over an array than a map.

Unless, there will be another access to the memory range that will be added in the future that queries the ranges with directly accessing to the map elements.

gapis/api/vulkan/framegraph.go
```diff
@@ -166,6 +212,45 @@ type framegraphInfoHelpers struct {
...
+		for _, aspect := range image.Aspects().All() {
+			for _, layer := range aspect.Layers().All() {
+				for _, level := range layer.Levels().All() {
+					pool := level.Data().Pool()
+					memRange := level.Data().Range()
+					if _, ok := helpers.imageLookup[pool]; !ok {
+						helpers.imageLookup[pool] = make(map[memory.Range][]*ImageObjectʳ)
+					}
+					imgObj := image
+					helpers.imageLookup[pool][memRange] = append(helpers.imageLookup[pool][memRange], &imgObj)
```
yalcinmelihyasin, line 238: We haven't check if the range is already existing or any existing range includes the current range. If I understand correctly, what happens here behind the scene is;

- check if memRange exist for the map stored imageLookup[pool] 
- if exists add &imgObj to the slice stored in imageLookup[pool][memRange]
- if not, then append imgObj to a nil slice(which creates it) then assign that slice to imageLookup[pool][memRange] which again allocates some stuff on the map.

If this is the intended behaviour then it's ok. Otherwise, please reconsider.

gapis/api/vulkan/framegraph.go
```diff
@@ -166,6 +212,45 @@ type framegraphInfoHelpers struct {
...
+		}
+	}
+}
+
+// lookupImages returns all the images that contain (pool, memRange).
+func (helpers *framegraphInfoHelpers) lookupImages(pool memory.PoolID, memRange memory.Range) []*ImageObjectʳ {
+	images := []*ImageObjectʳ{}
+	for imgRange := range helpers.imageLookup[pool] {
+		if imgRange.Includes(memRange) {
+			images = append(images, helpers.imageLookup[pool][imgRange]...)
```
yalcinmelihyasin, line 250: can one memRange included by different memRanges in the map?

gapis/api/vulkan/framegraph.go
```diff
@@ -209,6 +302,28 @@ func (helpers *framegraphInfoHelpers) processSubCommand(ctx context.Context, dep
 	if helpers.rpInfo != nil {
 		nodeID := dependencyGraph.GetCmdNodeID(api.CmdID(subCmdIdx[0]), subCmdIdx[1:])
 		helpers.rpInfo.nodes = append(helpers.rpInfo.nodes, nodeID)
+
+		for _, memAccess := range dependencyGraph.GetNodeAccesses(nodeID).MemoryAccesses {
```
yalcinmelihyasin, line 306: You may want to cache this unless you are modifying it in the loop. I am not sure, double check this, but go may try to call dependencyGraph.GetNodeAccesses(nodeID) in every cycle in the loop.

yalcinmelihyasin (commented):
I don't have a very deep understanding of how memory ranges works that is used in this PR. Therefore some comments that I made can be not perfectly valid.

There are a few things that I wanted to point out and if those questions are known or not an issue, then it's OK to merge this PR with LGTM from another person.

PS: Nit & Very Nit are optional.

gapis/api/vulkan/framegraph.go
hevrard: Turns out you do, because Go won't let you do `&(imgView.Image())` (hence this new variable here).

gapis/api/vulkan/framegraph.go
hevrard: Memoization is more about lazily computing results, storing them, and then lookup them up next time they are queried. I think lookup table is more accurate here since we populate the entire table upfront, and then we always use it for lookups. I also feel that it's semantics details at this point :)

gapis/api/vulkan/framegraph.go
hevrard: Go will evaluate it only once, the Go spec says (https://golang.org/ref/spec#For_statements , section For statements with range clause):

> The range expression x is evaluated once before beginning the loop, with one exception: if at most one iteration variable is present and len(x) is constant, the range expression is not evaluated.

gapis/api/vulkan/framegraph.go
hevrard: Here in theory there could be several imgRange that could include memRange, yes. Maybe not in practice, but it doesn't hurt to check them all, since in practice there are almost always only one range associated to a poolID (so the loop on `helpers.imageLookup[pool]` will have only one iteration).

gapis/api/vulkan/framegraph.go
hevrard: In practice, given how memory accesses are recorded at capture time, there is almost always only one memory range associated to a memory pool. In the lookup, we know the poolID so we directly access to the second-level map, then we iterate over it and it's not a big deal since it will almost always have only one element.

gapis/api/vulkan/framegraph.go
hevrard: I use a second-level map indexed by ranges in order to avoid storing duplicate ranges. Again, in practice for each poolID we almost always have only a single range anyway.

gapis/api/vulkan/framegraph.go
hevrard: You got the idea, we append to helpers.imageLookup[pool][memRange], which may be nil.

This seems to be the classic way to add items to a slice member of a map, quoting https://blog.golang.org/maps:

> Another instance of helpful zero values is a map of slices. Appending to a nil slice just allocates a new slice, so it's a one-liner to append a value to a map of slices; there's no need to check if the key exists.

PR #552, opened 2020-11-11
Tab close button & Tab bar overflow fix

johnplate:
This adds close buttons to tabs in the folder bars, and wraps the tabs into multiple rows if they don't fit in the given bar width. An expand icon indicates the existence of more than one row. These rows appear when hovering over the expand icon, and disappear when leaving the expanded tab bar. This also works when dragging tabs between folders.
The close and expand icons are from [material.io](https://material.io/resources/icons/?search=clear&icon=clear&style=baseline).

b/170298055 & second PR for b/170298289


pmuetschard (commented):
- use the clear icon from material (https://material.io/resources/icons/?search=clear&icon=clear&style=baseline) it has a thinner stroke and is not as blurry.
- don't do the box thing, if you want to highlight the close icon on hover, do it like Chrome and change the background. VSCode doesn't even bother with a highlight, which I think is a good starting point.
- allocate space for the close icon in each tab, rather than drawing the close icon over the tab name. If you want to show/hide the close icon on non selected tabs (ala VSCode) that's OK, but I'd also be happy with just always showing the close buttons like Chrome.
- undo the tightening of the margin (10->6 points). The tabs are too cramped now. You can have the margin shrink as we run out of space if you'd like, however,...
- ... the shrinking doesn't truly fix the not-enough space problem. I'd prefer the approach seen in other tools with an overflow drop-down menu. Although, I'm Ok to leave that for later.

johnplate:
> * use the clear icon from material (https://material.io/resources/icons/?search=clear&icon=clear&style=baseline) it has a thinner stroke and is not as blurry.
> * don't do the box thing, if you want to highlight the close icon on hover, do it like Chrome and change the background. VSCode doesn't even bother with a highlight, which I think is a good starting point.
> * allocate space for the close icon in each tab, rather than drawing the close icon over the tab name. If you want to show/hide the close icon on non selected tabs (ala VSCode) that's OK, but I'd also be happy with just always showing the close buttons like Chrome.

I am using the mentioned clear icon now, allocate space for it in the active tab, show it always and change the background when hovering, consistent with the rest of our UI.
I still prefer to save horizontal space by activating the close icon only for the active tab. I don't think the users will frequently close main tabs, and if they close pinned tabs, it will be good to see what you close as the titles will be the same. If we show close icons for all tabs we would later also show all pin icons, which would further increase the bar width.

> * undo the tightening of the margin (10->6 points). The tabs are too cramped now. You can have the margin shrink as we run out of space if you'd like, however,...
> * ... the shrinking doesn't truly fix the not-enough space problem. I'd prefer the approach seen in other tools with an overflow drop-down menu. Although, I'm Ok to leave that for later.

Yes, the shrinking wasn't the ideal solution. Many other solutions would be tricky with the tab dragging feature. I have now implemented a multi-row drop-down tab bar, see the updated PR description.

pmuetschard:
I like the drop-down overflow.

I don't like the tabs changing size when they are selected. I got into the situation where selecting the last visible tab, made it disappear, as after selecting it, it is too big and went into the overflow. This feels odd. Allocate the space for the X for each tab, even if it's not visible for non-selected tabs.

I also think the blue background looks odd, when it cuts off just before the X, due to the blue line underneath. I think the highlight should highlight the entire tab, and when the X is hovered, shrink to just the X. Something like this:
image
I also notice that the tabs bounce: if there are five tabs, and I switch between the first two, the last 3 tabs change position. The tabs should stay stationary.

I think the X is too big, I think I would prefer the 16px version.

We should also think how we want to change it when selecting a tab from the overflow. As it is implemented right now, the tab bar will not show the title of the currently selected tab.

johnplate:
> I like the drop-down overflow.

Thank you!

> I don't like the tabs changing size when they are selected. I got into the situation where selecting the last visible tab, made it disappear, as after selecting it, it is too big and went into the overflow. This feels odd. Allocate the space for the X for each tab, even if it's not visible for non-selected tabs.

I also don't like it, but I still prefer to save horizontal space. Don't forget that more tabs and more icons are coming. I have solved the mentioned problem by reserving enough space for all tabs in a row, so no tab will move into overflow when any in a row is selected.

> I also think the blue background looks odd, when it cuts off just before the X, due to the blue line underneath. I think the highlight should highlight the entire tab, and when the X is hovered, shrink to just the X.

Sure, done.

> I also notice that the tabs bounce: if there are five tabs, and I switch between the first two, the last 3 tabs change position. The tabs should stay stationary.

This problem is based on the fact that the bold font of selected tabs increases the label width. If I keep the tab width fixed, we get too big margins on the right for all deselected tabs, which looks very ugly especially for long titles. Unless we change the font I prefer the dynamic tab positions over the ugly margins.

> I think the X is too big, I think I would prefer the 16px version.

Sure, done.

> We should also think how we want to change it when selecting a tab from the overflow. As it is implemented right now, the tab bar will not show the title of the currently selected tab.

I thought about it before and didn't like the concept of other tabs who rearrange the tab order to keep the selected visible. I would like to preserve the tab order, so I thought about it again. I changed the drop down concept to an expanded bar concept. It is very similar, but now the collapsed title bar will show the tab row which contains the active tab. When expanding the bar, the collapsed row will stay where it is and the additional rows will appear on top and bottom. The expand icon will show the direction of expansion, up, down, or both.

I also fixed the performance problem that the tab bar was redrawn on each mouse move. It will now only be redrawn when the hover object changes.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
```diff
@@ -1091,18 +1376,51 @@ private Hover(Type type, Group group, int index, int cursor, Folder folder, Tab
...
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null) {
+        return false;
+      }
+      if (getClass() != o.getClass()) {
+        return false;
+      }
```
pmuetschard, line 1389:
```suggestion
      if (this == o) {
        return true;
      } else if (!(o instanceof Hover)) {
        return false;
      }
```

That's the pattern we've been using everywhere else.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
```diff
@@ -1211,4 +1530,8 @@ public void close() {
       shell.dispose();
     }
   }
+
+  public interface TabManager {
+    public void closeTab(Object id);
```
pmuetschard, line 1535: Instead, add `onTabClosed(TabInfo tab)` in the existing `Listener`.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
```diff
@@ -149,39 +157,58 @@ protected boolean flushCache(Control control) {
...
+        switch (mouseDown.type) {
+          case Tab:
+            if (mouseDown.folder.updateCurrent(mouseDown.tab.control)) {
+              listeners.fire().onTabShown(mouseDown.tab.info);
+            }
+            break;
+          default:
+            // Do nothing.
+        }
+      }
```
pmuetschard, line 175: Why was this moved out of the `MouseDown` handler? I think it makes more sense there, especially for the immediate mouse-down-drag on a non selected tab. I think it's fine to move it back, I don't see any code reason for it to have to be moved here.

pmuetschard (commented):
1) I think it would be better for the expand icons to stay visible when the expand bar is shown.
2) I have a slight preference towards not showing the close button on hovers. I find it too easy to close a tab when switching tabs. So I think we should show the close button only on the currently selected tab, but show it always.

Both of the above are not needed right now. We can get more feedback from the rest of the team, but please do address the code comments.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
johnplate: Sure, done

gapic/src/main/com/google/gapid/widgets/TabComposite.java
johnplate: Thanks, done, I also prefer that solution. I also added onTabCreated to be consistent.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
johnplate: This was actually a fix for non-selected tabs. When MouseDown is received, it is not clear yet if the user wants to select or drag the tab, so the select is postponed to MouseUp if no drag happened. The current implementation selects the tab first before it is dragged, which doesn't allow to maintain the selected tab if the user wants to drag a non-selected tab. All other tab UIs that I tested keep the selected tab if a non-selected tab is dragged.

johnplate:
> 1. I think it would be better for the expand icons to stay visible when the expand bar is shown.

If the expand icon would stay visible, the user would intuitively expect that something happens if you click on it, for example that the expanded bar collapses again. So I think it is better to hide it.

> 2. I have a slight preference towards not showing the close button on hovers. I find it too easy to close a tab when switching tabs. So I think we should show the close button only on the currently selected tab, but show it always.

The close button was intended to be always visible on the selected tab, this is fixed.
I have a preference to show the close buttons when hovering over non-selected tabs, as this is done in several other tab UIs. I don't think it can easily happen to close a tab when selecting it as users will typically click on the titles and not the empty space between titles. We can wait for some feedback on this potential issue.

> Both of the above are not needed right now. We can get more feedback from the rest of the team, but please do address the code comments.

Agreed, done.

gapic/src/main/com/google/gapid/GraphicsTraceView.java
```diff
@@ -104,9 +108,28 @@ public void store(TabArea.FolderInfo[] folders) {
 
     tabs.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
 
+    @SuppressWarnings("synthetic-access")
```
pmuetschard, line 111: Just make the `syncTabMenuItem` method protected instead of ignoring the warning.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
pmuetschard: I disagree. Both Chrome and VSCode activate tabs on mouse down not mouse up. I remember now that I did it this way on purpose. Please restore the original functionality.

pmuetschard:
> I think it would be better for the expand icons to stay visible when the expand bar is shown.
>> If the expand icon would stay visible, the user would intuitively expect that something happens if you click on it, for example that the expanded bar collapses again. So I think it is better to hide it.

I don't think so. It's very strange for an interactive element to disappear when you interact with it. Buttons don't disappear when you click them, nor have I ever seen anyone make something hovered disappear.

gapic/src/main/com/google/gapid/GraphicsTraceView.java
johnplate: Sure, done.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
johnplate: I didn't say that all other UIs activate tabs on mouse up, even though there are some, for example Eclipse. I said that in all other UIs after dragging a non-selected tab to a destination folder, the selected tab in the source folder will be the same as before. This is also what a user intuitively would expect and this is not happening in AGI.
As requested I have restored the original functionality and filed a bug. b/174851718

johnplate:
> > I think it would be better for the expand icons to stay visible when the expand bar is shown.
> > > If the expand icon would stay visible, the user would intuitively expect that something happens if you click on it, for example that the expanded bar collapses again. So I think it is better to hide it.
> I don't think so. It's very strange for an interactive element to disappear when you interact with it. Buttons don't disappear when you click them, nor have I ever seen anyone make something hovered disappear.

I agree that it is unusual that an interactive element disappears, but why is it very strange? In this case the button disappears at the same time when the tab bar expands and reappears when the bar collapses, so the user should easily and intuitively understand what is happening. I think it would be more confusing if we keep showing an interactive element which doesn't work anymore.

pmuetschard (approved):

PR #560, opened 2020-11-17
Fix propagation of replay fatal error notifications

hevrard:
The replayer can send replay notifications to GAPIS, and the handling
of these notification can be customized for different kinds of
replays (e.g. report replays collect validation errors this way).

However, when a fatal error message notification is sent by the
replayer (e.g. because it has failed to create a Vulkan instance),
GAPIS should always treat this as an error, independently from the
custom notification handling.

In particular, this caused the frame profiling to hang when the replay
failed due to not being able to create a Vulkan instance.

Bug: b/172447447, b/173510056
Test: manual

### DIFF
--- a/gapis/replay/executor.go
+++ b/gapis/replay/executor.go
@@ -117,7 +117,7 @@ func (e executor) HandleNotification(ctx context.Context, notification *gapir.No
 	// However notifications are handled, never survive a fatal error
 	errMsg := notification.GetErrorMsg()
 	if errMsg != nil && errMsg.Severity == severity.Severity_FatalLevel {
-		return log.Errf(ctx, fmt.Errorf("%s", errMsg.Msg), "Received a fatal error during replay")
+		return fmt.Errorf("Replay failed with a fatal error notification: %v", errMsg.Msg)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


gapis/replay/executor.go
```diff
@@ -112,6 +114,11 @@ func (e executor) HandlePostData(ctx context.Context, postData *gapir.PostData)
 // HandleNotification implements gapir.ReplayResponseHandler interface.
 func (e executor) HandleNotification(ctx context.Context, notification *gapir.Notification) error {
 	e.handleNotification(notification)
+	// However notifications are handled, never survive a fatal error
+	errMsg := notification.GetErrorMsg()
+	if errMsg != nil && errMsg.Severity == severity.Severity_FatalLevel {
+		return log.Errf(ctx, fmt.Errorf("%s", errMsg.Msg), "Received a fatal error during replay")
```
pmuetschard, line 120: `errors.New(errorMsg.Msg)` if you want to wrap the string into an error. However, that doesn't make much sense here, so why not just `return fmt.Errorf("Replay failed: %v", errMsg.Msg)`?

hevrard: Done. I felt like we had to use log.Errf() to pass the `ctx` context here.

johnplate (approved):
LGTM now with the fix for Pascal's comment.

PR #562, opened 2020-11-18
Drop debug utils extension at replay time

hevrard:
At replay time, if Vulkan instance creation fail and we detect that
the validation layer or the debug report extension is requested, we
have a fallback where we try to create the instance again after
dropping the validation layer and the debug report extension. This
change adds the debug utils extension alongside the debug report
extension.

Note that both of these extensions are provided by the validation
layer. The typical failure is: the app requests the validation layer
at trace time, and one or both of the debug extensions. At replay
time, unless we do a report replay, the validation layer won't be
loaded, hence these extensions are not present.

More generally, we won't support replaying a workload that relies on
extensions provided by a layer, since this layer won't be present at
replay time. The validation layer is a special case since it is the
one layer we may provide at replay time (as it is present in our
APKs). See discussion at b/163754399

Bug: b/172447447
Test: manual

### DIFF
--- a/gapir/cc/context.cpp
+++ b/gapir/cc/context.cpp
@@ -231,7 +231,7 @@ void Context::registerCallbacks(Interpreter* interpreter) {
           // drop them and try to create VkInstance again.
           if (Vulkan::hasValidationLayers(pCreateInfo->ppEnabledLayerNames,
                                           pCreateInfo->enabledLayerCount) ||
-              Vulkan::hasDebugReportOrDebugUtilsExtension(
+              Vulkan::hasDebugExtension(
                   pCreateInfo->ppEnabledExtensionNames,
                   pCreateInfo->enabledExtensionCount)) {
             onDebugMessage(
--- a/gapir/cc/context.cpp
+++ b/gapir/cc/context.cpp
@@ -231,13 +231,14 @@ void Context::registerCallbacks(Interpreter* interpreter) {
           // drop them and try to create VkInstance again.
           if (Vulkan::hasValidationLayers(pCreateInfo->ppEnabledLayerNames,
                                           pCreateInfo->enabledLayerCount) ||
-              Vulkan::hasDebugReportExtension(
+              Vulkan::hasDebugReportOrDebugUtilsExtension(
                   pCreateInfo->ppEnabledExtensionNames,
                   pCreateInfo->enabledExtensionCount)) {
             onDebugMessage(
                 LOG_LEVEL_WARNING, Vulkan::INDEX,
-                "Failed to create VkInstance with validation layers and "
-                "debug report extension, drop them and try again");
+                "Failed to create VkInstance with validation layers or "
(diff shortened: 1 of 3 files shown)


gapir/cc/context.cpp
```diff
@@ -231,13 +231,14 @@ void Context::registerCallbacks(Interpreter* interpreter) {
           // drop them and try to create VkInstance again.
           if (Vulkan::hasValidationLayers(pCreateInfo->ppEnabledLayerNames,
                                           pCreateInfo->enabledLayerCount) ||
-              Vulkan::hasDebugReportExtension(
+              Vulkan::hasDebugReportOrDebugUtilsExtension(
```
pmuetschard, line 234: The strategy to keep adding to this function name is not going to work out. Let's make it more general and call it `hasDebugExtension`.

gapir/cc/context.cpp
```diff
@@ -231,13 +231,14 @@ void Context::registerCallbacks(Interpreter* interpreter) {
...
                   pCreateInfo->ppEnabledExtensionNames,
                   pCreateInfo->enabledExtensionCount)) {
             onDebugMessage(
                 LOG_LEVEL_WARNING, Vulkan::INDEX,
-                "Failed to create VkInstance with validation layers and "
-                "debug report extension, drop them and try again");
+                "Failed to create VkInstance with validation layers or "
+                "debug report or debug utils extension, drop them and "
+                "try again");
```
pmuetschard, line 241:
```suggestion
                "Failed to create VkInstance with validation layers or "
                "debug extensions, dropping them and retrying");
```

gapis/api/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -107,7 +108,7 @@ bool Vulkan::replayCreateVkInstanceImpl(Stack* stack,
                                         const VkInstanceCreateInfo* pCreateInfo,
                                         VkAllocationCallbacks* pAllocator,
                                         VkInstance* pInstance,
-                                        bool dropValidationLayersAndDebugReport,
+                                        bool dropValidationLayersAndDebugReportAndDebugUtils,
```
pmuetschard, line 111: `dropValidationAndDebugLayers`?

gapir/cc/context.cpp
hevrard: Agreed! I hesitated, since debug report and debug utils seems to be the only two debug-related extensions existing so far, so I though it might be OK to stay explicit at the price of this mouthful. But `hasDebugExtension` is also fine with me.

gapis/api/templates/vulkan_gfx_api_extras.tmpl
hevrard: I went with `dropValidationAndDebug`, since it's a mix of layers and extensions.

pmuetschard (approved):

PR #563, opened 2020-11-19
Fix broken command tree structure when a renderpass contains subpasses.

stellama0208:
- There are two main missing details resulting in this bug:
 - 1. When renderpass begins, two command grouping markers get pushed
 into the grouping stack; but when renderpass ends, the grouping stack
 only pops once. Hence the stack will get unbalanced after meeting a
 renderpass with subpasses and later introduce wrong nesting result.
 - 2. Two adjecent subpasses fight to include the same vkCmdNextSubpass
 command and thus the command grouping gets confused, this PR reduce the
 ending command index by 1 for the earlier subpass, so that the
 contention disappear and the tree grouping will not break.

 Bug: b/173682026.
 Test: Could test with the trace in the bug ticket, and observe the
 command tree grouping.


hevrard (approved):
:+1:

hevrard:
The macOS CI failure seems to be a flake.

johnplate (approved):

PR #566, opened 2020-11-19
Add a stack dumping handler for SIGQUIT.

pmuetschard:
Any app built using our core/app library will register a signal handler for SIGQUIT (kill -3) that will dump the stack trace of all 
goroutines to stdout.

### DIFF
--- a/core/app/stack.go
+++ b/core/app/stack.go
@@ -1,0 +1,42 @@
+// Copyright (C) 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


core/app/stack.go
```diff
@@ -0,0 +1,42 @@
...
+	"os/signal"
+	"runtime"
+	"syscall"
+
+	"github.com/google/gapid/core/app/crash"
+)
+
+// Adds a signal handler for SIGQUIT to dump all go-routine stacks.
+func init() {
+	sigchan := make(chan os.Signal)
```
hevrard, line 29: According to the os/signal pkg doc, this channel should be buffered with size 1.
https://golang.org/pkg/os/signal/#Notify

> Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient.

```suggestion
	sigchan := make(chan os.Signal, 1)
```

pmuetschard: I don't really care if we miss any of these signals, but ok.

hevrard (approved):

PR #568, opened 2020-11-19
Adds a (hidden) flag that makes replay device matching lenient.

pmuetschard:
If the flag is set, the device matching ignores the driver and API version when checking for replay device compatibility. Using this flag is unsupported, but allows opening traces when driver version mismatch.

### DIFF
--- a/gapis/api/vulkan/replay.go
+++ b/gapis/api/vulkan/replay.go
@@ -16,6 +16,7 @@ package vulkan
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/replay.go
```diff
@@ -34,6 +35,10 @@ import (
 	"github.com/google/gapid/gapis/stringtable"
 )
 
+var (
+	lenientDeviceMatching = flag.Bool("lenient-device-matching", false, "_Makes replay device matching more lenient")
+)
+
```
hevrard, line 41: Shouldn't this be in `cmd/gapis/main.go` ?

hevrard (approved):
LGTM with a doubt on where the flag definition should be

gapis/api/vulkan/replay.go
pmuetschard: No. Think about the library dependencies.

PR #570, opened 2020-11-22
Updated glslang lib and building doc for gcc-10 on Windows

johnplate:
If you follow the AGI build doc for Windows currently, building gcc_wrapper ends up with internal compiler error. My guess is that gcc 9.2.0 doesn't like that MSYS2 installs gcc-libs 10.2.0, which is a requirement for many other packages. A workaround could be to require developers to maintain an outdated version of MSYS2, which would be a burden if they use MSYS2 for anything else than building AGI. The better solution is to compile AGI with gcc-10, which coincidentally also fixes the crash in Perfetto when building AGI not optimized (this is not fixed by the recent Perfetto update!).

This PR updates the glslang lib to a version which includes a [commit](https://github.com/KhronosGroup/glslang/commit/8486b2e0db150a3e0761db6930a7b41cc7a8c83f#diff-8bfb66f7619ec31ff49b6934cb5cca06b5a459852c6f6d3562d065ea4000de22) that fixes a compile error with gcc-10, and it updates the building doc to use gcc-10 for Windows.

It also fixes a minor wrong indent.

The compiler used by the Windows CI should also be updated to gcc-10 once this PR is approved.

Fixes b/173712032 and b/162405255.

### DIFF
--- a/tools/build/workspace.bzl
+++ b/tools/build/workspace.bzl
@@ -145,8 +145,8 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
         locals = locals,
         organization = "KhronosGroup",
         project = "glslang",
-        commit = "3ee5f2f1d3316e228916788b300d786bb574d337",  # SDK-candidate-26-Jul-2020
-        sha256 = "8f4ebd7ff348a7f752e4646e43d4e41b4747323d635f5ae49096771d5c64fb31",
+        commit = "740ae9f60b009196662bad811924788cee56133a",  # 10-11.0.0
+        sha256 = "c015e7d81c0a248562c25a1e484fb8528eb6a765312cf5de3bdb658b03562b3f",
     )
 
     maybe_repository(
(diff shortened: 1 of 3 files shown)


hevrard:
Thanks for fixing this!

Why not update to gcc-10 on windows CI in this same PR?

hevrard:
OK I see you have https://github.com/google/agi/pull/571 in progress to update to gcc-10 on the CI.

hevrard (approved):

tools/build/workspace.bzl
```diff
@@ -145,8 +145,8 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
         locals = locals,
         organization = "KhronosGroup",
         project = "glslang",
-        commit = "8db9eccc0baf30c9d22c496ab28db0fe1e4e97c5",  # 8.13.3559
-        sha256 = "5c11a228d41ec011918b9c8beb60b6556745d30c8c856ec622beab5c5469152d",
+        commit = "3ee5f2f1d3316e228916788b300d786bb574d337",  # SDK-candidate-26-Jul-2020
```
pmuetschard, line 148: Let's go with https://github.com/KhronosGroup/glslang/releases/tag/10-11.0.0, which is marked as a stable release.

pmuetschard (approved):

tools/build/workspace.bzl
johnplate: Sure, done

PR #573, opened 2020-11-23
Framegraph: add compute workloads

hevrard:
We used to collect information only for renderpasses, this change adds
information gathering for compute workloads as well.

The key change is the renderpassInfo type becoming workloadInfo: it
now stores info that are common to a renderpass and a
compute (e.g. image and buffer accesses), and has a pair of pointer
fields to store renderpass-specific or compute-specific info. The
rpInfo field of helpers is accordingly renamed wlInfo, which leads to
a lot of small-and-trivial-renaming diffs.

Two other noticeable changes:

- In processSubCommand, the detection of renderpass start/end used to
  rely on catching vkCmdBeginRenderpass and vkCmdEndRenderpass: this
  is now a bit more complex since we need to also detect the start/end
  of compute workloads, which can be any of `vkCmdDispatch*`.

- We now list image usage bits as separate booleans in the image
  proto, this mirrors what we do for buffers and leads to a more
  homogeneous representation of images and buffers in framegraph data.
  This effectively replaces the FramegraphImageNature message.

Bug: b/173007235
Test: manual

### DIFF
--- a/gapis/api/vulkan/framegraph.go
+++ b/gapis/api/vulkan/framegraph.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/google/gapid/core/image"
+	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/gapis/api"
 	"github.com/google/gapid/gapis/api/sync"
 	"github.com/google/gapid/gapis/capture"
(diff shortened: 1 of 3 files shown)


gapis/api/vulkan/framegraph.go
```diff
@@ -461,12 +593,12 @@ func (API) GetFramegraph(ctx context.Context, p *path.Capture) (*api.Framegraph,
...
-func updateDependencies(rpInfos []*renderpassInfo, dependencyGraph dependencygraph2.DependencyGraph) {
+func updateDependencies(workloadInfos []*workloadInfo, dependencyGraph dependencygraph2.DependencyGraph) {
 	// isInsideRenderpass: node -> renderpass it belongs to.
 	isInsideRenderpass := map[dependencygraph2.NodeID]uint64{}
-	for _, rpInfo := range rpInfos {
-		for _, n := range rpInfo.nodes {
-			isInsideRenderpass[n] = rpInfo.id
+	for _, wlInfo := range workloadInfos {
+		for _, n := range wlInfo.nodes {
+			isInsideRenderpass[n] = wlInfo.id
```
stellama0208, line 601: I'm wondering whether this part still works as expected? After the nodes becomes more general and not only being renderpass node?

stellama0208 (commented):
The changes looks good to me on the whole, but not tested, and may need another pair of eyes that's more familiar with this part of code.

gapis/api/vulkan/framegraph.go
hevrard: Good catch, thanks! The logic is still valid, but the variable names and comments do no match anymore, indeed. I've update them to reflect that this now operates on workloads (and not only renderpasses).

johnplate (approved):
LGTM, but as Stella I'm also not very familiar with the code.

PR #575, opened 2020-11-24
Remove the Go memory arena.

pmuetschard:
It was never completely used (i.e. no calls to `Allocate`, etc) and was part of the now abandoned compiler effort.


hevrard (approved):
LGTM! That was a big cleanup, thanks for making it happen!

PR #577, opened 2020-11-26
Only show command nodes with performance values in perftab.

stellama0208:
- A step to decouple the performance tree from the command tree, and to
 help make full use of the spaces inside the perftab by reducing blank area.
 - Implementation changes:
   1. Complete the GPU slices grouping logic at the server side,
   create the tree structure between the groups.
   2. When creating the performance tree at the client side, stop
   reusing the command tree's class, instead use a new tree that
   implements only the needed functions.
   3. Make sure the linking behavior between Profile, Command, and
   Performance tab doesn't regress.
 - Bug: b/169169693.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/PerformanceView.java
+++ b/gapic/src/main/com/google/gapid/views/PerformanceView.java
@@ -16,23 +16,28 @@
 package com.google.gapid.views;
 
 import static com.google.gapid.util.Loadable.MessageType.Error;
+import static com.google.gapid.widgets.Widgets.createTreeColumn;
+import static com.google.gapid.widgets.Widgets.createTreeViewer;
+import static com.google.gapid.widgets.Widgets.packColumns;
 
-import com.google.gapid.models.Analytics.View;
 import com.google.gapid.models.Capture;
 import com.google.gapid.models.CommandStream;
-import com.google.gapid.models.CommandStream.CommandIndex;
 import com.google.gapid.models.Models;
 import com.google.gapid.models.Profile;
+import com.google.gapid.models.Profile.PerfNode;
 import com.google.gapid.perfetto.Unit;
 import com.google.gapid.perfetto.models.CounterInfo;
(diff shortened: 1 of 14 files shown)


gapis/trace/android/adreno/profiling_data.go
```diff
@@ -294,3 +305,35 @@ func processCounters(ctx context.Context, processor *perfetto.Processor, desc *d
 	}
 	return counters, nil
 }
+
+// For a renderPass leafy group, find its parent group and return it.
+// If the parent groups are not created yet, create them and store in the map.
+func findParentGroup(ctx context.Context, subOrder, cb uint64, groupsMap map[api.CmdSubmissionKey]*service.ProfilingData_GpuSlices_Group, links map[api.CmdSubmissionKey][]api.SubCmdIdx, capture *path.Capture) *service.ProfilingData_GpuSlices_Group {
```
hevrard, line 311: This logic seems to be duplicated between adreno and mali, would there be a way to asbtract it in a single place to avoid duplication?

gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -146,78 +166,86 @@ public void onProfileLoaded(Loadable.Message error) {
...
+      TreeItem result = findItem(root, requirement);
+      if (result != null) {
+        return result;
+      }
     }
+    return null;
+  }
+
+  private PerfNode cast(Object o) {
+    return (PerfNode)o;
```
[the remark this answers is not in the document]
stellama0208: Yeah, I've removed the cast function, it's not used that much anyway. :D

gapis/trace/android/adreno/profiling_data.go
stellama0208: Thanks for pointing it out, just done it in a new commit!

hevrard (approved):
LGTM

PR #580, opened 2020-11-30
Avoid GAPIS panics when mutate context is cancelled

### LINKED ISSUE #59: panicOnError() does not behave properly with  context cancellation
GAPIS can arbitrarily crash when a context is cancelled, if it happens in some api.go functions.

hevrard:
If the context gets cancelled during capture command mutation,
panicOnError() may be triggered and leads to a GAPIS panic. This
change catches such panics and transform them to regular errors.

Another option would have been to get rid of panicOnError and
propagate the errors across all generated code inside Mutate
functions: I had a quick try at this, and it gets quite
cumbersome, especially things like:

```
foo := bar(..., MustRead(...), ...)
```

must be turned into:

```
tmp, err := MustRead(...)
if err != nil {
  return nil, err
}
foo := bar(..., tmp, ...)
```

Using panic for control flow is not awesome, but it is very convenient
here to jump from possibly deep-nested calls in generated code up to
the top-level of Mutate.

Fix #59
Bug: b/149701723
Test: manual, by hard-coding panics and errors

### DIFF
--- a/core/event/task/doc.go
+++ b/core/event/task/doc.go
@@ -13,9 +13,4 @@
 // limitations under the License.
 
 // Package task contains types that can be used to create cancelable tasks.
-//
-// This package builds on top of Go's standard "context" package. Before
-// touching this package internals, make sure to understand the standard Go
-// context. This article gives a good introduction:
-// https://blog.golang.org/context
 package task
--- a/core/event/task/doc.go
+++ b/core/event/task/doc.go
@@ -13,4 +13,9 @@
 // limitations under the License.
 
 // Package task contains types that can be used to create cancelable tasks.
+//
+// This package builds on top of Go's standard "context" package. Before
+// touching this package internals, make sure to understand the standard Go
+// context. This article gives a good introduction:
+// https://blog.golang.org/context
 package task
--- a/gapis/api/templates/api.go.tmpl
+++ b/gapis/api/templates/api.go.tmpl
@@ -219,6 +219,8 @@ var ( // Don't error if these packages aren't used.
   {{template "Go.GeneratedHeader" (Global "OutputDir")}}
 
 import (
+    "context"
+    "errors"
     "github.com/google/gapid/core/data/id"
     ϟapi "github.com/google/gapid/gapis/api"
 )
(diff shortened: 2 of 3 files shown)


pmuetschard (commented):
Instead of doing this in every mutate, can we do this at the higher levels?

hevrard:
> Instead of doing this in every mutate, can we do this at the higher levels?

I find this more "risky" as developers will have to know that whenever they call Mutate, they have to catch panics and filter some errors.

Are you concerned by performance with the current approach?

AWoloszyn:
This type of change deserves a performance anlysis as well. I dont know what golang's panic cost is, but it may be enough to warrant another approach.

Also this does not fix EVERY case for this error, there are still hand-written Mutates that exist, as well as this being able to happen inside of transform code /etc.
So this only reduces the frequency the crashes.

hevrard:
I've just pushed a commit that tries a new alternative: filter out task cancellation in panicOnError (hence renamed panicOnFatalError), and rely on having mutate returning `task.StopReason()` to catch task cancellations.

@pmuetschard @AWoloszyn how does that sound?

AWoloszyn:
> I've just pushed a commit that tries a new alternative: filter out task cancellation in panicOnError (hence renamed panicOnFatalError), and rely on having mutate returning `task.StopReason()` to catch task cancellations.
> 
> @pmuetschard @AWoloszyn how does that sound?

That defeats the point of "*Must*Read". Now the return from such a MustRead is potentially partially uninitialized data. We will then use that potentially uninitialized data to do things in the mutate. 

Concrete example, if we have a struct with a pNext/sType, we will switch on the stype.
When we go and try and switch on that sType, we will hit the default case.
A switch without a default in Go looks like:

```
switch(....) { 
   case .....:
   default:
      panic(fmt.Errorf("Unmatched switch()")
}
```

MustRead really DOES expect the value that is returned to be valid.

AWoloszyn:
To be clear, the problem here is that the database does *NOT* return valid bytes if it the context was cancelled. This *is* a fatal error in that we cannot continue processing in any "legal" way.

hevrard:
@AWoloszyn oh yeah I see that's not solving the problem, thanks for the clarification.

I reverted this commit.

So among the possible solutions we have:
- propagate error and do error checking across all generated code: this is painful and probably has a significant overhead
- catch and silence panics due to task cancellation:
  - systematically at all RPC entry points in `gapis/server/grpc.go` (as suggested by Andrew)
  - on a case-by-case basis wherever in Mutate and MustRead/MustWrite callers (as hinted by Pascal)

I would prefer doing systematic catching at the RPC entry points, since this let us not have to think about this whenever we use Mutate or MustRead/MustWrite.

@pmuetschard what do you think?

AWoloszyn:
I was thinking about this a bit earlier, and it's not JUST RPC entries. You would also have to do probably propagate the panic through database resolves.

hevrard:
Superseded by #639

PR #582, opened 2020-12-02
Spirv reflect cleanup

hevrard:
See the 3 commits:
- add some log
- remove unused function
- fix existing function

Bug: n/a


johnplate (approved):
LGTM, but maybe better to have an additional review

pmuetschard (approved):

PR #585, opened 2020-12-04
Framegraph: add coherent/mapped memory info

hevrard:
Buffers and images are backed by memory that may be coherent with the
host, and may be mapped at the time of their use. This change surfaces
this information in the framegraph's buffer and image types.

This turns out to be a rather big change because of some type
updates. The helpers lookup tables used to store references to Vulkan
BufferObject / ImageObject, from which we created framegraphBuffer /
framegraphImage later on. However, these objects do not store whether
the memory is coherent, for this we need to refer to the vulkan state
to access physical devices memory types. Therefore, this change
updates the lookup tables to store api.FramegraphBuffer /
api.FramegraphImage directly, where we can store the coherent memory
info. This info is retrievable when we update the lookup tables since
we have access to the vulkan state at this point.

A side-effect of storing api.FramegraphBuffer/api.FramegraphImages is
that for indexes of bufferAccesses/imageAccesses maps, we don't use
the typed VkBuffer/VkImage handles, but rather their uint64 value that
is stored in api.FramegraphBuffer/api.FramegraphImages. This also
triggers several trivial-but-noisy changes.

Also, document that readMappedCoherentMemory() results in a read-write
operation despite its name.

Bug: b/174471220
Test: manual

### DIFF
--- a/gapis/api/vulkan/framegraph.go
+++ b/gapis/api/vulkan/framegraph.go
@@ -30,29 +30,59 @@ import (
 	"github.com/google/gapid/gapis/service/path"
 )
 
-func newFramegraphBuffer(buf *BufferObjectʳ) *api.FramegraphBuffer {
+// backedByCoherentMemory returns true if the device memory object is backed by
+// memory where the VK_MEMORY_PROPERTY_HOST_COHERENT_BIT is set.
+// Note that there exists a generated subIsMemoryCoherent(), but it needs all
+// the arguments of generated functions, which are not easily retrievable here.
+func backedByCoherentMemory(state *State, mem DeviceMemoryObjectʳ) bool {
+	phyDevHandle := state.Devices().Get(mem.Device()).PhysicalDevice()
+	phyDevObj := state.PhysicalDevices().Get(phyDevHandle)
+	propFlags := phyDevObj.MemoryProperties().MemoryTypes().Get(int(mem.MemoryTypeIndex())).PropertyFlags()
+	return uint32(propFlags)&uint32(VkMemoryPropertyFlagBits_VK_MEMORY_PROPERTY_HOST_COHERENT_BIT) != 0
+}
+
+func newFramegraphBuffer(state *State, buf *BufferObjectʳ) *api.FramegraphBuffer {
 	usage := uint32(buf.Info().Usage())
 	return &api.FramegraphBuffer{
-		Handle:       uint64(buf.VulkanHandle()),
-		Size:         uint64(buf.Info().Size()),
-		Usage:        usage,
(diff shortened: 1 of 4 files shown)


gapis/api/vulkan/framegraph.go
```diff
@@ -30,29 +30,59 @@ import (
...
 
-func newFramegraphBuffer(buf *BufferObjectʳ) *api.FramegraphBuffer {
+// backedByCoherentMemory returns true if the device memory object is backed by
+// memory where the VK_MEMORY_PROPERTY_HOST_COHERENT_BIT is set.
+// Note that there exists a generated subIsMemoryCoherent(), but it needs all
+// the arguments of generated functions, which are not easily retrievable here.
+func backedByCoherentMemory(state *State, mem DeviceMemoryObjectʳ) bool {
+	phyDevHandle := state.Devices().Get(mem.Device()).PhysicalDevice()
+	phyDevObj := state.PhysicalDevices().Get(phyDevHandle)
+	propFlags := phyDevObj.MemoryProperties().MemoryTypes().Get(int(mem.MemoryTypeIndex())).PropertyFlags()
```
stellama0208, line 40: Only a little bit worried on the closely connected queries here, about whether there would be nil pointer error on edge cases. But if the auto-generated files actually handle the nil pointer well, or quite some tests have been done, I think it's fine though.

stellama0208 (approved):
LGTM! The extra info and type changes looks fine! :)

gapis/api/vulkan/framegraph.go
```diff
@@ -30,29 +30,59 @@ import (
...
 }
 
-func newFramegraphImage(img *ImageObjectʳ) *api.FramegraphImage {
+func newFramegraphImage(state *State, img *ImageObjectʳ) *api.FramegraphImage {
 	format, err := getImageFormatFromVulkanFormat(img.Info().Fmt())
 	if err != nil {
 		panic("Unrecognized Vulkan image format")
 	}
+
+	coherentMemory := false
```
AWoloszyn, line 70: So.. what happens here with sparse binding?

hevrard: Thanks for spotting this! I added tracking of sparse memory bindings.

gapis/api/vulkan/framegraph.go
hevrard: Good point. I believe here we can assume that we process valid Vulkan, so any VkDeviceMemory will have a valid physical device, where we can lookup memory properties. The generated `subIsMemoryCoherent()` does not check for nil pointers either.

PR #588, opened 2020-12-07
Pinned texture view

johnplate:
This implements the concept of resource view tabs where the content can be preserved ("pinned") until the tab is closed. This concept is first applied to texture view tabs. The pin icon is from material.io.

It also adds a history of selected tabs for each tab folder, which are used to activate the most recently selected tab after the selected tab is closed or dragged to another folder.

Third PR for b/170298289, and fix for b/174851718.


gapic/src/main/com/google/gapid/widgets/TabComposite.java
```diff
@@ -913,12 +952,49 @@ public void addTabToLargestFolder(TabInfo tab) {
...
+      }
+      return null;
+    }
+
+    @Override
+    public void disposePinnedTabs() {
+      int index = 0;
+      while (index < tabs.size()) {
+        Tab tab = tabs.get(index);
+        if (tab.info.id == tab.content) {
```
stellama0208, line 978: This predicate seems a bit unintuitive to me to check whether this tab is a pinned tab. I'm wondering is there a reason that `tab.content.isPinned()` doesn't fit in here?

stellama0208 (approved):
The PR LGTM as the start of bringing pinning feature into AGI, thanks for making it! Also verified the feature by manual testing. Only with some small suggestions.

gapic/src/main/com/google/gapid/widgets/Theme.java
```diff
@@ -96,6 +96,7 @@
   @Icon(file = "open.png") public Image open();
   @Icon(file = "overdraw.png") public Image overdraw();
   @Icon(file = "pan_mode.png") public Image panMode();
+  @Icon(file = "pin.png") public Image pin();
```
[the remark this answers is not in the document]
johnplate: I reused the old one before, but I replaced it with a version from material.io after Pascal asked to use the close icon from there. Another reason was that all tab bar icons have a size of 24 pixel, but the old pin icons are 16 pixel. I am open to use the old one, but I assume Pascal would object.

johnplate:
> The PR LGTM as the start of bringing pinning feature into AGI, thanks for making it! Also verified the feature by manual testing. Only with some small suggestions.

Thank you for your review! I have answered to your suggestions.

gapic/src/main/com/google/gapid/widgets/ImagePanel.java
```diff
@@ -479,6 +479,10 @@ protected void save() {
     }
   }
 
+  public boolean hasImage() {
+    return image != MultiLayerAndLevelImage.EMPTY;
```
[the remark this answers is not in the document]
johnplate: image is initialized and used consistently in a way that assumes that it will never be null, so in my opinion there is no null check necessary.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
johnplate: Thanks, good point! I have added the function Tab.isPinned() to make it more readable.
The class TabComposite does not control if TabContent.isPinned() reliably always returns true when it is pinned, so there is a danger of a crash if it is not correctly implemented. However TabComposite.Folder does control the value of tab.info.id when it pins a tab, so it is more safe to use that value here.

gapic/src/main/com/google/gapid/widgets/Theme.java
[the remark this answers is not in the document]
pmuetschard: No I agree, we should use the existing pin icon, it comes from the material icons.

gapic/src/main/com/google/gapid/widgets/TabArea.java
```diff
@@ -47,7 +48,8 @@ public void onTabCreated(TabInfo tab) {
 
       @Override
       public void onTabClosed(TabInfo tab) {
-        analytics.postInteraction(tab.view, ClientAction.Disable);
+        analytics.postInteraction(tab.view,
```
pmuetschard, line 51: This is leaking an implementation detail out of the `TabComposite`. Just pass a boolean `pinned` to the `onTabClosed` event.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
```diff
@@ -20,6 +20,7 @@
 import com.google.gapid.models.Analytics;
 import com.google.gapid.models.Analytics.View;
 import com.google.gapid.util.Events;
+import com.google.gapid.views.TabContent;
```
pmuetschard, line 23: Yikes. This is a dependency between packages I don't like. I think the solution is to:

1) keep the original `views.Tab` interface with `reinitialize` method
2) Create the new content interface as an inner class here.
3) make `views.Tab` extend the new content interface.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
```diff
@@ -362,13 +381,13 @@ private Hover find(int mx, int my) {
    * Information about a single tab in a folder.
    */
   public static class TabInfo {
-    public final Object id;
+    public final Object id;  // holds GraphicsTraceView.MainTab.Type or TabContent for pinned tabs
```
pmuetschard, line 384:
```suggestion
    public final Object id;  // Externally provided ID or TabContent for pinned tabs.
```

gapic/src/main/com/google/gapid/widgets/Theme.java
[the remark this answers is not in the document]
johnplate: Ok, I removed the new icon and used the existing again. I was not able to find our existing pin icon on material.io, so I guess they updated it.

gapic/src/main/com/google/gapid/widgets/TabArea.java
johnplate: Yes, thanks. I have added the function TabInfo.isPinned() to hide the implementation detail.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
johnplate: Sure, done.

gapic/src/main/com/google/gapid/widgets/TabComposite.java
johnplate: Done

pmuetschard (approved):

PR #592, opened 2020-12-09
Create a function that uses SPIRV-Reflect to obtain source code

hliatis:
Note: Source code might not exist in OpSource

Bug: b/175214028

### DIFF
--- a/gapis/shadertools/shadertools.go
+++ b/gapis/shadertools/shadertools.go
@@ -21,6 +21,7 @@ package shadertools
 import "C"
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


gapis/shadertools/shadertools.go
```diff
@@ -189,6 +189,35 @@ func descriptorBindingLess(a DescriptorBinding, b DescriptorBinding) bool {
...
+	shaderPtr := unsafe.Pointer(nil)
+	if len(shader) > 0 {
+		shaderPtr = unsafe.Pointer(&shader[0])
+	}
+	if err := spvReflectErr(C.spvReflectCreateShaderModule(
+		C.size_t(len(shader)*4),
+		shaderPtr,
+		&module)); err != nil {
+		return "", err
+	}
```
pmuetschard, line 211:
```suggestion
	err := spvReflectErr(C.spvReflectCreateShaderModule(C.size_t(len(shader)*4), shaderPtr, &module))
	if err != nil {
		return "", err
	}
```

gapis/shadertools/shadertools.go
```diff
@@ -189,6 +189,35 @@ func descriptorBindingLess(a DescriptorBinding, b DescriptorBinding) bool {
...
+		if res == C.SPV_REFLECT_RESULT_SUCCESS {
+			return nil
+		}
+		return fmt.Errorf("SPIRV-Reflect failed with error code %v\n", res)
+	}
+
+	module := C.SpvReflectShaderModule{}
+
+	shaderPtr := unsafe.Pointer(nil)
+	if len(shader) > 0 {
```
pmuetschard, line 203: Instead of relying on SPIRV-Reflect to deal with empty shaders, we should just return the empty string/an error here.

gapis/shadertools/shadertools.go
```diff
@@ -189,6 +189,35 @@ func descriptorBindingLess(a DescriptorBinding, b DescriptorBinding) bool {
...
+	if err := spvReflectErr(C.spvReflectCreateShaderModule(
+		C.size_t(len(shader)*4),
+		shaderPtr,
+		&module)); err != nil {
+		return "", err
+	}
+	defer C.spvReflectDestroyShaderModule(&module)
+
+	if module.source_source == nil {
+		return "", fmt.Errorf("No Source File Detected\n")
```
pmuetschard, line 215: remove the `\n`. Also use `errors.New("...")`, since you aren't actually formatting anything.

gapis/shadertools/shadertools.go
```diff
@@ -189,6 +189,35 @@ func descriptorBindingLess(a DescriptorBinding, b DescriptorBinding) bool {
 	return a.SpirvId < b.SpirvId
 }
 
+func ParseSourceCode(shader []uint32) (string, error) {
```
pmuetschard, line 192: This name is misleading/inaccurate. I think something like `ExtractDebugSource` would be better.

hliatis: I'm also going to put the SPIRV-Cross logic in this function. Do you recommend I put it somewhere else?

hliatis:
Commenting on the PR to see if the CLA will get unstuck

gapis/shadertools/shadertools.go
hliatis: Note: I use SPIRV-Reflect to set some of the compiler options, like version number, for SPIRV-Cross

pmuetschard (approved):

PR #593, opened 2020-12-11
Clarify capture duration

hevrard:
Clarify capture duration, especially for 1-frame capture.

This change clarifies these durations in both the trace capture
settings dialog, and the capture progress dialog.

Bug: b/173103059
Test: manual


hevrard:
For the frame duration option, I though about keeping the spinner and disable it (to prevent changes), this leads to:

stopAfter

I find hard to see the number when the spinner is disabled and greyed out.

Hence I opted for a special-case label that I think is cleaner:

duration

johnplate (approved):

hevrard:
The Linux CI failure is a swarming flake due to adb device connection lost.

PR #594, opened 2020-12-11
Add capture start time unit, remove MEC support flag

hevrard:
Since AGI dropped GLES support, now all supported APIs support MEC, so
remove the mid_execution_capture_support flag which is now obsolete.

In the UI, this lets us refactor mecWarningLabel into a startUnit
label that is used to show the time unit (seconds) for "Time" capture
start type.

Bug: b/173098919
Test: manual


johnplate (approved):

hevrard:
The Linux CI failure is a flake due to some 3rd party repo not being available.

PR #595, opened 2020-12-12
Set gpu debug property to enable profiling feature before tracing

rschiu:
Bug: 173625393

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -515,7 +515,8 @@ func (b *binding) PrepareGpuProfiling(ctx context.Context, installedPackage *and
 		log.W(ctx, "Failed to query developer driver: %v, assuming no developer driver found.", err)
 	}
 
-	// Setup debug property.  The GPU driver may enable profiling only if the debug property is set.
+	// Setup debug property.  The GPU driver must enable profiling and have proper
+	// instrumentation if the debug property is set to 1.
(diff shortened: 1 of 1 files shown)


core/os/android/adb/commands.go
```diff
@@ -513,11 +515,21 @@ func (b *binding) PrepareGpuProfiling(ctx context.Context, installedPackage *and
 		log.W(ctx, "Failed to query developer driver: %v, assuming no developer driver found.", err)
 	}
 
+	// Setup debug property.  The GPU driver may enable profiling only if the debug property is set.
```
hevrard, line 518: Nit: "The GPU driver ..." -> "Some Qualcomm drivers ..."

hevrard (approved):
LGTM with a nit

core/os/android/adb/commands.go
pmuetschard: No. Others/future drivers may now rely on this as well.

lpy: Just to clarify, this is now the our expectation of all GPU drivers behaviour. All GPU vendors should rely on this property to check whether commands must be instrumented.

lpy, line 518: I think it would be better if the comment is: `The GPU driver must enable profiling and have property instrumentation if the debug property is set to 1.`

lpy (approved):

PR #598, opened 2020-12-14
Properly Parse/Handle the ASAN ABIs.

pmuetschard:
Bug: http://b/169846576


gapidapk/deviceinfo.go
```diff
@@ -105,6 +105,7 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
 	apk.Stop(ctx)
 
 	var cleanup app.Cleanup
+	instance := d.Instance()
```
hevrard, line 108: While we're at it, should we use `instance` on L138 (in `flock.Lock`)?

hevrard (approved):

PR #602, opened 2020-12-16
Split shader tab

johnplate:
This request cleaned up the shader tab from features that used to be required for OpenGL support, and splits it into a list and a view tab, similar to the texture tab.

Fourth (and last) PR for b/170298289

### DIFF
--- a/gapic/src/main/com/google/gapid/views/ShaderView.java
+++ b/gapic/src/main/com/google/gapid/views/ShaderView.java
@@ -76,6 +76,7 @@ public class ShaderView extends Composite
   private final Group spirvGroup;
   private final Group sourceGroup;
   private final Label crossCompileLabel;
+  private final GridData crossCompileGridData;
   private final SourceViewer spirvViewer;
   private final SourceViewer sourceViewer;
   private final Button pushButton;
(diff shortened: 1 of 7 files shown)


gapic/src/main/com/google/gapid/views/ShaderView.java
```diff
@@ -218,519 +125,106 @@ public Control getControl() {
...
-    if (models.commands.isLoaded() && models.resources.isLoaded()) {
-      if (models.commands.getSelectedCommands() == null) {
-        loading.showMessage(Info, Messages.SELECT_COMMAND);
-      } else {
-        loading.stopLoading();
-      }
+  @Override
+  public void onResourcesLoaded() {
+    if (!pinned) {
+      loading.showMessage(Info, Messages.SELECT_TEXTURE);
```
pmuetschard, line 169: This is the wrong message.

gapic/src/main/com/google/gapid/views/ShaderView.java
```diff
@@ -218,519 +125,106 @@ public Control getControl() {
...
+      onCaptureLoadingStart(false);
+    } else {
+      loadShader(models.resources.getSelectedShader());
+    }
   }
 
   @Override
-  public void onCaptureLoadingStart(boolean maintainState) {
-    loading.showMessage(Info, Messages.LOADING_CAPTURE);
+  public boolean isPinnable() {
```
pmuetschard, line 136: Missing `supportsPinning()` returning true.

pmuetschard (commented):
Please also rebase on top of master, since there's merge conflicts (I moved `API.ResourceType` to `Path.ResourceType` and renamed the enum constants).

gapic/src/main/com/google/gapid/views/ShaderView.java
johnplate: Oops, thanks, fixed

gapic/src/main/com/google/gapid/views/ShaderView.java
johnplate: Thanks, good catch. Fixed

pmuetschard (approved):

PR #603, opened 2020-12-17
Framegraph: fix reporting of mapped memory

hevrard:
This change sets the memoryMapped flag of framegraph buffers/images
only when the mapped memory is actually overlapping a range of memory
that backs the buffer/image.

Also, when updating the buffer/image lookup tables, make sure to also
scan the sparse memory of each resource.

Also, stop assuming that there is always a non-sparse memory bound,
this removes some possible NPE.

Bug: b/175394793, b/174505556
Test: manual

### DIFF
--- a/gapis/api/vulkan/framegraph.go
+++ b/gapis/api/vulkan/framegraph.go
@@ -43,14 +43,39 @@ func backedByCoherentMemory(state *State, mem DeviceMemoryObjectʳ) bool {
 
 func newFramegraphBuffer(state *State, buf *BufferObjectʳ) *api.FramegraphBuffer {
 	usage := uint32(buf.Info().Usage())
-	coherentMemory := backedByCoherentMemory(state, buf.Memory())
-	memoryMapped := !(buf.Memory().MappedLocation().IsNullptr())
+	coherentMemory := false
+	memoryMapped := false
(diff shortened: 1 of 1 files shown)


hevrard:
A gentle ping on this one

johnplate (approved):
LGTM, but should be reviewed also by someone who is more familiar with this code part.

gapis/api/vulkan/framegraph.go
```diff
@@ -82,25 +107,61 @@ func newFramegraphImage(state *State, img *ImageObjectʳ) *api.FramegraphImage {
...
+							if !(mem.MappedLocation().IsNullptr()) {
+								memRange := memory.Range{
+									Base: uint64(block.MemoryOffset()),
+									Size: uint64(block.Size()),
+								}
+								mappedRange := memory.Range{
+									Base: uint64(mem.MappedOffset()),
+									Size: uint64(mem.MappedSize()),
+								}
+								memoryMapped = memoryMapped || memRange.Overlaps(mappedRange)
```
AWoloszyn, line 162: You should be able to drop out of this nested loop if coherent&&mapped.

hevrard: Added early exits in loops.

PR #607, opened 2020-12-18
Add more logs when aborting AGI.

lpy:
Layer setup is essentially setting up the settings global variables provided by
the settings provider. However, different OEMs may enforce extra permissions on
that due to security concerns, etc. Hence we must log the errors before AGI
aborts.

Bug: b/175579064


lpy:
ptal

pmuetschard (approved):

PR #608, opened 2020-12-22
Show decompiled shader source

hliatis:
Bug: b/175214232. b/175214587

### DIFF
--- a/gapis/api/vulkan/resources.go
+++ b/gapis/api/vulkan/resources.go
@@ -835,8 +835,12 @@ func (s ShaderModuleObjectʳ) ResourceData(ctx context.Context, t *api.GlobalSta
 	if err != nil {
 		return nil, fmt.Errorf("Could not get resource data %v", err)
 	}
-	source := shadertools.DisassembleSpirvBinary(words)
-	return api.NewResourceData(&api.Shader{Type: api.ShaderType_Spirv, Source: source}), nil
+	source, sourceLanguage, isCross, err := shadertools.ExtractDebugSource(words)
+	if err != nil {
+		log.E(ctx, "Error decompiling shader: %v", err)
+	}
+	spirv := shadertools.DisassembleSpirvBinary(words)
+	return api.NewResourceData(&api.Shader{Type: api.ShaderType_Spirv, Source: source, SpirvSource: spirv, SourceLanguage: sourceLanguage, CrossCompiled: isCross}), nil
 }
 
 func (shader ShaderModuleObjectʳ) SetResourceData(
@@ -1082,8 +1086,12 @@ func commonShaderDataGroups(ctx context.Context,
 			module := stage.Module()
 
 			words, _ := module.Words().Read(ctx, nil, s, nil)
-			source := shadertools.DisassembleSpirvBinary(words)
-			shader := &api.Shader{Type: api.ShaderType_Spirv, Source: source}
+			source, sourceLanguage, isCross, err := shadertools.ExtractDebugSource(words)
+			if err != nil {
+				log.E(ctx, "Error decompiling shader: %v", err)
+			}
+			spirv := shadertools.DisassembleSpirvBinary(words)
+			shader := &api.Shader{Type: api.ShaderType_Spirv, Source: source, SpirvSource: spirv, SourceLanguage: sourceLanguage, CrossCompiled: isCross}
 
 			dsetRows := []*api.Row{}
 			for _, usedSet := range usedSets {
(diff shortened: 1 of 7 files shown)


tools/build/third_party/spirv-cross.BUILD
```diff
@@ -21,6 +21,11 @@ cc_library(
     ]),
     hdrs = [
         "spirv_glsl.hpp",
+        "spirv_hlsl.hpp",
+    ],
+    defines = [
+        "SPIRV_CROSS_C_API_GLSL",
+        "SPIRV_CROSS_C_API_HLSL",
```
pmuetschard, line 28: This adds these defines to everyone that depends on this. Since they are not used in the header files, use `local_defines` instead.

gapis/shadertools/shadertools.go
```diff
@@ -208,15 +209,51 @@ func ExtractDebugSource(shader []uint32) (string, error) {
...
-		return "", err
+		return "", "", false, err
 	}
 	defer C.spvReflectDestroyShaderModule(&module)
 
+	sourceLanguage := C.GoString(C.spvReflectSourceLanguage(module.source_language))
+
 	if module.source_source == nil {
-		return "", errors.New("No Source File Detected")
+		var context C.spvc_context = nil
```
pmuetschard, line 219: make it 
```
if module.source_source != nil {
  return C.GoString(...), nil
}

// Get using SPIR-V Cross.
...
```
It's easier to follow this way.

gapis/shadertools/shadertools.go
```diff
@@ -190,7 +191,7 @@ func descriptorBindingLess(a DescriptorBinding, b DescriptorBinding) bool {
 	return a.SpirvId < b.SpirvId
 }
 
-func ExtractDebugSource(shader []uint32) (string, error) {
+func ExtractDebugSource(shader []uint32) (string, string, bool, error) {
```
pmuetschard, line 194: Please add a doc string comment to this function and say what all the returned values are.

gapis/shadertools/shadertools.go
```diff
@@ -208,15 +209,51 @@ func ExtractDebugSource(shader []uint32) (string, error) {
...
+		var ir C.spvc_parsed_ir = nil
+		var compiler C.spvc_compiler = nil
+		var options C.spvc_compiler_options = nil
+		var result *C.char = nil
+
+		C.spvc_context_create(&context)
+		C.spvc_context_parse_spirv(context, (*C.uint)(shaderPtr), (C.ulong)(len(shader)), &ir)
+		switch sourceLanguage {
+		case "GLSL":
+			if C.spvc_context_create_compiler(context, C.SPVC_BACKEND_GLSL, ir, C.SPVC_CAPTURE_MODE_TAKE_OWNERSHIP, &compiler) != C.SPVC_ERROR_INVALID_ARGUMENT {
```
pmuetschard, line 229: Here and below, also invert the condition and use an early exit to make it easier to read.

gapis/api/vulkan/resources.go
```diff
@@ -1080,8 +1084,12 @@ func commonShaderDataGroups(ctx context.Context,
 			module := stage.Module()
 
 			words, _ := module.Words().Read(ctx, nil, s, nil)
-			source := shadertools.DisassembleSpirvBinary(words)
-			shader := &api.Shader{Type: api.ShaderType_Spirv, Source: source}
+			source, sourceLanguage, isCross, err := shadertools.ExtractDebugSource(words)
+			if err != nil {
+				source = err.Error()
```
pmuetschard, line 1089: Don't return the error as the source. Just return an empty string and handle it in the UI to not show the source tab.
You can log the error, so it's not lost, and in the future, we could return it properly to the UI, but we don't really have good support for returning values along with errors right now.

gapis/api/vulkan/resources.go
```diff
@@ -835,8 +835,12 @@ func (s ShaderModuleObjectʳ) ResourceData(ctx context.Context, t *api.GlobalSta
 	if err != nil {
 		return nil, fmt.Errorf("Could not get resource data %v", err)
 	}
-	source := shadertools.DisassembleSpirvBinary(words)
-	return api.NewResourceData(&api.Shader{Type: api.ShaderType_Spirv, Source: source}), nil
+	source, sourceLanguage, isCross, err := shadertools.ExtractDebugSource(words)
+	if err != nil {
+		return nil, fmt.Errorf("Could not get decompile shader: %v", err)
```
pmuetschard, line 840: This should not be a fatal error. Do the same here as you did for the Pipelines below: log the error and return a SPIRV-only shader.

pmuetschard (commented):
Please rebase on top of master and fix the compile errors (I've removed some, at the time, unused imports from the PipelineView, that you are now using).

pmuetschard (approved):

PR #610, opened 2020-12-28
Parallelize generate and replay execution

rschiu:
Bug: 155057258

### DIFF
--- a/gapis/replay/executor.go
+++ b/gapis/replay/executor.go
@@ -48,6 +48,8 @@ type executor struct {
 // called.
 func Execute(
 	ctx context.Context,
+	deviceID id.ID,
+	replayStatus *status.Replay,
 	dependent string,
 	payload gapir.Payload,
 	handlePost builder.PostDataHandler,
@@ -63,7 +65,7 @@ func Execute(
 
 	// The memoryLayout is specific to the ABI of the requested capture,
 	// while the OS is not. Thus a device.Configuration is not applicable here.
-	return executor{
+	e := executor{
 		payload:            payload,
 		dependent:          dependent,
 		handlePost:         handlePost,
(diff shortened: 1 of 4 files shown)


rschiu:
Screen Shot 2020-12-28 at 5 02 07 AM

Example of patch in action.  Notice batch is run when "Execute Replay" is also running on another thread

gapis/replay/manager.go
```diff
@@ -153,13 +162,32 @@ func (m *manager) createScheduler(ctx context.Context, device bind.Device) {
...
+			for {
+				args := <-ch
+				// TODO(pmuetschard): Make the "state reconstruction" an actual event status and have gapir let us
+				// know how far along it is in executing it. Also, there is technically still some server work
+				// happening after this (e.g. storing the payload in the db), but that is fast and passing things
+				// down isn't worth it.
+				// This transitions the status from building to executing.
+				args.replayStatus.Progress(ctx, 0, 1, 0)
+				ctx = status.Start(ctx, "Execute Replay")
+				args.exec.execute(ctx, m, args.key)
```
AWoloszyn, line 178: The error code here is swallowed and not returned to anyone.

gapis/replay/executor.go
```diff
@@ -72,7 +74,16 @@ func Execute(
...
+	}
+
+	args := executeArgs {
+		exec:         e,
+		key:          key,
+		replayStatus: replayStatus,
+	}
+
+	m.(*manager).executors[deviceID] <- args
+	return nil
```
AWoloszyn, line 86: How do we propagate the errors that used to come from executor{}.execute?

gapis/replay/manager.go
```diff
@@ -153,13 +162,32 @@ func (m *manager) createScheduler(ctx context.Context, device bind.Device) {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
 	m.schedulers[deviceID] = scheduler.New(ctx, deviceID, m.batch)
+	ch := make(chan executeArgs, 16)
```
pmuetschard, line 165: This channel should be size zero. We want batching to still happen. If there is a replay being executed on the device, and we've already finished generating the next one, it is not advantageous to generate a 3rd one. Setting the size to zero, will still allow the second payload to be generated, but it will then block and wait for the first replay to finish before returning control to the generator, thus 3rd and subsequent requests can still be batched.

pmuetschard (changes requested):
Please also address Andrew's concerns. Remember, we talked about the error handling: it can be done correctly, since it's already asynchronous. It just requires the error handling to be pushed down in the new go-routine.

PR #613, opened 2020-12-29
Faster Resource Handle to ID Map Creation

pmuetschard:
To get this map, we do not need to resolve the resource data at the given command, but can use the global resource information and a little bit of logic to cover the handle re-use case to generate the mapping, when needed.

This gets rid of an extra mutate loop, when fetching the pipelines at a given command, i.e. speeds up command selection.

### DIFF
--- a/gapis/resolve/resource_meta.go
+++ b/gapis/resolve/resource_meta.go
@@ -57,3 +57,30 @@ func (r *ResourceMetaResolvable) Resolve(ctx context.Context) (interface{}, erro
 	}
 	return result, nil
 }
+
+// ResourceIDMap returns the ResourceMap at the given command.
+func ResourceIDMap(ctx context.Context, after *path.Command, r *path.ResolveConfig) (api.ResourceMap, error) {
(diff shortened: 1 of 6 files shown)


gapis/resolve/resource_meta.go
```diff
@@ -51,9 +41,32 @@ func (r *ResourceMetaResolvable) Resolve(ctx context.Context) (interface{}, erro
...
+	resources, err := Resources(ctx, after.Capture, r)
+	if err != nil {
+		return nil, err
+	}
+
+	m := api.ResourceMap{}
+	created := map[string]*path.Command{}
+	for _, ty := range resources.Types {
+		for _, res := range ty.Resources {
+			if !after.IsAfter(res.Created) {
```
hevrard, line 58: Shouldn't this check that the `after` command is after the resource creation, to map only the resources which are created by the time of the `after` command?

```suggestion
			if after.IsAfter(res.Created) {
```

pmuetschard: Fixed.

hevrard (approved):

PR #624, opened 2021-01-06
Faster Pipeline Resolve

pmuetschard:
Instead of resolving the bound pipelines by getting all Pipeline resource datas and then filtering to the bound ones, use the state to find the bound pipelines. This avoids having to do another mutate run to resolve all the pipeline resource datas, most of which are then summarily ignored.

For a resource heavy trace, this brings the time it takes to get the bound pipelines on command selection in the UI from ~5 seconds to well below one second.

### DIFF
--- a/gapis/api/pipeline.go
+++ b/gapis/api/pipeline.go
@@ -1,0 +1,40 @@
+// Copyright (C) 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import (
+	"context"
+	"errors"
+
+	"github.com/google/gapid/gapis/service/path"
+)
+
+var (
+	// ErrPipelineNotAvailable is an error returned by PipelineProvider if the
+	// pipelines are requested on an object that does not have pipelines.
+	ErrPipelineNotAvailable = errors.New("Pipelines not available at this command")
+)
+
+// BoundPipeline represents a pipeline resource and its data currently bound.
+type BoundPipeline struct {
+	Pipeline Resource
+	Data     *ResourceData
+}
+
+// PipelineProvider is the interface implemented by types that provide pipelines.
+type PipelineProvider interface {
+	// BoundPipeline returns the pipeline bound at object o.
+	BoundPipeline(ctx context.Context, o interface{}, p *path.Pipelines, r *path.ResolveConfig) (BoundPipeline, error)
+}
--- a/gapis/api/vulkan/draw_call_pipeline.go
+++ b/gapis/api/vulkan/draw_call_pipeline.go
@@ -1,0 +1,77 @@
+// Copyright (C) 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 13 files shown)


gapis/api/pipeline.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (C) 2020 Google Inc.
```
hevrard, line 1: 2021

gapis/api/vulkan/draw_call_pipeline.go
```diff
@@ -0,0 +1,77 @@
+// Copyright (C) 2020 Google Inc.
```
hevrard, line 1: 2021

gapis/api/vulkan/draw_call_pipeline.go
```diff
@@ -0,0 +1,77 @@
...
+
+	lastQueue := c.LastBoundQueue()
+	if lastQueue.IsNil() {
+		return bound, fmt.Errorf("No previous queue submission")
+	}
+
+	if cmd.CmdFlags().IsExecutedDraw() {
+		lastDrawInfo, ok := c.LastDrawInfos().Lookup(lastQueue.VulkanHandle())
+		if !ok {
+			return bound, fmt.Errorf("There have been no previous draws")
```
hevrard, line 60: typo: no previous draws -> no previous draw

gapis/api/vulkan/draw_call_pipeline.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		bound.Pipeline = lastDrawInfo.GraphicsPipeline()
+	} else {
+		lastComputeInfo, ok := c.LastComputeInfos().Lookup(lastQueue.VulkanHandle())
+		if !ok {
+			return bound, fmt.Errorf("There have been no previous draws")
```
hevrard, line 66: maybe rephrase: `no previous compute dispatch` ?

gapis/resolve/pipeline.go
```diff
@@ -0,0 +1,91 @@
+// Copyright (C) 2020 Google Inc.
```
hevrard, line 1: 2021

gapis/resolve/pipeline.go
```diff
@@ -0,0 +1,91 @@
...
+	obj, err := ResolveInternal(ctx, p.Parent(), r)
+	if err != nil {
+		return nil, err
+	}
+
+	if pipelines, err := pipelinesFor(ctx, obj, p, r); err != api.ErrPipelineNotAvailable {
+		return pipelines, err
+	}
+	return nil, &service.ErrDataUnavailable{Reason: messages.ErrNotADrawCall()}
+}
```
hevrard, line 39: Here and in other places, the api.ErrPipelineNotAvailable is replaced by messages.ErrNotADrawCall, I'm not sure why we don't simply forward the api.ErrPipelineNotAvailable error?

hevrard (approved):
LGTM with nits

gapis/resolve/pipeline.go
pmuetschard: `ErrPipelineNotAvailable` is not a "service error" and would be wrapped inside an `ErrInternalError`. However, we want to return the `ErrDataUnavailable` error here, because it has a specific semantic meaning to the client. `ErrPipelineNotAvailable` is not so much an "error" but a "null" indicator, kind of like `io.EOF`, i.e. not unexpected.

gapis/api/vulkan/draw_call_pipeline.go
pmuetschard: That would be grammatically incorrect ("There have been" is plural).

gapis/api/vulkan/draw_call_pipeline.go
pmuetschard: Done.

gapis/resolve/pipeline.go
pmuetschard: Done.

gapis/api/pipeline.go
pmuetschard: Done.

gapis/api/vulkan/draw_call_pipeline.go
pmuetschard: Done.

PR #626, opened 2021-01-07
Don't call protoutil.OneOf on fields that are not proto oneof's.

pmuetschard:
Doing so 1) skips over a path element to validate, and 2) will cause a panic in the future, where proto messages will contain hidden fields.

### DIFF
--- a/gapis/service/path/validate.go
+++ b/gapis/service/path/validate.go
@@ -171,22 +171,22 @@ func (n *DeviceTraceConfiguration) Validate() error {
 
 // Validate checks the path is valid.
 func (n *Events) Validate() error {
-	return checkNotNilAndValidate(n, protoutil.OneOf(n.Capture), "capture")
+	return checkNotNilAndValidate(n, n.Capture, "capture")
 }
 
 // Validate checks the path is valid.
 func (n *FramebufferObservation) Validate() error {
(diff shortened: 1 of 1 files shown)


gapis/service/path/validate.go
```diff
@@ -171,22 +171,22 @@ func (n *DeviceTraceConfiguration) Validate() error {
...
 // Validate checks the path is valid.
 func (n *FramebufferObservation) Validate() error {
-	return checkNotNilAndValidate(n, protoutil.OneOf(n.Command), "command")
+	return checkNotNilAndValidate(n, n.Command, "command")
 }
 
 // Validate checks the path is valid.
 func (n *FramebufferAttachments) Validate() error {
-	return checkNotNilAndValidate(n, protoutil.OneOf(n.After), "command")
+	return checkNotNilAndValidate(n, n.After, "command")
```
stellama0208, line 184: Not directly related to the purpose of this PR, but should we also fix a nit btw, by changing the third parameter to "after"? same for the next changed line.

stellama0208 (approved):
LGTM!

gapis/service/path/validate.go
pmuetschard: Done.

PR #628, opened 2021-01-08
Update the ASTC encoder

yalcinmelihyasin:
Update the ASTC encoder library and refactor the code too use new API


core/image/astc/astc.cc
```diff
@@ -13,77 +13,109 @@
 // limitations under the License.
 
 #include <stdio.h>
+#include <string.h>
 
 #include "astc.h"
+#include "third_party/astc-encoder/Source/astcenccli_internal.h"
 
-#include "third_party/astc-encoder/Source/astc_codec_internals.h"
+int init_astc_for_decode(astcenc_profile profile,
```
pmuetschard, line 21: Instead of having a go callback that just ends up printing to stdout, return the actual error code and do the string handling in the go side to turn it into proper go errors. I think you can reference the enum values via `C.<enum name>` and should be able to call the `astcenc_get_error_string` function directly via `C.astcenc_get_error_string`.

pmuetschard (commented):
Since you are basically rewriting the astc C++ code, please rename the file to `astc.cpp` and make sure it passes the linter.

core/image/astc/astc.cc
yalcinmelihyasin: Makes sense

yalcinmelihyasin: I have just added a wrapper function as we have to expose the astcenc_get_error_string function to be able to call from Go side.

I think that rather than exposing the direct astcenc function and entangle the astc library with go code, it is better to decouple them with a wrapper in between.

core/image/astc/astc.h
```diff
@@ -12,17 +12,25 @@
...
 
-void decompress_astc(uint8_t* in, uint8_t* out, uint32_t width, uint32_t height,
-                     uint32_t block_width, uint32_t block_height);
+const char* get_error_string(astc_error error_code);
 
 #ifdef __cplusplus
 }  // extern "C"
 #endif
+
+#endif
```
pmuetschard, line 36: Add the newline, please.

core/image/astc/astc.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+  unsigned int thread_count = get_cpu_count();
+  astcenc_context* codec_context;
+  result = astcenc_context_alloc(config, thread_count, &codec_context);
+  if (result != ASTCENC_SUCCESS) {
+    return result;
+  }
+
+  unsigned int bitness = 8;
+  astcenc_image* output_image = alloc_image(
```
pmuetschard, line 91: Can this alloc fail?

core/image/astc/astc.cpp
```diff
@@ -0,0 +1,120 @@
...
+      bitness, input_image.dim_x, input_image.dim_y, input_image.dim_z, 0);
+
+  astcenc_swizzle swz_decode{ASTCENC_SWZ_R, ASTCENC_SWZ_G, ASTCENC_SWZ_B,
+                             ASTCENC_SWZ_A};
+  result =
+      astcenc_decompress_image(codec_context, input_image.data,
+                               input_image.data_len, *output_image, swz_decode);
+  if (result != ASTCENC_SUCCESS) {
+    return result;
```
pmuetschard, line 100: Should the context and image be freed here?

core/image/astc/astc.cpp
yalcinmelihyasin: The implementation of the particular function in `astc_encoder` does not check for an allocation failure. In any case of allocation failure we would crash in alloc_image function. We can try to change that but I am quite sure lots of similar scenarios could happen at any place in the encoding library. Therefore I suggest not diverge from their source code until we have a problem.

core/image/astc/astc.cpp
yalcinmelihyasin: Good catch!

core/image/astc/astc.cpp
pmuetschard: What I meant was: can `alloc_image` ever return `null`? It sounds like the answer is no, so we don't have to check for that.

pmuetschard (approved):

core/image/astc/astc.cpp
yalcinmelihyasin: Yes it does not.

PR #635, opened 2021-01-15
[do not merge] debug continuous testing issue

hevrard:


hevrard:
Proper fix is in https://github.com/google/agi/pull/644

PR #637, opened 2021-01-16
Implement column filtering and column customization remembering feature in perftab.

stellama0208:
In performance tab, AGI users could now filter interested GPU counter columns
easily. At the same time, they could also save some most used combinations into a
"preset", which will be remembered in the AGI's setting file. In this way,
the next time they open another AGI session in the future, these "presets"
will directly show up in the performance tab for quick filtering.

Bug: b/159739794.
Test: manual test in performance tab.


stellama0208:
The above force push is only to revert a wrong commit pushed here accidentally, never mind about it, PR ready for review.

gapis/trace/android/adreno/profiling_data.go
```diff
@@ -294,12 +299,18 @@ func processCounters(ctx context.Context, processor *perfetto.Processor, desc *d
 			timestamps[i] = uint64(t)
 		}
 		values := countersColumns[1].GetDoubleValues()
+
+		var spec *device.GpuCounterDescriptor_GpuCounterSpec
+		if s, ok := nameToSpec[names[i]]; ok {
+			spec = s
+		}
```
pmuetschard, line 306:
```suggestion
		spec, _ := nameToSpec[names[i]]
```

gapis/trace/android/mali/profiling_data.go
```diff
@@ -288,12 +293,18 @@ func processCounters(ctx context.Context, processor *perfetto.Processor, desc *d
 			timestamps[i] = uint64(t)
 		}
 		values := countersColumns[1].GetDoubleValues()
+
+		var spec *device.GpuCounterDescriptor_GpuCounterSpec
+		if s, ok := nameToSpec[names[i]]; ok {
+			spec = s
+		}
```
pmuetschard, line 300:
```suggestion
		spec, _ := nameToSpec[names[i]]
```

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -426,7 +426,7 @@ public default Control asControl() {
     }
   }
 
-  private static class BasicInputArea extends Composite implements InputArea {
+  public static class BasicInputArea extends Composite implements InputArea {
```
pmuetschard, line 429: Leave this one private, but make the GpuCountersDialog not be an inner class of this, but just of `TraceConfigDialog`.

gapic/src/main/com/google/gapid/widgets/Widgets.java
```diff
@@ -408,6 +408,14 @@ public static Button createButton(Composite parent, String text, Listener listen
     return result;
   }
 
+  public static Button createButton(Composite parent, String text, int style, Color color, Listener listener) {
```
pmuetschard, line 411: Make the style the second argument after the parent to stick with the established pattern.

gapic/src/main/com/google/gapid/widgets/Widgets.java
```diff
@@ -408,6 +408,14 @@ public static Button createButton(Composite parent, String text, Listener listen
...
+    Button result = new Button(parent, style);
```
pmuetschard, line 412: I think the style should be or'ed **here** with `SWT.PUSH`. I see usage only with `SWT.FLAT`, which is rendering hint, and so the buttons are created without the type style and it defaults to push buttons. To be explicit, what you want is `SWT.PUSH | SWT.FLAT`, but with these factory methods, we've removed the `SWT.PUSH` vs. `SWT.CHECK` vs. `SWT.RADIO` non-sense from the callers, so we should stay with that and only use it to specify hint styles.

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
stellama0208: Thanks for suggestion, this organization would suit the current situation better!

gapic/src/main/com/google/gapid/widgets/Widgets.java
stellama0208: Done, thanks for explaining, now added it to make it explicit and conform with other factory functions.

stellama0208:
The above comments are addressed with a new commit, feel free to take another look!

pmuetschard (approved):

PR #638, opened 2021-01-18
Add compression capability for ASTC.

yalcinmelihyasin:


core/image/BUILD.bazel
```diff
@@ -19,7 +19,7 @@ load("@rules_proto//proto:defs.bzl", "proto_library")
 go_library(
     name = "go_default_library",
     srcs = [
-        "astc.go",
+        "astc_format.go",
```
pmuetschard, line 22: Why the rename? They're all formats.

core/image/astc/astc.go
```diff
@@ -152,4 +157,81 @@ func init() {
...
+	for _, conversion := range compressionSupportMap {
+		// Intentional local copy
+		conv := conversion
+		image.RegisterConverter(conv.src, conv.dst, func(src []byte, width, height, depth int) ([]byte, error) {
+			return compressorFunc(src, width, height, depth, conv.dst)
+		})
+	}
+}
+
+func compressorFunc(src []byte, width int, height int, depth int, compressionFormat *image.Format) ([]byte, error) {
```
pmuetschard, line 201: compressorFunc -> compress
compressionFormat -> format or dstFormat

core/image/astc/astc.h
```diff
@@ -23,6 +23,10 @@ extern "C" {
 
 typedef int astc_error;
 
+int compress_astc(uint8_t* in, uint8_t* out, uint32_t width, uint32_t height,
```
pmuetschard, line 26: The return should be `astc_error`

core/image/astc_format.go
```diff
@@ -42,3 +45,42 @@ func (f *FmtASTC) check(data []byte, w, h, d int) error {
 func (*FmtASTC) channels() stream.Channels {
 	return stream.Channels{stream.Channel_Red, stream.Channel_Green, stream.Channel_Blue, stream.Channel_Alpha}
 }
+
+func ASTCFrom(data []byte) (*Data, error) {
```
pmuetschard, line 49: Add a doc string.

core/image/compress_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	imageASTC, err := image.ASTCFrom(imageData)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to read '%s': %v", name, err)
+	}
+
+	return imageASTC, nil
+}
+
+func getPlainImage(name string) (*image.Data, error) {
```
pmuetschard, line 85: `getInputImage`

core/image/BUILD.bazel
yalcinmelihyasin: the thing is because there is a package with same name, debugger gets confused where to jump when you try to debug. I changed so it can parse the package name correctly when the go runtime was calling "astc.init()". The other option is renaming the package. We can rename it do "astc_encoder" or something as well. As long as they have different name, both are fine for me.

core/image/compress_test.go
yalcinmelihyasin: Actually not, because this function should get a "plain" non compressed image. When it's named input image, it does not clarify that. With more test added, this function will remain as a function only loads a plain, non compressed image where as another function will grow to load images with different formats.

core/image/BUILD.bazel
yalcinmelihyasin: I have tried the rename the package, it requires a lot name change and it actually causes we call "astc_encoder.foo()" instead of "astc.foo()". So, if it is not very important to keep this astc.go I suggest it to be astc_format(or any other name) on a parallel work I am actually planning to do the same for etc1 & etc2 as well. I think being able to debug the package astc overcomes the a bit longer and unusual naming of a single source file.

pmuetschard: I don't know what you mean with debugger jumping, but don't do any renames - package or file - but file a bug to the debugger and will live with that for now.

core/image/astc/astc.go
pmuetschard: You haven't addressed this.

core/image/astc_format.go
```diff
@@ -42,3 +45,44 @@ func (f *FmtASTC) check(data []byte, w, h, d int) error {
 func (*FmtASTC) channels() stream.Channels {
 	return stream.Channels{stream.Channel_Red, stream.Channel_Green, stream.Channel_Blue, stream.Channel_Alpha}
 }
+
+// ASTCFrom reads a raw astc image, extracts the header and creates an
```
pmuetschard, line 49: Mention here that it is expecting the image with a header, which is a format that is only used when storing the images in our test data files. I think the astc_encoder CLI tool can output images in this format.

core/image/compress_test.go
pmuetschard: plain does not mean uncompressed. For a compression test, input is naturally uncompressed, but if you want it to be `getUncompressedImage` that would be fine, too.

core/image/BUILD.bazel
yalcinmelihyasin: The thing is Go has a special condition for init functions. it's called by go runtime. And when the call comes to delve it comes as "core/image/astc.init()" and at this point delve does not if it's init function in astc.go or init function in astc package. Why does it has not come as "core/image/astc/astc.init()" or why delve try to show an init function "core/image/astc.go" when it does not exist. But in any case, it makes it hard to debug anything in the particular init function. Which how I got realise the issue.

I don't know how it is benefit us to keep the filename same. I can try to put a bug report for delve but I don't know when it would be fixed-or would it be fixed.

Simply, anytime we have a bug in compression or decompression, delve will be confused. And simply everytime it happens to me I'll rename the file, compile, solve and rename back to astc.go. Which I don't know if this is a necessary effort to just keep the filenames same as before.

core/image/astc/astc.go
yalcinmelihyasin: this is done

core/image/astc_format.go
yalcinmelihyasin: Yes it can output with this format. 

Currently I am returning the images without the header simply because I don't know how we will use the images yet. If we require to have the header, I can add the header as well that's not a big issue. My guess is we will not need as we already have the header information when compressing but if we need, it's not hard to add.

pmuetschard (approved):

PR #639, opened 2021-01-18
Alternative take on context cancellation and mutate

hevrard:


hevrard:
This approach may add to the object churning, we need to perf measurement to see the impact (but I'm currently facing issues when I try to use `gapit benchmark`)

hevrard:
Discussion notes: keep this approach, but:
- move creation of the "ignore cancel context" at the top of Mutate functions
- in the database code: hoist the check for stopreason in above (separate commit)
- replace MustRead/MustWrite (and siblings like MustReadSlice etc) in all code by regular Read / Write with error checking
- rename MakeUncancellable to something like "IgnoreCancel"
- staterebuilder is very very unlikely to be cancelled, so it's OK to not treat the MustReadSlice etc... for now.

hevrard:
I've updated the whole PR with 6 commits, to be reviewed in isolation.

Most commit are small, expect for the two commits that replace MustRead/MustWrite by Read/Write: these are big-but-boring commits that just adds error checking propagation.

Bug: b/149701723

hevrard:
Fix #59

johnplate (approved):
The implementation in detail LGTM. I recommend to wait for another approval from someone who is more familiar with possible side-effects of a context that ignores cancellation and if is is safe not to check if the context was cancelled in every resolve.

AWoloszyn:
Seems ok to me.

hevrard:
Fix #59

pmuetschard (approved):

PR #643, opened 2021-01-20
Fix minor issue with cross-compile label in Shader View

hliatis:
Bug: b/178037398


johnplate (approved):

johnplate:
Forced CI to run again to check if Linux smoke test failure was a flake

PR #644, opened 2021-01-21
Avoid a possible NPE in setTimeMetrics method.

stellama0208:
This NPE was caught by a swarming session. Add a warning log
and a null check, instead of failing the server directly.

Bug: b/177526271.


hevrard (approved):
Thanks for this!

PR #647, opened 2021-01-22
Upgrade SWT version for macOS to fix UI bugs with the new macOS Big Sur.

stellama0208:
Bug: b/178132584, b/178133135.


hevrard:
LGTM in principle, but I don't have a mac to try this out, so I'll let Pascal approve this one.

pmuetschard:
I've sent you #650, since we should update JFace along with SWT, and I've had this change lying around some time now trying to update everything (#627). Can you test it with this change and see if things are better?

stellama0208:
Close this one with #650 being a more complete update.

PR #650, opened 2021-01-22
Update SWT and JFace.

pmuetschard:


stellama0208 (approved):
Tested and verified this update helps solve many minor UI issues on mac Big Sur.

PR #657, opened 2021-02-01
Signed to unsigned integer conversion

hliatis:
Bug: b/171392811

### DIFF
--- a/core/stream/convert.go
+++ b/core/stream/convert.go
@@ -183,7 +183,7 @@ func (m *mapping) conv(count int, min, max float64) error {
 			return intCast(count, m.dst, m.src)
 		}
 		// Source is normalized
-		if d.DataType.Signed == s.DataType.Signed {
+		if d.DataType.Signed == s.DataType.Signed || s.DataType.Signed {
 			if d.DataType.GetInteger().Bits > s.DataType.GetInteger().Bits {
 				return intExpand(count, m.dst, m.src)
 			}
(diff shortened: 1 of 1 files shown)


core/stream/convert.go
```diff
@@ -290,29 +299,59 @@ func intCollapse(count int, dst, src buf) error {
...
 	destStream := binary.BitStream{Data: dst.bytes, WritePos: dst.offset}
 	maxPossibleValue := math.Pow(2, float64(srcBitsExcSign)) - 1
+	maxWithSign := math.Pow(2, float64(srcBitsIncSign)) - 1
 	min, max := maxPossibleValue, float64(0)
 	if src.component.Channel == Channel_Depth {
 		for i := 0; i < count; i++ {
-			f := float64(sourceStream.Read(srcBitsExcSign)) / maxPossibleValue
+			f := float64(0)
+
+			if srcTy.Signed && !signed {
```
yalcinmelihyasin, line 308: This is a bit at the borderline of being nit but as we are not assigning any vaule to `signed` maybe it might be a good idea to change it `dstTy.Signed` for readablity. Then it's more clear that it's actually from signed to unsigned in one pass of reading.

core/stream/convert.go
```diff
@@ -290,29 +299,59 @@ func intCollapse(count int, dst, src buf) error {
...
 			if f > max {
 				max = f
 			}
+			sourceStream.ReadPos += src.stride - srcBitsIncSign
 		}
 		sourceStream.ReadPos = src.offset
 	}
 	for i := 0; i < count; i++ {
-		v := sourceStream.Read(srcBitsExcSign)
+		v := uint64(0)
```
yalcinmelihyasin, line 326: Readability: I wonder if it would be possible to have a separate function for each case rather than lot's of if together, or instead of checking signed condition everywhere, just checking it in the beginning and doing the rest under that single if. e.g. moving the channel check under the first if for both cases.

yalcinmelihyasin (commented):
I don't have a big objection. Just maybe it could be changed it to be more readable(although this might be subjective) otherwise it's good if another person approves.

core/stream/convert.go
hliatis: I agree. I only left it in there because I try to leave pre-existing stuff like that in there just in case.

core/stream/convert.go
hliatis: I'll need a second opinion on this.

yalcinmelihyasin (approved):

core/stream/convert.go
hliatis: Now that I think about it, I might be able to get rid of the of checks inside the depth channel block since I don't think there are any signed integer depth formats.

pmuetschard (approved):

pmuetschard (approved):

PR #659, opened 2021-02-03
Add early minimal version of frame boundary extension

hevrard:
The VK_EXT_frame_extension is still being discussed in Khronos, this
is an early minimal version that should be enough to capture Vulkan
workloads that do not use vkQueuePresentKHR (but that are still
preparing one frame at a time).

See the description in
gapis/api/vulkan/extensions/ext_frame_boundary.api

Bug: b/156105650

### DIFF
--- a/gapii/cc/call_observer.h
+++ b/gapii/cc/call_observer.h
@@ -183,10 +183,10 @@ class CallObserver : public context_t {
   void observeTimestamp();
 
   // setEndOfFrame marks this call as being an end of frame.
-  void setEndOfFrame() { mEndOfFrame = true; };
+  void setEndOfFrame() { mEndOfFrame = true; }
 
   // getEndOfFrame returns true if this call is marked as an end of frame.
-  bool getEndOfFrame() { return mEndOfFrame; };
+  bool getEndOfFrame() { return mEndOfFrame; }
 
  private:
   // shouldObserve returns true if the given slice is located in application
--- a/gapii/cc/call_observer.h
+++ b/gapii/cc/call_observer.h
@@ -182,6 +182,12 @@ class CallObserver : public context_t {
   // observeTimestamp encodes a timestamp extra in the trace
   void observeTimestamp();
 
+  // setEndOfFrame marks this call as being an end of frame.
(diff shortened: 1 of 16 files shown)


gapii/cc/call_observer.h
```diff
@@ -182,6 +182,12 @@ class CallObserver : public context_t {
   // observeTimestamp encodes a timestamp extra in the trace
   void observeTimestamp();
 
+  // setEndOfFrame marks this call as being an end of frame.
+  void setEndOfFrame() { mEndOfFrame = true; };
+
+  // getEndOfFrame returns true if this call is marked as an end of frame.
+  bool getEndOfFrame() { return mEndOfFrame; };
```
johnplate, line 189: Please remove the unnecessary semicolons at the end.

hevrard:
I open this PR for discussion, some remarks:

- this is a minimal version of the extension, in particular it does not let the app code indicate which images/buffers contain the end results of the "frame". Does HWUI actually needs this?

- some parts of gapis rely on the "endOfFrame" command flag / annotation which the current change is likely to break, e.g. in `transform_read_framebuffer.go:TransformCommand()` endOfFrame is used to know if a command is a vkQueuePresentKHR and conditionally mutate it or not. My current change adds `@frame_end` to vkQueueSubmit, so these logics will break. I think we might need to distinguish "is end of frame" from "is present" now, any thoughts on that?

- AGI implements this extension, not the underlying driver. This means AGI must advertise this extension, but also remove it at replay time. Currently the removal happens in `Vulkan::replayCreateVkDeviceImpl`, but we can also think about having a transform to do this, I'm not sure which option is best.

johnplate (commented):
Implementation so far LGTM, just one nit.

gapii/cc/call_observer.h
hevrard: Done, thanks!

pmuetschard:
I'm not sure this is the right approach. This tangles the queue submits with swapchain management by assuming that the code that issues the queue submits is actually aware of whether the submit is on a swapchain image, or, say, a temporary FB.

hevrard:
Superseded by #670

PR #662, opened 2021-02-05
Download and Install the ANGLE APK for the User

pmuetschard:
If we detect that there is a newer version of ANGLE available, provide an option to the user to request AGI to download and install the ANGLE apk.

### DIFF
--- a/core/os/android/adb/bind.go
+++ b/core/os/android/adb/bind.go
@@ -77,3 +77,9 @@ type binding struct {
 
 // verify that binding implements Device
 var _ Device = (*binding)(nil)
+
+// InstallApp implements the bind.Device interface, which can be implemented using
+// just the android.Device interface.
+func (b *binding) InstallApp(ctx context.Context, path string) error {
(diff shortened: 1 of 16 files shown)


stellama0208 (approved):
The code LGTM! also verified the feature by testing. 
I noticed that the first time after I installed the new update, the ANGLE install warning message will appear back a bit, but after restarting the AGI it will no longer appear. Will try to see whether could reproduce it after I find a way to uninstall ANGLE

stellama0208:
Just found that I could delete ANGLE by `adb uninstall org.chromium.angle`. To reproduce the outdated ANGLE warning behavior above:
1. Uninstall ANGLE. 
2. Start AGI, choose Capture Trace, and install ANGLE by the link inside the warning message.
3. Install successfully and the warning message disappear.
4. Close and restart the Capture Trace dialog.
5. The warning message re-appear. 
6. But If I restart the AGI session, the message will no longer show up.

PR #664, opened 2021-02-10
Implement GPU counter data's detail visualization in perftab.

stellama0208:
Add a canvas in the perftab to draw GPU counter detail curve. The curve
correspond to the selected cell in the performance table.
This detail counter curve surves mainly for 2 purposes:
1. Help improve those non-time based rate counter's aggregation.
With the detailed graph, users could notice a counter's spike inside a
long renderpass, otherwise the spike may gets flattened and loses attention
with merely the performance number.
2. Help AGI developers or early perftab users notice whether there's
calculation bugs on the performance number.

Bug: b/158057709.
Test: Manual test by openning the Performance(Experimental) tab.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/PerformanceView.java
+++ b/gapic/src/main/com/google/gapid/views/PerformanceView.java
@@ -15,6 +15,8 @@
  */
 package com.google.gapid.views;
 
+import static com.google.gapid.perfetto.views.StyleConstants.threadStateSleeping;
+import static com.google.gapid.perfetto.views.StyleConstants.mainGradient;
 import static com.google.gapid.util.Loadable.MessageType.Error;
 import static com.google.gapid.widgets.Widgets.createButton;
 import static com.google.gapid.widgets.Widgets.createComposite;
@@ -26,8 +28,10 @@ import static com.google.gapid.widgets.Widgets.createTreeColumn;
 import static com.google.gapid.widgets.Widgets.createTreeViewer;
 import static com.google.gapid.widgets.Widgets.packColumns;
 import static com.google.gapid.widgets.Widgets.withMargin;
+import static com.google.gapid.widgets.Widgets.withLayoutData;
 
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.gapid.models.Capture;
 import com.google.gapid.models.CommandStream;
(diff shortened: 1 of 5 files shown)


gapis/service/service.proto
```diff
@@ -1323,18 +1323,22 @@ message ProfilingData {
...
     }
 
     // Perf includes a best-guessing performance value and a confidence range.
     message Perf {
       double estimate = 1;
       double min = 2;
       double max = 3;
+      map<int32, double> estimateSamples = 4;  // {index} -> {sample weight}
+      map<int32, double> minSamples = 5;       // {index} -> {sample weight}
+      map<int32, double> minSamples = 5;       // {index} -> {sample weight}
```
hevrard, line 1341: It looks like this duplicate line makes the CI builds fail.

stellama0208: Oops, included this when I was desperately solving the presubmit format issues exposed by the remote kokoro result. Just fixed, thanks for noticing!

gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -547,4 +598,167 @@ protected void okPressed() {
...
+    public CounterDetailUi(Composite parent, Theme theme) {
+      super(parent, SWT.NONE);
+
+      setLayout(withMargin(new GridLayout(2, false), 0, 5));
+      Label titleLabel = withLayoutData(createLabel(this, "GPU Counter Detail Graph: "),
+          new GridData(SWT.LEFT, SWT.FILL, true, false, 2, 1));
+      titleLabel.setFont(theme.selectedTabTitleFont());
+      coordinateLabel = withLayoutData(createLabel(this, ENTRY_LABEL_PREFIX + "; " + METRIC_LABEL_PREFIX),
+          new GridData(SWT.LEFT, SWT.FILL, true, false));
+      withLayoutData(createLabel(this, "* (counter sample weight). ** Only for GPU counters."),
```
hevrard, line 622: Are these stars `*` supposed to appear in other labels in the UI (I didn't see them anywhere)? Or are they "footnotes" for the Detail Graph (in which case maybe we can just put this in parenthesis, without stars)?

gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -547,4 +598,167 @@ protected void okPressed() {
...
+          ctx.drawLine(right + AXIS_ARROW_LENGTH, bottom, right + AXIS_ARROW_LENGTH - AXIS_ARROW_SIZE, bottom + AXIS_ARROW_SIZE);
+          ctx.drawLine(right + AXIS_ARROW_LENGTH, bottom, right + AXIS_ARROW_LENGTH - AXIS_ARROW_SIZE, bottom - AXIS_ARROW_SIZE);
+          double x = left;
+          for (int i : indexes) {
+            // A counter sample at time {tn} is seen as holding the value {vn} during the implicit
+            // time range between {tn-1} ~ {tn}.
+            if (i > 0) {
+              ctx.drawText(Fonts.Style.Normal, Unit.NANO_SECOND.format(
+                  counter.getTimestamps(i - 1) - traceStart), x, bottom + AXIS_ARROW_SIZE);
+            }
```
hevrard, line 743: Looks like `counter.getTimestamps(i - 1) - traceStart` may be negative, is that expected?

perftab-negative-start

hevrard (approved):
LGTM with a few questions.

gapic/src/main/com/google/gapid/views/PerformanceView.java
stellama0208: Yeah, my idea was to use it as a "footnotes", we may need a consistent UX for providing explanations I guess. And for the parenthesis, with `(counter sample weight)`, I was trying to say that in the graph, the number inside parenthesis is the counter weight, so there's both star and parenthesis. might not be a good footnote I guess...

gapic/src/main/com/google/gapid/views/PerformanceView.java
stellama0208: Yes, this is expected. ;) the time start based on the start time of the first gpu slice, but the counter samples covers longer time. this exactly counter sample covers partly with the first gpu slice so I still count its contribution. 
But considering that this may cause confusion, I may need to consult Jay for better practices on UX, also for the above explanation issue, and address them later in future PR.

gapic/src/main/com/google/gapid/views/PerformanceView.java
hevrard: OK, thanks for the clarification. I think it's OK to merge like this, we can polish afterwards.

gapic/src/main/com/google/gapid/views/PerformanceView.java
hevrard: OK to fix/polish that later.

stellama0208:
The above force push is only to rebase PR onto master tot, and resolve merge conflict in `gapis/service/service.proto` file.

PR #672, opened 2021-02-12
Make the sample use frame boundary

hevrard:
@pmuetschard this may be useful to you. I don't think we want to merge this, but I might work on a PR that adds a flag to make the sample use an AHB-backed swapchain.

### DIFF
--- a/cmd/vulkan_sample/main.cpp
+++ b/cmd/vulkan_sample/main.cpp
@@ -41,6 +41,10 @@
 
 #include "vulkan/vulkan.h"
 
+#if defined(__ANDROID__)
+typedef void (VKAPI_PTR *PFN_vkFrameBoundaryANDROID)(VkDevice device, VkSemaphore semaphore, VkImage image, AHardwareBuffer* buffer);
+#endif
+
 const uint32_t kBufferingCount = 2;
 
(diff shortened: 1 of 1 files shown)


cmd/vulkan_sample/main.cpp
```diff
@@ -749,6 +760,9 @@ int main(int argc, const char** argv) {
   VkFence ready_fences[kBufferingCount];
   VkSemaphore swapchain_image_ready_semaphores[kBufferingCount];
   VkSemaphore render_done_semaphores[kBufferingCount];
+#if defined(__ANDROID__)
+  VkSemaphore frame_boundary_done_semaphores[kBufferingCount];
+#endif
```
pmuetschard, line 765: Why the extra semaphore? We could just use `render_done_semaphore`.

hevrard: Because `render_done_semaphore` wait is already done on vkQueuePresent (and binary semaphores cannot have more than one wait per signal).

pau-baiget:
Is this PR moving forward at some point?

hevrard:
Nope, this PR was just a helper to test another one. Closing.

PR #673, opened 2021-02-12
Make memory struct view handles pnext pointers better.

stellama0208:
The behavior described in the bug ticket is resulted from several layers
of bugs. This commit tries to address the 1st and the 2nd layer.
1. StructNode hiding. Sometimes there may be multiple StructNodes with the
same root address, when there's a pointer field pointing to this address,
before this PR, there would be only one node getting appended. Fixed now to
append all of those nodes.
2. StructNode duplicating. When we try to `simplifyTrees`, we would call
`appendPointedNodes` both iteratively and recursively. Under certain conditions
one node would be treated with this method multiple times. This commit includes
a new `treated` variable to record the status, and thus avoid this problem.
3. After the above two changes, we could still notice the fields pnext pointing
to have several information in common. More work could be done to check whether
it is intended behavior and whether we should combine them. This part is likely
to happen at the server side.

Bug: b/179943828.
Test: manual test in Memory->Struct tab, non-nil pnext node could be found sometimes
in vkCreateDevice command or vkCreateInstance command.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Memory.java
+++ b/gapic/src/main/com/google/gapid/models/Memory.java
@@ -649,16 +649,21 @@ public class Memory extends DeviceDependentModel<Memory.Data, Memory.Source, Voi
      */
     public static List<StructNode> simplifyTrees(List<StructNode> trees) {
       // Remove redundant layers.
-      Map<Long, StructNode> nodes = new HashMap<Long, StructNode>();
+      List<StructNode> allNodes = Lists.newArrayList();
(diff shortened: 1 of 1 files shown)


pmuetschard:
Thanks for looking into this. I will take a closer look next week, but I agree that de-duping the "header struct" with the "actual struct" would be server side. I think it would even be acceptable to make a special case for it in the server, since it's such a common occurrence, and filter out the header ones.

pmuetschard (approved):

PR #674, opened 2021-02-12
Add Experiments to the UI

yalcinmelihyasin:
Adds Experiments to the dropdown menu of the command tree and the Experimental Performance tab. This also allows users to run disable AF experiment. All the changes are behind a flag.


gapic/src/main/com/google/gapid/models/Experiments.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gapid.models;
+
+public class Experiments {
+  public static final boolean ShouldEnableExperiments = false;
```
pmuetschard, line 19: constants should use uppercase snakecase.

gapic/src/main/com/google/gapid/models/Experiments.java
```diff
@@ -0,0 +1,34 @@
...
+
+  boolean shouldDisableAF;
+
+  public Experiments() {
```
pmuetschard, line 23: Make this class immutable. All fields can be `public final`.

gapic/src/main/com/google/gapid/models/Profile.java
```diff
@@ -161,6 +163,19 @@ protected void onUiThread(CommandStream.Node node) {
...
+  public Experiments getExperiments() {
+    return experiments;
+  }
+
+  public void setExperiments(Experiments experiments) {
+    this.experiments = experiments;
+  }
+
+  public void reload() {
+    // Melih TODO: Add experiments to source and set this to false
```
pmuetschard, line 175: Do this now. Simply put it in the class below, fix the equals and hashcode and you should be good to go. Then, in the `setExperiments` method update the source. Take a look at `updateSemantics` in the `Geometries` model for an example.

gapic/src/main/com/google/gapid/views/CommandExperiments.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Logger;
+
+public class CommandExperiments {
+  private static final Logger LOG = Logger.getLogger(CommandEditor.class.getName());
+  private final Models models;
+  private final Theme theme;
+
+  public CommandExperiments(Client client, Models models, Theme theme) {
```
pmuetschard, line 45: You are not using the client here, which you shouldn't anyways in the view package.

gapic/src/main/com/google/gapid/server/Client.java
```diff
@@ -239,13 +241,16 @@ public void setGapidClient(GapidClient client) {
...
+  public ListenableFuture<Service.ProfilingData> profile(Path.Capture capture, Path.Device device, Experiments experiments) {
     return call(() -> String.format(
         "RPC->profile(%s, %s)", shortDebugString(capture), shortDebugString(device)),
         stack -> MoreFutures.transformAsync(
             client.profile(Service.GpuProfileRequest.newBuilder()
                 .setCapture(capture)
                 .setDevice(device)
+                .setExperiments(ProfileExperiments.newBuilder()
+                    .setDisableAnisotropicFiltering(experiments.getShouldDisableAF())
+                    .build())
```
pmuetschard, line 253: Add a `toProto` method to the `Experiments` class.

gapic/src/main/com/google/gapid/models/Experiments.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gapid.models;
+
+public class Experiments {
```
pmuetschard, line 18: Rename to `ProfileExperiments`

gapic/src/main/com/google/gapid/views/CommandExperiments.java
```diff
@@ -0,0 +1,107 @@
...
+      return "Profile Experiments";
+    }
+
+    @Override
+    protected Control createDialogArea(Composite parent) {
+      Composite area = (Composite)super.createDialogArea(parent);
+      Composite container = Widgets.createComposite(area, new GridLayout(2, false));
+      container.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+      Widgets.createLabel(container, "Disable Anisotropic Filtering");
+      experimentControls.put("disableAF",
```
pmuetschard, line 95: Just keep a reference to the checkbox in this instance, there is no need to go via a map.

gapic/src/main/com/google/gapid/views/CommandExperiments.java
```diff
@@ -0,0 +1,107 @@
...
+  private static final Logger LOG = Logger.getLogger(CommandEditor.class.getName());
+  private final Models models;
+  private final Theme theme;
+
+  public CommandExperiments(Client client, Models models, Theme theme) {
+    this.models = models;
+    this.theme = theme;
+  }
+
+  public static boolean shouldShowExperimentsPopup() {
```
pmuetschard, line 50: what's this for?

gapic/src/main/com/google/gapid/models/Experiments.java
```diff
@@ -0,0 +1,34 @@
...
+  public static final boolean ShouldEnableExperiments = false;
+
+  boolean shouldDisableAF;
```
pmuetschard, line 21: remove the `should` and don't abbreviate anisotropic filtering. The getter would still be `shouldDisable...`, but the field does not need it.

gapic/src/main/com/google/gapid/models/Experiments.java
yalcinmelihyasin: Even though it looks simpler with one experiment in it, I think that would make the use of this class very ugly after a few experiment added. Anytime we change an experiment settings, we would need to create new Experiments and copy all the other fields from the existing one which would end up with lots of unnecessary code. I think having this as a final but a mutable class is much cleaner.

gapic/src/main/com/google/gapid/models/Profile.java
```diff
@@ -161,11 +161,28 @@ protected void onUiThread(CommandStream.Node node) {
...
+  public void updateExperiments(ProfileExperiments experiments) {
+    load(Source.withExperiments(getSource(), experiments), false);
+  }
+
   public static class Source {
     public final Path.Capture capture;
-
-    public Source(Path.Capture capture) {
+    public final ProfileExperiments experiments;
+    public Source(Path.Capture capture, ProfileExperiments experiments) {
```
pmuetschard, line 176: add a blank line

gapic/src/main/com/google/gapid/models/ProfileExperiments.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package com.google.gapid.models;
+
+import com.google.gapid.proto.service.Service;
+
+public final class ProfileExperiments {
+  public static final boolean SHOULD_ENABLE_EXPERIMENTS = false;
+
+  final boolean disableAnisotropicFiltering;
```
pmuetschard, line 23: Either make this public and remove the getter, or make it private.

gapic/src/main/com/google/gapid/models/ProfileExperiments.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public ProfileExperiments() {
+    this(false);
+  }
+
+  public ProfileExperiments(boolean disableAnisotropicFiltering) {
+    this.disableAnisotropicFiltering = disableAnisotropicFiltering;
+  }
+
+  public final boolean getShouldDisableAnisotropicFiltering() {
```
pmuetschard, line 33: just `shouldDisable...`

gapic/src/main/com/google/gapid/models/ProfileExperiments.java
```diff
@@ -0,0 +1,42 @@
...
+    return disableAnisotropicFiltering;
+  }
+
+  public final Service.ProfileExperiments toProto() {
+    return Service.ProfileExperiments.newBuilder()
+    .setDisableAnisotropicFiltering(disableAnisotropicFiltering)
+    .build();
```
pmuetschard, line 40: continuations should be indented 4 spaces.

gapic/src/main/com/google/gapid/models/ProfileExperiments.java
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gapid.models;
+
+import com.google.gapid.proto.service.Service;
+
+public final class ProfileExperiments {
+  public static final boolean SHOULD_ENABLE_EXPERIMENTS = false;
```
pmuetschard, line 21: This should be an experimental flag. See `Experimental.java`.

gapic/src/main/com/google/gapid/views/CommandExperiments.java
```diff
@@ -0,0 +1,96 @@
...
+      models.analytics.postInteraction(View.Commands, ClientAction.UpdateExperiments);
+      models.profile.updateExperiments(dialog.newExperiments);
+    }
+  }
+
+  public ProfileExperiments getExperiments() {
+    return this.models.profile.getExperiments();
+  }
+
+  public void SetDisableAF(boolean selected) {
```
pmuetschard, line 60: lowercase method name, don't abbreviate anisotropic filtering.

gapic/src/main/com/google/gapid/views/CommandExperiments.java
```diff
@@ -0,0 +1,96 @@
...
+      return "Profile Experiments";
+    }
+
+    @Override
+    protected Control createDialogArea(Composite parent) {
+      Composite area = (Composite)super.createDialogArea(parent);
+      Composite container = Widgets.createComposite(area, new GridLayout(2, false));
+      container.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+      Widgets.createLabel(container, "Disable Anisotropic Filtering");
+      disableAFControl = Widgets.createCheckbox(container, models.profile.getExperiments().getShouldDisableAnisotropicFiltering());
```
pmuetschard, line 86: There is a `createCheckbox` that takes in a label for the checkbox. It is preferred over this approach, as it makes the label clickable to toggle the checkbox. Just set the layout data on the control to span two columns in the layout.

gapic/src/main/com/google/gapid/views/CommandExperiments.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  public void SetDisableAF(boolean selected) {
+    models.analytics.postInteraction(View.Commands, ClientAction.DisableAF);
+    models.profile.updateExperiments(new ProfileExperiments(selected));
+  }
+
+  private static class ExperimentsDialog extends DialogBase {
+    private final Models models;
+    private Button disableAFControl;
```
pmuetschard, line 67: Call it `disableAnisotropicFiltering`. The type does not need to be part of the name.

gapic/src/main/com/google/gapid/views/CommandOptions.java
```diff
@@ -0,0 +1,100 @@
...
+
+    MenuItem editMenuItem = Widgets.createMenuItem(this.optionsMenu , "&Edit", SWT.MOD1 + 'E', e -> {
+      CommandStream.Node node = this.tree.getSelection();
+      if (node != null && node.getData() != null && node.getCommand() != null) {
+        this.widgets .editor.showEditPopup(this.optionsMenu.getShell(), lastCommand(node.getData().getCommands()),
+            node.getCommand(), node.device);
+      }
+    });
+
+    if(!ProfileExperiments.SHOULD_ENABLE_EXPERIMENTS) {
```
pmuetschard, line 50: whitespace between if and (.

Also in other places throughout.

gapic/src/main/com/google/gapid/views/CommandOptions.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    MenuItem experimentMenuItem = Widgets.createSubMenu(this.optionsMenu, "&Experiments", this.experimentsMenu);
+
+    MenuItem disableAF = Widgets.createCheckMenuItem(this.experimentsMenu, "&Disable Anisotropic Filtering", SWT.MOD1 + 'A', e -> {
+      CommandStream.Node node = this.tree.getSelection();
+      if (node != null && node.getData() != null) {
+        this.widgets.experiments.SetDisableAF(((MenuItem)e.widget).getSelection());
+      }
+    });
```
pmuetschard, line 75: Why is this still here, if we also have a dialog for it?

gapic/src/main/com/google/gapid/views/CommandExperiments.java
```diff
@@ -0,0 +1,96 @@
...
+  public ProfileExperiments getExperiments() {
+    return this.models.profile.getExperiments();
+  }
+
+  public void SetDisableAF(boolean selected) {
+    models.analytics.postInteraction(View.Commands, ClientAction.DisableAF);
+    models.profile.updateExperiments(new ProfileExperiments(selected));
+  }
+
+  private static class ExperimentsDialog extends DialogBase {
```
pmuetschard, line 65: I'm not sure we want to do this in a dialog, but I'm OK for it to start as such.

gapis/api/vulkan/transform_af_disabler.go
```diff
@@ -94,5 +95,7 @@ func (afDisabler *afDisablerTransform) disableAFInSamplerCreation(ctx context.Co
 		newCmd.AddWrite(w.Range, w.ID)
 	}
 
+	log.W(ctx, "AF Disabled")
```
pmuetschard, line 98: This should not be a warning.

gapis/service/service.proto
```diff
@@ -770,6 +770,9 @@ enum ClientAction {
   Overdraw = 55;
   Pin = 56;
   ClosePinned = 57;
+  ShowExperiments = 58;
+  UpdateExperiments = 59;
+  DisableAF = 60;
```
pmuetschard, line 775: `DisableAnisotropicFiltering`.

gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -119,6 +126,13 @@ public PerformanceView(Composite parent, Models models, Widgets widgets) {
     filterButton.setImage(widgets.theme.more());
     filterButton.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, false, false));
 
+    if(ProfileExperiments.SHOULD_ENABLE_EXPERIMENTS) {
```
pmuetschard, line 129: This button should be before the filter button, since the filter button goes along with the presets.

gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -119,6 +126,13 @@ public PerformanceView(Composite parent, Models models, Widgets widgets) {
...
+      Button experimentsButton =createButton(buttonsComposite, SWT.FLAT, "Experiments",buttonColor,
+      e -> widgets.experiments.showExperimentsPopup(getShell()));
```
pmuetschard, line 131: fix whitespace.

gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -119,6 +126,13 @@ public PerformanceView(Composite parent, Models models, Widgets widgets) {
...
+      experimentsButton.setImage(widgets.theme.clipboard());
```
pmuetschard, line 132: Why the clipboard icon?

yalcinmelihyasin: It's just a placeholder. We haven't decided what to put yet and it's not visible to the user anyway. We can change when we decide what icon to put or not put at all.

gapic/src/main/com/google/gapid/views/CommandOptions.java
yalcinmelihyasin: To have a quick access. So that user does not have to open the settings popup every time to just enable one experiment. If we think that this is redundant we can remove it. I would suggest to get devrel opinion on this one.

pmuetschard: You are making it a dialog and then work around the fact that it is a dialog. Maybe it shouldn't be a dialog in the first place. As I mentioned elsewhere in the PR, I'm ok with keeping it a dialog for now, but remove this menu option from here, it makes no sense here.

gapic/src/main/com/google/gapid/views/PerformanceView.java
pmuetschard: I'd prefer we pick an icon now. How about the "science" icon from material. It's a flask, I think that works.

core/app/flags/experimental.go
```diff
@@ -23,7 +23,8 @@ import (
 //     2) be by default false/off
 //     3) be removed once the feature is no longer in experiment
 var (
-	EnableVulkanTracing = flag.Bool("experimental-enable-vulkan-tracing", false, "Enable the experimental feature Vulkan tracing.")
-	EnableAngleTracing  = flag.Bool("experimental-enable-angle-tracing", false, "Enable the experimental feature ANGLE tracing.")
-	EnablePerfTab       = flag.Bool("experimental-enable-perf-tab", false, "Enable the experimental feature Counter performance tab.")
+	EnableVulkanTracing      = flag.Bool("experimental-enable-vulkan-tracing", false, "Enable the experimental feature Vulkan tracing.")
+	EnableAngleTracing       = flag.Bool("experimental-enable-angle-tracing", false, "Enable the experimental feature ANGLE tracing.")
+	EnablePerfTab            = flag.Bool("experimental-enable-perf-tab", false, "Enable the experimental feature Counter performance tab.")
+	EnableProfileExperiments = flag.Bool("experimental-enable-profile-experiments", false, "Enable Profile Experiments.")
```
pmuetschard, line 29: I don't think we need this to be a server-side flag. Just make it a UI flag only for now.

gapic/src/main/com/google/gapid/models/ProfileExperiments.java
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gapid.models;
+
+import com.google.gapid.proto.service.Service;
+
+public final class ProfileExperiments {
+  public final boolean shouldDisableAnisotropicFiltering;
```
pmuetschard, line 21: Sorry, we're going around in circles here. Boolean variables do not need the `is`, `should`, `was`, etc prefix. Only use those prefixes in function names. This is similar to, for example `int foo` and `int getFoo()`. You wouldn't call the variable `getFoo`. So for example:
```
[public|private] final boolean disableAnisotropicFiltering;

public boolean shouldDisableAnisotropicFiltering {
  return disableAnisotropicFiltering;
}
```

Since, in our case, it's public, we don't need the getter.

gapic/src/main/com/google/gapid/views/CommandExperiments.java
```diff
@@ -0,0 +1,96 @@
...
+import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Shell;
+
+import java.util.logging.Logger;
+
+public class CommandExperiments {
```
pmuetschard, line 37: Why is this called `CommandExperiments`? Should it just be called `Experiments`?

gapic/src/main/com/google/gapid/views/CommandExperiments.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public void setDisableAnisotropicFiltering(boolean selected) {
+    models.analytics.postInteraction(View.Commands, ClientAction.DisableAnisotropicFiltering);
+    models.profile.updateExperiments(new ProfileExperiments(selected));
+  }
+
+  private static class ExperimentsDialog extends DialogBase {
+    private final Models models;
+    private Button disableAnisotropicFiltering;
+    public ProfileExperiments newExperiments;
```
pmuetschard, line 68: Drop the `new`.

gapic/src/main/com/google/gapid/views/CommandOptions.java
```diff
@@ -0,0 +1,103 @@
...
+
+    MenuItem editMenuItem = Widgets.createMenuItem(this.optionsMenu , "&Edit", SWT.MOD1 + 'E', e -> {
+      CommandStream.Node node = this.tree.getSelection();
+      if (node != null && node.getData() != null && node.getCommand() != null) {
+        this.widgets .editor.showEditPopup(this.optionsMenu.getShell(), lastCommand(node.getData().getCommands()),
+            node.getCommand(), node.device);
+      }
+    });
+
+    if(!Experimental.enableProfileExperiments(this.models.settings)) {
```
pmuetschard, line 53: Still missing spaces after `if`. Here and throughout. Look at the Google Java style guide.

gapic/src/main/com/google/gapid/views/CommandTree.java
```diff
@@ -91,6 +89,7 @@
   private final Label commandIdx;
   private final SelectionHandler<Control> selectionHandler;
   private final SingleInFlight searchController = new SingleInFlight();
+  private final CommandOptions commandPopup;
```
pmuetschard, line 92: You are not actually using this instance anywhere. So, I suggest:
1) make `CommandOptions` not instantiatable  (i.e. empty private constructor)
2) add a static void function in `CommandOptions` that has all the logic you currently have in the constructor.

gapic/src/main/com/google/gapid/views/PerformanceView.java
yalcinmelihyasin: Sure, I will update it.

gapic/src/main/com/google/gapid/views/CommandOptions.java
yalcinmelihyasin: Why don't we have a linter for these?

gapic/src/main/com/google/gapid/views/CommandOptions.java
```diff
@@ -0,0 +1,97 @@
...
+
+    MenuItem editMenuItem = Widgets.createMenuItem(optionsMenu , "&Edit", SWT.MOD1 + 'E', e -> {
+      CommandStream.Node node = tree.getSelection();
+      if (node != null && node.getData() != null && node.getCommand() != null) {
+        widgets.editor.showEditPopup(optionsMenu.getShell(), lastCommand(node.getData().getCommands()),
+            node.getCommand(), node.device);
+      }
+    });
+
+    if (!Experimental.enableProfileExperiments(models.settings)) {
```
pmuetschard, line 47: This duplicates code. Please just guard the creation of the extra items.

gapic/src/main/com/google/gapid/views/Experiments.java
```diff
@@ -0,0 +1,96 @@
...
+      return "Profile Experiments";
+    }
+
+    @Override
+    protected Control createDialogArea(Composite parent) {
+      Composite area = (Composite)super.createDialogArea(parent);
+      Composite container = Widgets.createComposite(area, new GridLayout(2, false));
+      container.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+      Widgets.createLabel(container, "Disable Anisotropic Filtering");
+      disableAnisotropicFiltering = Widgets.createCheckbox(container, models.profile.getExperiments().disableAnisotropicFiltering);
```
pmuetschard, line 86: Use the factory method that combines the label with the checkbox.

gapic/src/main/com/google/gapid/views/CommandOptions.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    MenuItem experimentMenuItem = Widgets.createSubMenu(optionsMenu, "&Experiments", experimentsMenu);
+
+    MenuItem disableAF = Widgets.createCheckMenuItem(experimentsMenu, "&Disable Anisotropic Filtering", SWT.MOD1 + 'A', e -> {
+      CommandStream.Node node = tree.getSelection();
+      if (node != null && node.getData() != null) {
+        widgets.experiments.setDisableAnisotropicFiltering(((MenuItem)e.widget).getSelection());
+      }
+    });
```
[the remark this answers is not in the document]
yalcinmelihyasin: I am not sure if we decided to remove this. I would want to ask devrel first.

gapic/src/main/com/google/gapid/views/CommandOptions.java
```diff
@@ -0,0 +1,69 @@
...
+    MenuItem editMenuItem = Widgets.createMenuItem(optionsMenu , "&Edit", SWT.MOD1 + 'E', e -> {
+      CommandStream.Node node = tree.getSelection();
+      if (node != null && node.getData() != null && node.getCommand() != null) {
+        widgets.editor.showEditPopup(optionsMenu.getShell(), lastCommand(node.getData().getCommands()),
+            node.getCommand(), node.device);
+      }
+    });
+
+    if (Experimental.enableProfileExperiments(models.settings)) {
+      MenuItem experimentMenuItem = Widgets.createMenuItem(optionsMenu, "&Experiments", SWT.MOD1 + 'S', e -> {
```
pmuetschard, line 48: Put the `&` in front of the accelerated letter, since that'll make it underlined. You need to use something other than `S`, though, since we already use that for saving captures. Maybe use `P` or none.

gapic/src/main/com/google/gapid/views/CommandOptions.java
```diff
@@ -0,0 +1,69 @@
...
+        CommandStream.Node node = tree.getSelection();
+        if (node != null && node.getData() != null) {
+          widgets.experiments.showExperimentsPopup(experimentsMenu.getShell());
+        }
+      });
+
+      experimentMenuItem.setEnabled(true);
```
pmuetschard, line 55: This is not needed, they are enabled by default.

gapic/src/main/com/google/gapid/views/Experiments.java
```diff
@@ -0,0 +1,95 @@
...
+    public String getTitle() {
+      return "Profile Experiments";
+    }
+
+    @Override
+    protected Control createDialogArea(Composite parent) {
+      Composite area = (Composite)super.createDialogArea(parent);
+      Composite container = Widgets.createComposite(area, new GridLayout(2, false));
+      container.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+      disableAnisotropicFiltering = Widgets.createCheckbox(container, "Disable Anisotropic Filtering", models.profile.getExperiments().disableAnisotropicFiltering);
```
pmuetschard, line 85: You need to make this widget take up two columns, by giving it a `GridData` object with a column span. Otherwise things won't line up correctly when you start adding new ones. See `Widgets.withSpans`.

pmuetschard (approved):

gapic/src/main/com/google/gapid/views/Experiments.java
pmuetschard: The default `GridData()` should be sufficient and these methods are intended to be statically imported to make it read better. Like this:
```
disableAnisotropicFiltering = withLayoutData(
    createCheckbox(...),
    withSpans(new GridData(), 2, 1));
```

pmuetschard (approved):

PR #676, opened 2021-02-15
Keep only one per-thread persistent CallObserver

hevrard:
This change removes the code related to stacking of call
observers (e.g. the logic involving mParent): that stacking was a
reminiscence of a graphics API (GVR I think) needs, where commands
could call other commands in a nested way.

In addition, rather than creating a new CallObserver each time we
enter a command, this change also use persistent, per-thread call
observers. They are stored in the thread-local `tl_callObserver`,
which used to be named `gContext`. This avoids to create and destroy a
CallObserver for each observed command.

This is based on a cherry-pick of
https://github.com/AWoloszyn/gapid/commit/6f66f91fe0b2c641d62d847042c82e555aea7ae8

Bug: b/151400181
Test: manual

### DIFF
--- a/gapii/cc/call_observer.cpp
+++ b/gapii/cc/call_observer.cpp
@@ -42,29 +42,28 @@ const size_t MEMORY_MERGE_THRESHOLD = 256;
 namespace gapii {
 // Creates a CallObserver with a given spy and applies the memory space for
 // observation data from the spy instance.
-CallObserver::CallObserver(SpyBase* spy, CallObserver* parent, uint8_t api)
+CallObserver::CallObserver(SpyBase* spy, uint8_t api)
     : mSpy(spy),
-      mParent(parent),
       mSeenReferences{{nullptr, 0}},
       mCurrentCommandName(nullptr),
       mObserveApplicationPool(spy->shouldObserveApplicationPool()),
       mApi(api),
       mShouldTrace(false),
       mCurrentThread(core::Thread::current().id()) {
-  // context_t initialization.
-  this->context_t::id = 0;
-  this->context_t::next_pool_id = &spy->next_pool_id();
-  this->context_t::globals = nullptr;
-  this->context_t::arena = reinterpret_cast<arena_t*>(spy->arena());
-  mShouldTrace = mSpy->should_trace(mApi);
-
-  if (parent) {
-    mEncoderStack.push(mShouldTrace ? parent->encoder() : mSpy->nullEncoder());
-  } else {
-    mEncoderStack.push(mSpy->getEncoder(mApi));
-  }
-
   mPendingObservations.setMergeThreshold(MEMORY_MERGE_THRESHOLD);
+  this->context_t::arena = reinterpret_cast<arena_t*>(mSpy->arena());
+  this->context_t::globals = nullptr;
+  this->context_t::id = 0;
+}
+
+void CallObserver::beginCommand(const char* name) {
+  this->context_t::next_pool_id = &mSpy->next_pool_id();
+  mShouldTrace = mSpy->should_trace(mApi);
+  mEncoderStack.push(mSpy->getEncoder(mApi));
+  mObserveApplicationPool = mSpy->shouldObserveApplicationPool();
+  mSeenReferences.clear();
+  mPendingObservations.clear();
+  mCurrentCommandName = name;
 }
 
 // Releases the observation data memory at the end.
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -65,7 +65,7 @@ const uint32_t kMaxFramebufferObservationHeight = 2560;
 
 const int32_t kSuspendIndefinitely = -1;
 
-thread_local gapii::CallObserver* tl_callObserver = nullptr;
+thread_local std::unique_ptr<gapii::CallObserver> tl_callObserver;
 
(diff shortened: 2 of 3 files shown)


johnplate (approved):

hevrard:
There was a bug caught by the swiftshader test on Linux build: the very first Spy::enter() was called with an api argument of value 0 (`enter("init", 0)`), I've updated this to use Vulkan's API index `Vulkan::kApiIndex` everywhere in non-generated code.

gapii/cc/call_observer.cpp
```diff
@@ -42,29 +42,28 @@ const size_t MEMORY_MERGE_THRESHOLD = 256;
...
-    mEncoderStack.push(mSpy->getEncoder(mApi));
-  }
+  this->context_t::id = 0;
+}
 
-  mPendingObservations.setMergeThreshold(MEMORY_MERGE_THRESHOLD);
+void CallObserver::beginCommand(const char* name) {
+  this->context_t::next_pool_id = &mSpy->next_pool_id();
+  mShouldTrace = mSpy->should_trace(mApi);
+  mEncoderStack.push(mSpy->getEncoder(mApi));
```
pmuetschard, line 62: I think this should only be done if `mShouldTrace` is true here.

gapii/cc/spy.cpp
```diff
@@ -239,18 +239,18 @@ Spy::~Spy() {
...
-  auto ctx = new CallObserver(this, gContext, api);
-  ctx->setCurrentCommandName(name);
-  gContext = ctx;
-  return ctx;
+  // This code assumes we only deal with one 'api' in the trace (that is the
+  // case in AGI, which only handles Vulkan): it creates a fresh observer
+  // (per-thread, since tl_callObserver is thread_local) for that API the first
+  // time and reuses it in subsequent calls.
+  if (tl_callObserver == nullptr) {
+    tl_callObserver = new CallObserver(this, api);
```
pmuetschard, line 247: I don't see a corresponding `delete` for this `new`. Is it done implicetly?

johnplate: I noticed that too. There is no implicit delete for the persistent CallObserver. The destructor won't be called, but the memory will be freed on application termination. This was ok for me, but it could actually lead to a problematic memory leak if threads are created frequently. So I propose to use a std::unique_ptr to store the CallObserver.

gapii/cc/spy.cpp
```diff
@@ -65,7 +65,7 @@ const uint32_t kMaxFramebufferObservationHeight = 2560;
 
 const int32_t kSuspendIndefinitely = -1;
 
-thread_local gapii::CallObserver* gContext = nullptr;
+thread_local gapii::CallObserver* tl_callObserver = nullptr;
```
johnplate, line 68:
```suggestion
thread_local std::unique_ptr<gapii::CallObserver> tl_callObserver;
```

gapii/cc/spy.cpp
johnplate, line 247:
```suggestion
  if (!tl_callObserver) {
    tl_callObserver.set(new CallObserver(this, api));
```

gapii/cc/spy.cpp
```diff
@@ -239,18 +239,18 @@ Spy::~Spy() {
...
-  return ctx;
+  // This code assumes we only deal with one 'api' in the trace (that is the
+  // case in AGI, which only handles Vulkan): it creates a fresh observer
+  // (per-thread, since tl_callObserver is thread_local) for that API the first
+  // time and reuses it in subsequent calls.
+  if (tl_callObserver == nullptr) {
+    tl_callObserver = new CallObserver(this, api);
+  }
+  tl_callObserver->beginCommand(name);
+  return tl_callObserver;
```
johnplate, line 250:
```suggestion
  return tl_callObserver.get();
```

gapii/cc/spy.cpp
```diff
@@ -239,18 +239,18 @@ Spy::~Spy() {
...
-  auto ctx = new CallObserver(this, gContext, api);
-  ctx->setCurrentCommandName(name);
-  gContext = ctx;
-  return ctx;
+  // This code assumes we only deal with one 'api' in the trace (that is the
+  // case in AGI, which only handles Vulkan): it creates a fresh observer
+  // (per-thread, since tl_callObserver is thread_local) for that API the first
+  // time and reuses it in subsequent calls.
+  if (!tl_callObserver) {
+    tl_callObserver.set(new CallObserver(this, api));
```
johnplate, line 247:
```suggestion
    tl_callObserver.reset(new CallObserver(this, api));
```

hevrard:
There's an issue with the encoder stack. I've just spent too much time trying to debug it, but I can't get round to fix all the things.

I think we should be good with pushing the spy->encoder in CallObserver constructor. Then all enter() and exit() should be balanced and properly push/pop encoders. I have the feeling we could even get rid of the encoder stack, since we don't have nested callObservers.

However, in the state serialization, there is an enter() without a matching exit():
https://github.com/google/agi/blob/e82fd1ae667fb06e09142076a447ac9c9a436bf3/gapii/cc/state_serializer.cpp#L25

I tried adding an exit() at the end of this function, but it leads to replay errors (invalid device) in the resulting traces (?).

Closing this as we have higher priority things to focus on. Let's revisit in due time.

PR #677, opened 2021-02-16
Handle a corner case where GPU counter spec is nil at the perftab backend.

stellama0208:
This bug is observed in our nightly swarming result.

Bug: b/180059345, b/180332153.


stellama0208:
This PR address the line where our mali swarming device recently fails at. Since I don't have a reproducible device with this bug, I will observe the swarming result after this PR gets approved and merged. 
Let's see whether this PR fixes the problem. Hopefully there're no other bugs behind this line's break!

hevrard (approved):
Would it make sense to add the same safety check on adreno side?

PR #681, opened 2021-02-17
Clarify dependency for builds in Linux zip archive

hevrard:
Bug: b/151914478


johnplate (approved):
Just a typo: depend instead of depends

PR #687, opened 2021-02-19
Allow suboptimal `vkQueuePresentKHR`

mattyclarkson:
Adds `vkResult` message conversion in the error message.
Helps quickly diagnose the crash reason.

### DIFF
--- a/cmd/vulkan_sample/main.cpp
+++ b/cmd/vulkan_sample/main.cpp
@@ -15,6 +15,7 @@
  */
 
 #include <math.h>
+
 #include <chrono>
 #include <cstdint>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


hevrard (approved):
Thanks for this! I'll wait to see what @pmuetschard says of it before merging.

pmuetschard (approved):
Personally, I feel the error code -> string conversion is a bit superfluous, given that this is supposed to be a simple sample app. I'd prefer to just have the error code included in the message instead, but I've no problem with this, since you've already done the work.

mattyclarkson:
Happy to nuke the error code conversion, if you want. I regexed the VkResult webpage.

PR #690, opened 2021-02-23
Prefer write typed observations over reads.

pmuetschard:
This adds a flag to the typed observation indicating whether they are write vs read observations. The server, however, currently still filters them and returns only one of them, but it will now prefer the write over the read.


stellama0208 (approved):
LGTM!

PR #691, opened 2021-02-24
Fix a corner case in memory block's page loading request.

stellama0208:
When request to load memory pages, in certain conditions, the left
memory size will get assigned to a negative value based on calculations.
In such case, we shouldn't emit a memory loading request for this
negative-sized page.
Add a value check to avoid this scenario.

Bug: b/181012589.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Memory.java
+++ b/gapic/src/main/com/google/gapid/models/Memory.java
@@ -204,7 +204,7 @@ public class Memory extends DeviceDependentModel<Memory.Data, Memory.Source, Voi
       List<ListenableFuture<Segment>> futures = Lists.newArrayList();
       futures.add(getPage(firstPage, getOffsetInPage(offset), PAGE_SIZE - getOffsetInPage(offset)));
       for (long page = firstPage + 1, left = length - PAGE_SIZE + getOffsetInPage(offset);
-          page <= lastPage; page++, left -= PAGE_SIZE) {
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/models/Memory.java
```diff
@@ -204,7 +204,7 @@ public long getEndAddress() {
       List<ListenableFuture<Segment>> futures = Lists.newArrayList();
       futures.add(getPage(firstPage, getOffsetInPage(offset), PAGE_SIZE - getOffsetInPage(offset)));
       for (long page = firstPage + 1, left = length - PAGE_SIZE + getOffsetInPage(offset);
-          page <= lastPage; page++, left -= PAGE_SIZE) {
+          page <= lastPage && left >= 0; page++, left -= PAGE_SIZE) {
```
pmuetschard, line 207: The logic to compute `left` seems correct. So it seems more like a garbage-in-garbage-out issue. Instead of guarding like that, we should figure what's causing it. I think `left` is only negative if `length` is negative or 0, if that's the case, we should check that `length` is valid instead/fix callers.

stellama0208: Indeed, just to confirm, when this happens, the previous call in the stack will have the `length` being 0.

stellama0208:
Close this PR since this bug is now addressed with another fix at #746.

PR #692, opened 2021-02-24
Add gapit trim_initial_state

hevrard:
The new gapit verb trim_initial_state takes a capture, records which
parts of the initial state are actually used by the capture commands,
and returns a new capture where some unused parts of the initial state
have been removed.

This first version removes unused descriptor sets and unused
pipelines.

This trimming should help reducing the memory consumed by GAPIS. On a
capture that has thousands of unused descriptor sets and pipelines,
the memory consumed by GAPIS after opening the trace in the UI drops
by 10% VIRT (total virtual memory) and 25% for RES (resident memory)
when using the trimmed version of the capture.

Bug: b/178699413, b/157557996

### DIFF
--- a/cmd/gapit/trim_initial_state.go
+++ b/cmd/gapit/trim_initial_state.go
@@ -1,0 +1,67 @@
+// Copyright (C) 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"io/ioutil"
+
+	"github.com/google/gapid/core/app"
+	"github.com/google/gapid/core/log"
+)
+
+type trimInitialStateVerb struct{ TrimInitialStateFlags }
+
+func init() {
+	verb := &trimInitialStateVerb{}
+	app.AddVerb(&app.Verb{
+		Name:      "trim_initial_state",
+		ShortHelp: "(WIP) Trims a gfx trace's initial state to the resources actually used by the commands",
+		Action:    verb,
+	})
+}
+
+func (verb *trimInitialStateVerb) Run(ctx context.Context, flags flag.FlagSet) error {
+	if flags.NArg() != 1 {
+		app.Usage(ctx, "Exactly one gfx trace file expected, got %d", flags.NArg())
+		return nil
+	}
+
+	client, capture, err := getGapisAndLoadCapture(ctx, verb.Gapis, verb.Gapir, flags.Arg(0), verb.CaptureFileFlags)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	c, err := client.TrimCaptureInitialState(ctx, capture)
+	if err != nil {
+		return log.Errf(ctx, err, "TrimCaptureInitialState(%v)", capture)
+	}
+
+	data, err := client.ExportCapture(ctx, c)
+	if err != nil {
+		return log.Errf(ctx, err, "ExportCapture(%v)", c)
+	}
+
+	output := verb.Out
+	if output == "" {
+		output = "initialStateTrimmed.gfxtrace"
+	}
+	if err := ioutil.WriteFile(output, data, 0666); err != nil {
+		return log.Errf(ctx, err, "Writing file: %v", output)
+	}
+	return nil
+}
--- a/gapis/api/test/test.go
+++ b/gapis/api/test/test.go
@@ -54,6 +54,11 @@ func (s *State) Root(ctx context.Context, p *path.State, r *path.ResolveConfig) 
 // or it can apply backward-compatibility fixes for older traces.
 func (*State) SetupInitialState(ctx context.Context, state *api.GlobalState) {}
 
+// TrimInitialState is needed to implement the State interface.
+func (*State) TrimInitialState(ctx context.Context, capturePath *path.Capture) error {
+	return nil
+}
+
 func (i Remapped) remap(cmd api.Cmd, s *api.GlobalState) (interface{}, bool) {
 	return i, true
 }
--- a/gapis/api/vulkan/vulkan.go
+++ b/gapis/api/vulkan/vulkan.go
@@ -113,13 +113,31 @@ func (s *State) TrimInitialState(ctx context.Context, capturePath *path.Capture)
 	postCmdCb := func(s *api.GlobalState, subCmdIdx api.SubCmdIdx, cmd api.Cmd) {
 		switch cmd := cmd.(type) {
 		case *VkFreeDescriptorSets:
-			count := cmd.DescriptorSetCount()
-			ds, err := cmd.PDescriptorSets().Slice(0, (uint64)(count), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			ds, err := cmd.PDescriptorSets().Slice(0, (uint64)(cmd.DescriptorSetCount()), s.MemoryLayout).Read(ctx, cmd, s, nil)
 			if err != nil {
-				panic("Read error")
+				panic(err)
 			}
 			for _, d := range ds {
 				descriptorSets[d] = struct{}{}
+			}
+
+		case *VkUpdateDescriptorSets:
+			// VkWriteDescriptorSet
+			writeinfos, err := cmd.PDescriptorWrites().Slice(0, (uint64)(cmd.DescriptorWriteCount()), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			if err != nil {
+				panic(err)
+			}
+			for _, wi := range writeinfos {
+				descriptorSets[wi.DstSet()] = struct{}{}
+			}
+			// VkCopyDescriptorSet
+			copyinfos, err := cmd.PDescriptorCopies().Slice(0, (uint64)(cmd.DescriptorCopyCount()), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			if err != nil {
+				panic(err)
+			}
+			for _, ci := range copyinfos {
+				descriptorSets[ci.SrcSet()] = struct{}{}
+				descriptorSets[ci.DstSet()] = struct{}{}
 			}
 
 		case *VkDestroyPipeline:
--- a/gapis/api/vulkan/vulkan.go
+++ b/gapis/api/vulkan/vulkan.go
@@ -78,6 +78,137 @@ func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
 	}
 }
 
+// TrimInitialState scans the capture commands to see which parts of the initial
+// state are actually used, and removes some unused parts from it.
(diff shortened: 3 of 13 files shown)


gapis/api/vulkan/vulkan.go
```diff
@@ -78,6 +78,137 @@ func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
...
+// to track which objects are accessed: this would avoid to re-encode state
+// tracking logic here. There might be some pitfalls though, e.g. when a command
+// just reads the handle of an object, the state watcher would not mark an
+// access to that object. For instance, when creating a derivate pipeline, a
+// VkPipeline handle is used in BasePipelineHandle, the API implementation reads
+// this handle, but does not access the corresponding object. So using
+// api.StateWatcher might need some wider design considerations.
+func (s *State) TrimInitialState(ctx context.Context, capturePath *path.Capture) error {
+	// Parts of the state we want to record the usage of.
+	descriptorSets := map[VkDescriptorSet]struct{}{}
```
yalcinmelihyasin, line 97: Should not this be `interface{}` instead `struct{}`?

gapis/api/vulkan/vulkan.go
```diff
@@ -78,6 +78,137 @@ func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
...
+	pipelines := map[VkPipeline]struct{}{}
+
+	// Record usage in initial state.
+	for _, i := range s.LastComputeInfos().All() {
```
yalcinmelihyasin, line 101: i here seems like an index variable even though it's not. Maybe we should give it a proper name or letter.

gapis/api/vulkan/vulkan.go
```diff
@@ -78,6 +78,137 @@ func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
...
+	for _, i := range s.LastDrawInfos().All() {
+		for _, d := range i.DescriptorSets().All() {
+			descriptorSets[d.VulkanHandle()] = struct{}{}
+		}
+		pipelines[i.GraphicsPipeline().VulkanHandle()] = struct{}{}
+	}
+
+	// Record usage in the trace commands
+	// top-level commands
+	postCmdCb := func(s *api.GlobalState, subCmdIdx api.SubCmdIdx, cmd api.Cmd) {
```
yalcinmelihyasin, line 113: Personal Opinion: I am usually not favor of long lambdas in function body(to me just capturing the necessary variables[which is what lambdas actually for imo] and passing it to the actual function is much more readable) but this is your personal choice. Just give a second thought :)

gapis/api/vulkan/vulkan.go
```diff
@@ -78,6 +78,137 @@ func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
...
+		switch cmd := cmd.(type) {
+		case *VkFreeDescriptorSets:
+			count := cmd.DescriptorSetCount()
+			ds, err := cmd.PDescriptorSets().Slice(0, (uint64)(count), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			if err != nil {
+				panic("Read error")
```
yalcinmelihyasin, line 119: Should this be panic or should we return error?

gapis/api/vulkan/vulkan.go
```diff
@@ -78,6 +78,137 @@ func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
...
+			pipelines[cmd.Pipeline()] = struct{}{}
+		}
+
+	}
+	// sub-commands
+	postSubCmdCb := func(state *api.GlobalState, subCmdIdx api.SubCmdIdx, cmd api.Cmd, i interface{}) {
+		vkState := GetState(state)
+		cmdRef, ok := i.(CommandReferenceʳ)
+		if !ok {
+			panic("In Vulkan, MutateWithSubcommands' postSubCmdCb 'interface{}' is not a CommandReferenceʳ")
```
yalcinmelihyasin, line 135: panic or error?

gapis/api/vulkan/vulkan.go
```diff
@@ -78,6 +78,137 @@ func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
...
+			}
+		}
+	}
+
+	// Remove unused parts.
+	var startSize int
+
+	startSize = s.DescriptorSets().Len()
+	for h := range s.DescriptorSets().All() {
+		if _, ok := descriptorSets[h]; !ok {
```
yalcinmelihyasin, line 187: so here the struct that we store is actually not used at all, it is just there because Go doesn't have sets(afaik) Can we just store a boolean or some other plain type?

yalcinmelihyasin (commented):
I think that it is good in general. I have minor comments/questions.

gapis/api/vulkan/vulkan.go
hevrard: This is a regular way to have a set in Go. Another classical approach is a map of booleans. I think using a map of `struct{}` is the regular way to have sets in our codebase. You can try `git grep 'struct{}{}` to see some examples.

Using `interface{}` is probably not a good idea since you could put anything in the map. With `struct{}`, the only thing you can put is `struct{}{}`.

gapis/api/vulkan/vulkan.go
hevrard: we cannot return an error in this callback, hence the panic. (Note that in these situtations we used to use "MustRead", which would panic on error anyway).

gapis/api/vulkan/vulkan.go
hevrard: This is really a "should never happen" thing, so panic is OK. We have the same pattern in the framegraph code.

gapis/api/vulkan/vulkan.go
hevrard: I think `struct{}` is better than `bool` since it is more clear in showing that we don't care about what we store in the map. Storing a bool may mean that you give a semantic to that boolean (i.e. you behave differently depending on that boolean value). With type `struct{}`, you have only one possible value (`struct{}{}`), so the only thing you can really test / care about is presence of the key in the map.

gapis/api/vulkan/vulkan.go
hevrard: Agreed, replaced with `ci` for compute info, and `di` for draw info.

gapis/api/vulkan/vulkan.go
hevrard: Yeah, the lambdas here are less than 20 lines each, I think it's OK.

cmd/gapit/trim_initial_state.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/google/gapid/core/log"
+)
+
+type trimInitialStateVerb struct{ TrimInitialStateFlags }
+
+func init() {
+	verb := &trimInitialStateVerb{}
+	app.AddVerb(&app.Verb{
+		Name:      "trim_initial_state",
+		ShortHelp: "(WIP) Trims a gfx trace's initial state to the resources actually used by the commands",
```
pmuetschard, line 32: Remove the WIP part. `gapit` is not meant for the faint of heart.

cmd/gapit/trim_initial_state.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/google/gapid/core/app"
+	"github.com/google/gapid/core/log"
+)
+
+type trimInitialStateVerb struct{ TrimInitialStateFlags }
+
+func init() {
+	verb := &trimInitialStateVerb{}
+	app.AddVerb(&app.Verb{
+		Name:      "trim_initial_state",
```
pmuetschard, line 31: Let's just call it `trim_state`, so it's shorter.

gapis/api/vulkan/vulkan.go
pmuetschard: Since these are not inline lambdas, this comment doesn't really apply. There is little difference between
`func foo() {}` and `foo := func() {}`

gapis/api/vulkan/vulkan.go
```diff
@@ -78,6 +78,137 @@ func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
...
+
+	// Record usage in the trace commands
+	// top-level commands
+	postCmdCb := func(s *api.GlobalState, subCmdIdx api.SubCmdIdx, cmd api.Cmd) {
+		switch cmd := cmd.(type) {
+		case *VkFreeDescriptorSets:
+			count := cmd.DescriptorSetCount()
+			ds, err := cmd.PDescriptorSets().Slice(0, (uint64)(count), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			if err != nil {
+				panic("Read error")
```
pmuetschard, line 119: Can you panic the error, instead. Or use `log.Errf` and panic its returned error.

gapis/api/vulkan/vulkan.go
```diff
@@ -78,6 +78,137 @@ func (s *State) SetupInitialState(ctx context.Context, state *api.GlobalState) {
...
+		for _, d := range di.DescriptorSets().All() {
+			descriptorSets[d.VulkanHandle()] = struct{}{}
+		}
+		pipelines[di.GraphicsPipeline().VulkanHandle()] = struct{}{}
+	}
+
+	// Record usage in the trace commands
+	// top-level commands
+	postCmdCb := func(s *api.GlobalState, subCmdIdx api.SubCmdIdx, cmd api.Cmd) {
+		switch cmd := cmd.(type) {
```
pmuetschard, line 114: This is missing handling of `VkCopyDescriptorSet` and `VkWriteDescriptorSet`.

gapis/server/server.go
```diff
@@ -303,6 +303,20 @@ func (s *server) SplitCapture(ctx context.Context, rng *path.Commands) (*path.Ca
...
+func (s *server) TrimCaptureInitialState(ctx context.Context, p *path.Capture) (*path.Capture, error) {
+	ctx = log.Enter(ctx, "TrimCaptureInitialState")
+	c, err := capture.ResolveGraphicsFromPath(ctx, p)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, api := range c.APIs {
+		c.InitialState.APIs[api].TrimInitialState(ctx, p)
+	}
```
pmuetschard, line 315: This appears to modify the capture. I think it would be better if we cloned the initial state, then trimmed the cloned state, and created a new capture with the trimmed state. Especially since that's what the comment in the service.proto says :)

cmd/gapit/trim_initial_state.go
hevrard: Done, renamed the file to `trim_state.go`

cmd/gapit/trim_initial_state.go
hevrard: done

gapis/api/vulkan/vulkan.go
hevrard: done

gapis/server/server.go
hevrard: indeed! done

pmuetschard (approved):

gapis/api/test/test.go
```diff
@@ -54,6 +54,11 @@ func (s *State) Root(ctx context.Context, p *path.State, r *path.ResolveConfig)
 // or it can apply backward-compatibility fixes for older traces.
 func (*State) SetupInitialState(ctx context.Context, state *api.GlobalState) {}
 
+// TrimInitialState is needed to implement the State interface.
```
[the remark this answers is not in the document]
hevrard: Woops, sorry just saw that after merging.

PR #695, opened 2021-02-26
Filter similar pnext pointer memory structs caused by type castings.

stellama0208:
In memory struct view, when we expand a pnext pointer, sometimes we may see
multiple structs containing very similar information but holds different titles.

1. Issue: this behavior may cause reading difficulty for users.
2. Reason: it is related to how we deal with pnext pointers in the api files,
we cast them into different intermediate types so that we have these highly
correlated memory observations.
3. Solution: we filter these structs inside gapis' memory resolving grpc session.
The precision of these similar structs typically follows as: const void* ->
VkStructureType*-> VkStructHeader* -> Vk<.*>. The latter the preciser. We will
try to filter and only keep the last memory struct we could see in this chain.

Bug: b/179943828.

### DIFF
--- a/gapis/resolve/memory.go
+++ b/gapis/resolve/memory.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/google/gapid/core/app/analytics"
 	coreid "github.com/google/gapid/core/data/id"
@@ -41,10 +42,11 @@ var (
 )
 
 // filterTypedRanges takes a list of typed ranges and
-//   combines ranges that have identical bases and types
-func filterTypedRanges(ranges []*service.TypedMemoryRange) []*service.TypedMemoryRange {
+//   combines ranges that have identical bases and types, and
+//   filters some highly correlated structs into smaller number.
+func filterTypedRanges(ranges []*service.TypedMemoryRange) ([]*service.TypedMemoryRange, error) {
 	if len(ranges) == 0 {
-		return ranges
+		return ranges, nil
 	}
 	sort.Slice(ranges, func(i, j int) bool {
 		return (ranges[i].Root < ranges[j].Root ||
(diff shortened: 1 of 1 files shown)


gapis/resolve/memory.go
```diff
@@ -85,7 +88,82 @@ func filterTypedRanges(ranges []*service.TypedMemoryRange) []*service.TypedMemor
...
+	}
+	return filteredRanges, nil
+}
+
+// Return the pnext pointer derived struct's precise level.
+// The larger the number, the more precise and specific the struct is. Return -1
+// for irrelevant structs.
+// The ranking is based on how the pnext pointers are type casted in api files,
+// a typical casting chain could be found at gapis/api/vulkan/api/device.api.
+func pnextPreciseLevel(typeIndex uint64) (int, error) {
```
pmuetschard, line 150: Call the function `pNextPrecision`

gapis/resolve/memory.go
```diff
@@ -85,7 +88,82 @@ func filterTypedRanges(ranges []*service.TypedMemoryRange) []*service.TypedMemor
...
+	filteredRanges := []*service.TypedMemoryRange{}
+	for i := 0; i < len(newRanges); i++ {
+		if shouldDelete := toDelete[i]; !shouldDelete {
+			filteredRanges = append(filteredRanges, newRanges[i])
+		}
+	}
+	return filteredRanges, nil
+}
+
+// Return the pnext pointer derived struct's precise level.
```
pmuetschard, line 145: Go documentation requires the first sentence to start with the name of the thing being documented.

pmuetschard (approved):

gapis/resolve/memory.go
stellama0208: Done in the new commit.

gapis/resolve/memory.go
stellama0208: Done in the new commit. Good to know that!

PR #700, opened 2021-03-05
Actually properly support defines in the API files.

pmuetschard:


hevrard (commented):
the code LGTM but the CI-linux-test fails:

```
FAIL: //gapil/bapi:go_default_test (see /tmpfs/tmp/bazel_out/execroot/gapid/bazel-out/k8-opt/testlogs/gapil/bapi/go_default_test/test.log)
INFO: From Testing //gapil/bapi:go_default_test:
==================== Test output for //gapil/bapi:go_default_test:
=== RUN   TestEncodeDecode
    TestEncodeDecode: testing.go:51: 01:10:26.311 E: Error:apis
            ⟦*semantic.Definition⟧ != ⟦*semantic.DefinitionUsage⟧ for v.Field<0>.Field<1>[196].Field<1>.Block.Statements[0].True.Statements[1].Local.Value.Target.Function.Block.Statements[0].Value.Value.RHS·type
            ⟦*semantic.Definition⟧ != ⟦*semantic.DefinitionUsage⟧ for v.Field<0>.Field<1>[196].Field<1>.Block.Statements[0].True.Statements[2].Local.Value.Target.Function.Block.Statements[0].Value.Value.RHS·type
            ⟦*semantic.Definition⟧ != ⟦*semantic.DefinitionUsage⟧ for v.Field<0>.Field<1>[197].Field<1>.Block.Statements[2].Block.Statements[0].Condition.RHS·type
            ⟦*semantic.Definition⟧ != ⟦*semantic.DefinitionUsage⟧ for v.Field<0>.Field<1>[787].Field<1>.Block.Statements[0].Target.Function.Block.Statements[7].Block.Statements[0].False.Statements[1].Block.Statements[0].True.Statements[0].Block.Statements[0].True.Statements[0].Local.Value.Value.RHS·type
            ⟦*semantic.Definition⟧ != ⟦*semantic.DefinitionUsage⟧ for v.Field<0>.Field<1>[787].Field<1>.Block.Statements[0].Arguments[3]·type
            ⟦*semantic.Definition⟧ != ⟦*semantic.DefinitionUsage⟧ for v.Field<0>.Field<1>[794].Field<1>.Block.Statements[0].Cases[40].Block.Statements[0].Target.Function.Block.Statements[7].Block.Statements[0].Target.Function.Block.Statements[0].Condition.RHS·type
            ⟦*semantic.Definition⟧ != ⟦*semantic.DefinitionUsage⟧ for v.Field<0>.Field<1>[794].Field<1>.Block.Statements[0].Cases[42].Block.Statements[0].Target.Function.Block.Statements[3].Block.Statements[0].Target.Function.Block.Statements[0].Condition.RHS·type
--- FAIL: TestEncodeDecode (4.60s)
FAIL
```

gapil/bapi/bapi.proto
```diff
@@ -420,6 +420,7 @@ message Expression {
     uint64 clone = 9;              // Clone ID
     uint64 create = 10;            // Create ID
     uint64 definition = 11;        // Definition ID
+    uint64 definition_usage = 43;  // Definition Usage ID
```
hevrard, line 423: Should this be `12` and all subsequent fields be incremented?

I know using `43` would be the right way to keep backward compatibility, but looking at the history of such proto definitions in GAPID/AGI, it looks like so far the approach has been to keep alphabetical order and change field numbers to match that.

I guess we might want to define on a project-wide rule for such proto updates.

hevrard (approved):
LGTM assuming this fixes the CI

gapil/bapi/bapi.proto
pmuetschard: Proto fields shouldn't really be re-numbered. We've done it in the past, because we just didn't care.

PR #702, opened 2021-03-05
Fix an infinite loop in the memory struct resolve.

pmuetschard:
If the void* observation was the last one, the code would never increment the loop counter passed the just processed observation and would get stuck.

### DIFF
--- a/gapis/resolve/memory.go
+++ b/gapis/resolve/memory.go
@@ -104,8 +104,7 @@ func filterTypedRanges(ranges []*service.TypedMemoryRange) ([]*service.TypedMemo
 	// and identify whether some are less precise and excessive structs caused by
 	// pnext pointer type castings. If found, filter them out.
 	toDelete := map[int]bool{}
-	index := 0
-	for index < len(newRanges) {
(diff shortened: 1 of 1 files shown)


gapis/resolve/memory.go
```diff
@@ -139,7 +139,10 @@ func filterTypedRanges(ranges []*service.TypedMemoryRange) ([]*service.TypedMemo
 					mostPreciseLevel = preciseLevel
 				}
 			}
-			index = dups[len(dups)-1] + 1
+			index++
+			if skip := dups[len(dups)-1] + 1; skip > index {
+				index = skip
+			}
```
stellama0208, line 145: Thanks for finding this corner case!! 
After another look, I find I have another thing missed here about array index check. Maybe this part should be further updated with something like:
```suggestion
			if len(dups) > 0 && dups[len(dups)-1] + 1 > index {
				index = dups[len(dups)-1] + 1
			}
```

stellama0208 (approved):
LGTM, thanks for fixing, this part looks really neat now!

PR #704, opened 2021-03-09
Disable drawcall UI

yalcinmelihyasin:
Adds the UI and related server work to disable or isolate a drawcall, renderpass or a secondary command buffer.


gapic/src/main/com/google/gapid/models/ProfileExperiments.java
```diff
@@ -16,21 +16,23 @@
 package com.google.gapid.models;
 
 import com.google.gapid.proto.service.Service;
+import com.google.gapid.proto.service.path.Path;
+
+import java.util.List;
 
 public final class ProfileExperiments {
   public final boolean disableAnisotropicFiltering;
+  public final List<Path.Command> disabledCommands;
```
pmuetschard, line 25: Use an `ImmutableList<Path.Command>` here to maintain `ProfileExperiments`'s immutablility.

gapic/src/main/com/google/gapid/models/ProfileExperiments.java
```diff
@@ -16,21 +16,23 @@
...
 
-  public ProfileExperiments() {
-    this(false);
-  }
```
pmuetschard, line 25: Keep the default constructor to create an "empty" one, so you don't have to keep changing all the places where you create a default instance.

gapis/api/vulkan/vulkan.go
```diff
@@ -162,6 +158,7 @@ const (
 	DebugMarker = iota
 	RenderPassMarker
 	DrawGroupMarker
+	otherGroupMarker
```
pmuetschard, line 161: Why is this one private, while the others aren't. And what does 'other' mean? can you give it a better name?

yalcinmelihyasin: Here, `other` means  literally `other`. Any grouping that is not a draw, renderpass or debug. It is only used in one place which says:

```
// Handle an edge case where a group of state setting commands are
// followed by something other than a drawing command.
```

This used to be tagged as a draw group even though it does not draw. Therefore I wanted to separate it so we don't mistakenly tag a non-draw command a draw command.

If you have any other name suggestion, I am happy to accept.

yalcinmelihyasin: And actually all of them should be private. I will change them to private.

pmuetschard (commented):
The reason the tree labels don't update is because the styled strings are cached. You'd need to invalidate the cached renderings for every label that should be changed before calling refresh. Take a look at `LinkifedTree.getLabel()` and `LinkifiedTree.update()`.

If I select "isolate" on a draw call it will not disable siblings that have not been loaded.

gapic/src/main/com/google/gapid/widgets/LinkifiedTree.java
```diff
@@ -411,9 +420,25 @@ private boolean shouldIgnoreColors(Event event) {
       return (event.detail & SWT.SELECTED) != 0;
     }
 
+    public void updateHierarchy(TreeItem item) {
```
pmuetschard, line 423: This is way to expensive. You update the item and its children. Then you recurse with the parent item, which will again, update all the items that have already been updated. And this will happen again for every ancestor...
In fact, calling this function with any item will end up updating the entire tree, so why pass in an item at all?

gapic/src/main/com/google/gapid/widgets/LinkifiedTree.java
```diff
@@ -411,9 +420,25 @@ private boolean shouldIgnoreColors(Event event) {
...
+      update(item);
+      Arrays.asList(item.getItems()).forEach(child -> updateWithChildren(child));
+    }
+
     private void update(TreeItem item) {
-      Label label = getLabelNoUpdate(item);
       T element = getElement(item);
+      if (element == null) {
+        return;
+      }
```
pmuetschard, line 439: This seems fishy. Why do we now need this guard?

gapic/src/main/com/google/gapid/widgets/Theme.java
```diff
@@ -207,6 +207,8 @@
   @TextStyle(foreground = 0xa9a9a9) public Styler structureStyler();
   @TextStyle(foreground = 0x0000ee) public Styler identifierStyler();
   @TextStyle(bold = true) public Styler labelStyler();
+  @TextStyle(bold = true, strikeout = true) public Styler disabledlabelStyler();
+  @TextStyle(bold = true, foreground = 0xa9a9a9) public Styler semiDisabledlabelStyler();
```
pmuetschard, line 211: Can we do a dotted line strikeout instead? I don't like that we use two different ways to communicate a similar thing.

pmuetschard (commented):
mostly nits

gapic/src/main/com/google/gapid/widgets/LinkifiedTree.java
yalcinmelihyasin: The problem is we actually need to update the entire visible tree especially in cases of disabling/enabling a command with children. In our luck without secondary command buffer, this is a tree with the depth of 2 and with secondary command buffer it's with a depth of 4. 

Still, this is an expensive operation but the actual operation triggers a new replay that will take at least seconds so I don't think that this is a bottleneck here. In terms of UI, in my laptop I couldn't see any issues.

To be able to improve this, when we started a disable experiment, we need to calculate and return which commands are actually affected and only update their children & parents. Which is doable but I am not sure performance concern justifies the complexity. 

I would suggest waiting if anyone actually is affected by UI unresponsiveness from this and if they do then we can optimize but currently it's good enough.

gapic/src/main/com/google/gapid/widgets/Theme.java
yalcinmelihyasin: Sure.

gapic/src/main/com/google/gapid/views/CommandOptions.java
```diff
@@ -44,24 +47,76 @@ public static void CreateCommandOptionsMenu(Control parent, Widgets widgets, Tre
       }
     });
 
+    MenuItem disableMenuItem;
+    MenuItem enableMenuItem;
+    MenuItem isolateMenuItem;
     if (Experimental.enableProfileExperiments(models.settings)) {
-      MenuItem experimentMenuItem = Widgets.createMenuItem(optionsMenu, "Ex&periments", SWT.MOD1 + 'P', e -> {
+      disableMenuItem = Widgets.createMenuItem(optionsMenu, "Disable Command", SWT.MOD1 + 'D', e -> {
```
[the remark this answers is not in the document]
yalcinmelihyasin: Seems reasonable. The only thing I have changed all the others from command -> drawcall for consistency. We are not "only" disable drawcalls(e.g. secondary command buffers) but also "Disable Other Commands" may mean disable "all the other commands" so, there will be always some confusion but draw call seems better to me.

gapic/src/main/com/google/gapid/widgets/Theme.java
yalcinmelihyasin: Do we have a dotted line strikeout in our library? I couldn't find one.

yalcinmelihyasin: Do we have a dotted line strikeout in our library? I couldn't find one.

gapic/src/main/com/google/gapid/widgets/LinkifiedTree.java
yalcinmelihyasin: When we traverse the tree, we get some artificial nodes without data. It's to prevent that.

gapic/src/main/com/google/gapid/widgets/LinkifiedTree.java
pmuetschard: My point is that you update most items many times. Each time you recurse into a parent, you re-update the entire part of the tree you just updated. So, if you need to update the entire tree, fine, but only update it once.

pmuetschard (approved):

PR #705, opened 2021-03-09
Remove AHardwareBuffer from vkFrameBoundaryANDROID

hevrard:
We agreed with pmuetschard@ that we can do without it.

Bug: n/a


johnplate (approved):
LGTM. MacOS failure is a flake, please restart CI

pmuetschard (approved):

PR #706, opened 2021-03-09
Fix a few more invalid `protoutil.OneOf` usages.

pmuetschard:


stellama0208 (approved):
LGTM! The presubmit failure seems to ask for cleaning up import in the build file as well.

PR #709, opened 2021-03-10
Create separate gpu slice group for each critical renderpass slice.

stellama0208:
By observation, among the gpu slices returned by the pre-release driver,
sometimes several slices occur at different time space, but they share
the same renderPass handle and renderTarget handle. In this way, the key of
{submissionOrder, commandBuffer, renderPass, renderTarget} couldn't differentiate
them. Previously, AGI bundles these slices into a shared group and causes confusion.
With this PR, we make sure that each critical renderpass slice(e.g. "Surface"
slice on adreno device) has its own group.

Changes:
1. Streamline the related code a bit. First create a group if the current slice is a
critical slice, and then look up the slice's group.
2. No longer create a group for each {submissionOrder, commandBuffer, renderPass,
renderTarget} key, but for each critical gpu slice.
3. Change slice's name at the right time to fix a minor bug: sometimes a wrong index
shows up as a gpu slice's name in the Profile tab.

Bug: b/181140214.


hevrard (approved):
LGTM, I appreciated the nice commit message that sets up the context!

yalcinmelihyasin (approved):

PR #710, opened 2021-03-10
Use wget instead of curl on macOS CI

hevrard:
Curl started to fail very often recently, with errors like:

    curl -L -k -O -s https://dl.google.com/android/repository/android-ndk-r21d-darwin-x86_64.zip
    [ID: 2822626] Build finished after 37 secs, exit value: 16

So use wget instead, which is available on the Kokoro macOS images.

Bug: b/182379736


pmuetschard:
I have a more encompassing fix coming shortly

hevrard:
Superseded by #711

PR #714, opened 2021-03-12
CI-windows: make Bazel operate outside of C:

hevrard:
This can hopefully reduce flakes.

Bug: b/182387034

### DIFF
--- a/kokoro/windows/build.bat
+++ b/kokoro/windows/build.bat
@@ -104,14 +104,25 @@ if "%KOKORO_GITHUB_COMMIT%." == "." (
   set BUILD_SHA=%DEV_PREFIX%%KOKORO_GITHUB_COMMIT%
 )
 
-%BUILD_ROOT%\bazel build -c opt --config symbols ^
+REM Use T:\src for Bazel output, where files are less likely to be
+REM locked by system checks.
+set BAZEL_OUTPUT_USER_ROOT="T:\src"
+
+REM Build in several steps in order to avoid running out of memory.
+
+REM Build GAPIS api modules.
+%BUILD_ROOT%\bazel ^
+    --output_user_root=%BAZEL_OUTPUT_USER_ROOT% ^
+    build -c opt --config symbols ^
     --define AGI_BUILD_NUMBER="%KOKORO_BUILD_NUMBER%" ^
     --define AGI_BUILD_SHA="%BUILD_SHA%" ^
     //gapis/api/vulkan:go_default_library
 if %ERRORLEVEL% GEQ 1 exit /b %ERRORLEVEL%
 
 REM Build everything else.
-%BUILD_ROOT%\bazel build -c opt --config symbols ^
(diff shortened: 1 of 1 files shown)


hevrard:
Maybe it would be even better to just set HOME to BUILD_ROOT at the start, such that all commands see the same HOME (rather than backing up HOME and restoring it). Let me know your thoughts.

johnplate:
> Maybe it would be even better to just set HOME to BUILD_ROOT at the start, such that all commands see the same HOME (rather than backing up HOME and restoring it). Let me know your thoughts.

I would not change HOME at all, might have bad side effects. On my Windows machine I use a .bazelrc file in my user directory, only containing the line `startup --output_user_root=D:/scratch/bazel`, which works fine. It would be even easier to add this option to both bazel commands instead of creating that file.

pmuetschard (commented):
1) I'm not convinced that this is the root cause of the flakes. So, I would remove that comment.
2) I agree with John, don't touch `%HOME%`, but pass the output root flag to each bazel invocation, instead.

hevrard:
Thanks both for the comments, I've updated the change accordingly, PTAL

johnplate (approved):
LGTM, but wait for the Windows CI to succeed and check if bazel actually uses the given directory.

johnplate:
> LGTM, but wait for the Windows CI to succeed and check if bazel actually uses the given directory.

I checked it already, works fine, ready to submit.

pmuetschard:
make sure you rebase on top of master and also add the flag to the smoketest invocation.

johnplate:
> make sure you rebase on top of master and also add the flag to the smoketest invocation.

I'm confused, why would the smoketest invocation need the flag?

hevrard:
@johnplate Pascal recently refactored the way smoketests is run, so after rebasing, this also needs the flag.

johnplate:
> @johnplate Pascal recently refactored the way smoketests is run, so after rebasing, this also needs the flag.

Thanks, got it. New push still LGTM

hevrard:
Rebased on master to have CI-windows download errors fixed.

kokoro/windows/build.bat
```diff
@@ -104,22 +104,36 @@ if "%KOKORO_GITHUB_COMMIT%." == "." (
   set BUILD_SHA=%DEV_PREFIX%%KOKORO_GITHUB_COMMIT%
 )
 
-%BUILD_ROOT%\bazel build -c opt --config symbols ^
+REM Use T:\src for Bazel output, where files are less likely to be
+REM locked by system checks.
+set BAZEL_OUTPUT_USER_ROOT="T:\src\bazel_user_root"
```
pmuetschard, line 109: Use `%BUILD_ROOT%` and name it something short. `out` is already used so why not `%BUILD_ROOT%\build`?

hevrard: done

pmuetschard (approved):

PR #716, opened 2021-03-12
Fix Windows CI SHA check

johnplate:
The SHA256 check in the Windows CI failed always as the Windows echo command printed quotation marks and CRLF newline, which both can't be handled by sha256sum. The failure was not noticed because it didn't terminate the batch.

This PR fixes the echo command and adds a termination on failure.


hevrard (approved):

hevrard:
thanks for this!

PR #722, opened 2021-03-15
Show static analysis counters in Shader View

hliatis:
Note: There's some repeated code with Pipeline View.
It'll be handled in a future PR.

Bug: b/180452518

### DIFF
--- a/gapic/src/main/com/google/gapid/views/ShaderView.java
+++ b/gapic/src/main/com/google/gapid/views/ShaderView.java
@@ -20,9 +20,15 @@ import static com.google.gapid.util.Loadable.MessageType.Info;
 import static com.google.gapid.widgets.Widgets.createBoldLabel;
 import static com.google.gapid.widgets.Widgets.createComposite;
 import static com.google.gapid.widgets.Widgets.createGroup;
+import static com.google.gapid.widgets.Widgets.createLabel;
+import static com.google.gapid.widgets.Widgets.createLink;
+import static com.google.gapid.widgets.Widgets.createScrolledComposite;
 import static com.google.gapid.widgets.Widgets.createStandardTabFolder;
 import static com.google.gapid.widgets.Widgets.createStandardTabItem;
+import static com.google.gapid.widgets.Widgets.disposeAllChildren;
+import static com.google.gapid.widgets.Widgets.withLayoutData;
 
+import com.google.common.base.Joiner;
 import com.google.gapid.lang.glsl.GlslSourceConfiguration;
 import com.google.gapid.models.Analytics.View;
 import com.google.gapid.models.Capture;
@@ -31,6 +37,7 @@ import com.google.gapid.models.Resources;
 import com.google.gapid.proto.service.Service;
 import com.google.gapid.proto.service.Service.ClientAction;
 import com.google.gapid.proto.service.api.API;
+import com.google.gapid.proto.service.path.Path;
 import com.google.gapid.rpc.Rpc;
 import com.google.gapid.rpc.RpcException;
 import com.google.gapid.rpc.SingleInFlight;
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/views/ShaderView.java
```diff
@@ -148,6 +161,10 @@ public ShaderView(Composite parent, Models models, Widgets widgets) {
     pushButton.setLayoutData(new GridData(SWT.RIGHT, SWT.BOTTOM, false, false));
     pushButton.setEnabled(false);
 
+    statGroup = Widgets.createGroup(loading.getContents(), "TEST");
```
pmuetschard, line 164: TEST?

gapis/api/service.proto
```diff
@@ -145,6 +145,7 @@ message Shader {
   string source_language = 3;
   string spirv_source = 4;
   bool cross_compiled = 5;
+  DataGroup static_analysis = 6;
```
pmuetschard, line 148: I think we've reached the point, where I'd prefer we'd define an explicit proto for this, rather than stick with the server generated key-value map approach. In this PR you're adding code to format all kinds of values, even though all we're currently (and probably ever) are going to get here are uint32s.

gapic/src/main/com/google/gapid/views/ShaderView.java
hliatis: Whoops!

gapis/api/service.proto
hliatis: I went ahead and updated it to a more explicit implementation. However, I am starting to think that the better approach would be to have an array of ints and an array of corresponding labels for said ints. Thoughts?

gapic/src/main/com/google/gapid/views/ShaderView.java
```diff
@@ -287,4 +312,74 @@ protected void onUiThread(API.Shader result) {
   private static boolean isKey(Event e, int stateMask, int keyCode) {
     return (e.stateMask & stateMask) == stateMask && e.keyCode == keyCode;
   }
+
+  private void CreateStaticAnalysisStat(Composite parent, String statText, int statData)
+  {
```
pmuetschard, line 317: new new-lines before braces (also below)

gapic/src/main/com/google/gapid/views/ShaderView.java
```diff
@@ -287,4 +312,74 @@ protected void onUiThread(API.Shader result) {
...
+    GridLayout statsLayout = new GridLayout(1, false);
+    statsLayout.marginHeight = 5;
+    statsLayout.marginWidth = 5;
+
+    Composite keyComposite = withLayoutData( createComposite(parent, statsLayout, SWT.BORDER),
```
pmuetschard, line 322: no space after open parens (a bunch of times here)

gapic/src/main/com/google/gapid/views/ShaderView.java
```diff
@@ -287,4 +312,74 @@ protected void onUiThread(API.Shader result) {
   private static boolean isKey(Event e, int stateMask, int keyCode) {
     return (e.stateMask & stateMask) == stateMask && e.keyCode == keyCode;
   }
+
+  private void CreateStaticAnalysisStat(Composite parent, String statText, int statData)
```
pmuetschard, line 316: lowercase function name

gapis/api/service.proto
```diff
@@ -145,6 +145,14 @@ message Shader {
   string source_language = 3;
   string spirv_source = 4;
   bool cross_compiled = 5;
+  StaticAnalysis static_analysis = 6;
+}
+
+message StaticAnalysis {
```
pmuetschard, line 151: Either put it insider the `Shader` message, or call it `ShaderStaticAnalysis`.

pmuetschard (approved):

PR #724, opened 2021-03-16
Ensure the port message always is on its own line.

pmuetschard:
Don't assume that `fmt.Printf` starts on a new line, in case it gets invoked in-between a log message printing its message and its new-line.

Bug: http://b/144158856

### DIFF
--- a/gapis/server/grpc.go
+++ b/gapis/server/grpc.go
@@ -66,7 +66,7 @@ func NewWithListener(ctx context.Context, l net.Listener, cfg Config, srvChan ch
 		done <- grpcutil.ServeWithListener(ctx, l, func(ctx context.Context, listener net.Listener, server *grpc.Server) error {
 			if addr, ok := listener.Addr().(*net.TCPAddr); ok {
 				// The following message is parsed by launchers to detect the selected port. DO NOT CHANGE!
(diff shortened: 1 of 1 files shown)


hevrard (approved):

gapis/server/grpc.go
```diff
@@ -66,7 +66,7 @@ func NewWithListener(ctx context.Context, l net.Listener, cfg Config, srvChan ch
 		done <- grpcutil.ServeWithListener(ctx, l, func(ctx context.Context, listener net.Listener, server *grpc.Server) error {
 			if addr, ok := listener.Addr().(*net.TCPAddr); ok {
 				// The following message is parsed by launchers to detect the selected port. DO NOT CHANGE!
-				fmt.Printf("Bound on port '%d'\n", addr.Port)
+				fmt.Printf("\nBound on port '%d'\n", addr.Port)
```
hevrard, line 69: I would add a note about the newline in the above comment, e.g.:

```
// The following message is parsed by launchers to detect the selected port. The newline at the start is here to make sure it stands on its own line. DO NOT CHANGE!
```

hevrard:
This potentially fixes b/144158856

PR #728, opened 2021-03-17
CI: Update to gcc-10

hevrard:
The CI started to fail to install gcc-8, so bumping to gcc-10.

Bug: b/182927077


hevrard:
Re-open to experiment with the CI build on Swarming bots.

PR #729, opened 2021-03-17
Update GCC to version 9

yalcinmelihyasin:


johnplate:
Not required anymore, the issue with gcc-8 has been resolved.

PR #730, opened 2021-03-17
Report render target's size for each renderpass slice in Profile tab.

stellama0208:
This was already in System Profiler and we got request from users to
have the same thing in Frame Profiler. The logic of appending the size
to a renderpass slice's name is already shared by Frame Profiler and
System Profiler, but the missing part is that in Frame Profiler we
didn't have argument set information in place, and thus we missed the
width/height information. This PR fills this gap.

Bug: b/169454084.
Test: Observe slices' name in Frame Profiler - Profile tab. Test on an
adreno phone since currently we only have width/height info in argument
set from these devices.


hevrard (approved):
Yay!

PR #739, opened 2021-03-22
Improve error message for app crash during capture

hevrard:
When the app being captured crashes, the error message presented to
the end user only said "EOF". This change indicate that such errors
are likely to be app early termination / app crash.

Bug: b/179234685
Test: Manual

### DIFF
--- a/gapii/client/capture.go
+++ b/gapii/client/capture.go
@@ -144,7 +144,7 @@ func handleCommError(ctx context.Context, commErr error, anyDataReceived bool) (
 		abort = true
 		// Most of the time, this error happens when the app crashed: rather
 		// than reporting just "EOF", hint that this was probably a crash.
-		err = log.Err(ctx, commErr, "It looks like the app terminated (crashed?) before we could finish the capture")
+		err = log.Err(ctx, commErr, "The application exited during the capture")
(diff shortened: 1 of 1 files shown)


hevrard:
See before/after screenshots attached.
before
after

gapii/client/capture.go
```diff
@@ -142,7 +142,9 @@ func handleCommError(ctx context.Context, commErr error, anyDataReceived bool) (
 	case errors.Cause(commErr) == io.EOF:
 		log.E(ctx, "unexpected end of stream")
 		abort = true
-		err = commErr
+		// Most of the time, this error happens when the app crashed: rather
+		// than reporting just "EOF", hint that this was probably a crash.
+		err = log.Err(ctx, commErr, "It looks like the app terminated (crashed?) before we could finish the capture")
```
pmuetschard, line 147: Remove the "it looks like" qualifier. I would also prefer a more neutral tone. Maybe "The application exited during tracing"

pmuetschard (approved):

gapii/client/capture.go
hevrard: OK, I went with `The application exited during the capture` as most user-facing text seems to use "capture" rather than "trace".

PR #741, opened 2021-03-22
Allow logging of initial commands to a capture file.

pmuetschard:


gapis/api/vulkan/replay.go
```diff
@@ -42,17 +42,22 @@ var (
...
-	initialCmds, im, _ := initialcmds.InitialCommands(ctx, capture)
+	initialCmds, im, _ := initialcmds.InitialCommands(ctx, c)
 	out.State().Allocator.ReserveRanges(im)
 	cmdGenerator := commandGenerator.NewLinearCommandGenerator(initialCmds, nil)
 
 	transforms := make([]transform.Transform, 0)
 	transforms = append(transforms, newMakeAttachmentReadable(false))
 	transforms = append(transforms, newDropInvalidDestroy("GetInitialPayload"))
+	if config.LogInitialCmdsToCapture {
+		if c, err := capture.ResolveGraphicsFromPath(ctx, c); err == nil {
```
hevrard, line 57: Style: `c := ...` shadows the function argument. I think we should avoid shadowing as much as feasible, I'll let you judge.

hevrard (approved):

gapis/api/vulkan/replay.go
pmuetschard: I think it's OK here, because 1) the new `c` only lives within the if and 2) it's still "the same thing" - capture path vs capture object.

PR #744, opened 2021-03-23
Support downloading Perfetto trace data during capture.

pmuetschard:
When using the direct client, if we're on R+ and not tracing into a file, download the trace data during capture via the ReadBuffers RPC.

### DIFF
--- a/cmd/perfetto/trace.go
+++ b/cmd/perfetto/trace.go
@@ -20,26 +20,31 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"time"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/google/gapid/core/app"
 	"github.com/google/gapid/core/app/crash"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/os/device/bind"
+	"github.com/google/gapid/gapis/perfetto"
 
 	config "protos/perfetto/config"
 )
 
 type traceVerb struct {
-	Config string `help:"File containing the trace configuration proto."`
-	Out    string `help:"The file to store the trace data in."`
+	Config string        `help:"File containing the trace configuration proto."`
+	Out    string        `help:"The file to store the trace data in."`
+	Read   time.Duration `help:"The time to wait in-between read requests."`
 }
 
 func init() {
 	verb := &traceVerb{
-		Out: "trace.perfetto",
+		Out:  "trace.perfetto",
+		Read: 100 * time.Millisecond,
 	}
 	app.AddVerb(&app.Verb{
 		Name:      "trace",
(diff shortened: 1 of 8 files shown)


core/os/android/adb/commands.go
```diff
@@ -349,10 +349,13 @@ func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.Perfet
...
 	result.GpuProfiling = gpu
 
-	// SurfaceFlinger frame lifecycle perfetto producer is mandated by Android 11 CTS, hence it will
-	// always exist.
 	if b.Instance().GetConfiguration().GetOS().GetAPIVersion() >= 30 {
+		// SurfaceFlinger frame lifecycle perfetto producer is mandated by Android 11 CTS, hence it will
+		// always exist.
 		result.HasFrameLifecycle = true
+
+		// This has anecdotally not worked well in Q, but appears to be fine in R.
```
hevrard, line 357: I think we should prefer `Android11` to Q and `Android12` to R in general, and here the comment above already uses "Android 11". I would be keen on having "Android" and its version number in one word, to help grepping for it in the future.

gapis/perfetto/android/trace.go
```diff
@@ -301,10 +319,30 @@ func (p *Process) captureWithClientApi(ctx context.Context, start task.Signal, s
...
 	ts.Start(ctx)
+
 	wait := make(chan error, 1)
 	crash.Go(func() {
 		wait <- ts.Wait(ctx)
 	})
+
+	if downloadWhileTracing {
+		crash.Go(func() {
+			ticker := time.NewTicker(100 * time.Millisecond)
```
hevrard, line 330: Maybe put this duration in a package named variable?

gapic/src/main/com/google/gapid/settings.proto
```diff
@@ -174,6 +174,8 @@ message Perfetto {
   }
   Vulkan vulkan = 5;
 
+  bool force_tracing_to_file = 8;
```
hevrard, line 177: Put it at the end of the message such that we keep field IDs in order.

hevrard (approved):

cmd/perfetto/trace.go
```diff
@@ -20,26 +20,31 @@ import (
...
 
 	config "protos/perfetto/config"
 )
 
 type traceVerb struct {
-	Config string `help:"File containing the trace configuration proto."`
-	Out    string `help:"The file to store the trace data in."`
+	Config string        `help:"File containing the trace configuration proto."`
+	Out    string        `help:"The file to store the trace data in."`
+	Read   time.Duration `help:"The time to wait in-between read requests."`
```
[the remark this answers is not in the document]
pmuetschard: I think the help text is quite clear. Flags should be short.

core/os/android/adb/commands.go
pmuetschard: You pry the letter names from my cold dead fingers.

PR #746, opened 2021-03-24
Fix an AGI freezing and OOM problem when loading block memory.

stellama0208:
This bug may happen when AGI has Memory tab checked, but not put at front.

When block memory UI tries to repaint, it will 1.check repainting area, 2.
calculate repainting memory range accordingly, 3.query the server to load the
memory range.
In AGI, we have been using gc.getClipping() to do the step #1 for quite some
time, but recently we find that it will not do #1 correctly when the Memory tab
is hidden in the back, and thus the following steps will get wrong and cause
incorrect server query. (It's possible but not sure, that SWT handles gc.getClipping()
differently in newer version and thus this bug surfaces recently.) This PR tries
to do the step #1 based on another data source and thus solve the problem.

Test: This bug may happen when AGI has Memory tab checked, but not put at
front. Reproduced on MacOS BigSur.
Bug: b/181012589.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/MemoryView.java
+++ b/gapic/src/main/com/google/gapid/views/MemoryView.java
@@ -676,17 +676,16 @@ public class MemoryView extends Composite
     }
 
     @Override
-    public void paint(BigInteger xOffset, BigInteger yOffset, GC gc) {
+    public void paint(BigInteger xOffset, BigInteger yOffset, GC gc, Rectangle area) {
       if (model == null) {
         return;
       }
 
       gc.setFont(font);
-      Rectangle clip = gc.getClipping();
-      BigInteger startY = yOffset.add(BigInteger.valueOf(top(clip)));
+      BigInteger startY = yOffset.add(BigInteger.valueOf(top(area)));
       long startRow = startY.divide(lineHeightBig)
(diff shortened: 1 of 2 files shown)


hevrard (approved):
Thank you very much for fixing this!

yalcinmelihyasin:
Thanks for solving this!

yalcinmelihyasin (approved):

pmuetschard (commented):
Why not simply add a `if (startRow >= endRow) { return }` check?

stellama0208:
> Why not simply add a `if (startRow >= endRow) { return }` check?

An early return by checking critical variables could work, and actually I was also thinking whether I should add one of the checks into this PR to make things double safe. 

There's one disadvantage of early return though, that based on some of my tests yesterday, I found a single early return will stop repainting, and thus when user switch back to Memory tab, they have to manually trigger another repaint event by clicking, so as not to see a blank area or outdated info.

Among some solutions:  1. different redraw area data source, 2.early return by checking variable, 3. combination of #1 and #2. 
This PR currently follows #1. Hey Pascal @pmuetschard, I'm wondering whether you'd prefer another way that we shape this PR? Welcome for advice!

pmuetschard (approved):
OK, thanks Stella. `Event.getBounds()` for paint events returns the clipping area, so this looks good.

stellama0208:
Thanks for confirmation!

PR #752, opened 2021-03-31
Fix display of representations for OpenGL draw commands/groups

### LINKED ISSUE #669: Frame capture: 1-draw-call difference in images between glDrawElements and vkCmdDrawIndexed
Note: This is described w.r.t. FIFA Soccer, but is true for all GLES games/applications.

Describe the issue:

- This may be an issue with other games, but I just noticed it.
- Looking at my frame capture of FIFA Soccer, I sometimes click on the a higher-level glDrawElements node, and sometimes I click on the GLES and/or Vulkan nodes underneath. For the same higher-level glDrawElements, I get different images for the GLES and Vulkan nodes.

Expected behavior:

- When I open and select different nodes of a frame capture of a GLES app, I expect the same image to be seen regardless of whether I am clicking on the GLES or Vulkan nodes

Actual behavior:

- I am seeing one-draw-call difference between the "OpenGL ES Commands" and the "DrawIndexed" hierarchies. Since a small thumbnail is included on some of the hierarchy, I started looking at the draws before and after big changes. That's when I noticed this.

Repro steps:

- With the capture that I point to (below), look at when the draws in the 2nd RenderPass change from black to green. Here's what the hierarchy looks like:
  - Host Coordination
  - vkQueueSubmit
    - pSubmits[0]
      - Command Buffer
        - Render Pass
        - Render Pass
          - glDrawElements
          - glDrawElements
          - glDrawElements
          - glDrawElements
          - glDrawElements
          - glDrawElements
          - glDrawElements
            - vkCmdBeginDebugUtilsLabelEXT
              - OpenGL ES Commands
                - vkCmdBeginDebugUtilsLabelEXT
                - gl*
                - ...
                - glDrawElements
                - vkCmdEndDebugUtilsLabelEXT
              - DrawIndexed
                - vkCmdBindPipeline
                - ...
                - vkCmdDrawIndexed
              - vkCmdEndDebugUtilsLabelEXT
            - glDrawElements
              - vkCmdBeginDebugUtilsLabelEXT
              - OpenGL ES Commands
                - vkCmdBeginDebugUtilsLabelEXT
                - gl*
                - ...
                - glDrawElements
                - vkCmdEndDebugUtilsLabelEXT
              - DrawIndexed
                - vkCmdBindPipeline
                - ...
                - vkCmdDrawIndexed
              - vkCmdEndDebugUtilsLabelEXT
            - glDrawElements
            - ...

- I chose those particular parts of the hierarchy because that's where the green soccer field first gets drawn.
- If you open up the "OpenGL ES Commands" and the "DrawIndexed" nodes above, and click on the lower-level glDrawElements and vkCmdDrawIndexed nodes. You will see that the GLES draw doesn't show the same image as the Vulkan draw.

Versions:

- AGI: v1.1.0-dev-20210110
- ANGLE (pre-built): <ANGLE version doesn't matter>
- App version: versionCode=754 minSdk=23 targetSdk=29

mark-lunarg:
In an ANGLE AGI trace, the thumbnail/resources of the containing OpenGL group and the OpenGL drawcall displayed out-of-date
data. Updated these representations at query-time to reflect the correct Vulkan state.

Passed presubmit tests.

Fixes #669

### DIFF
--- a/gapis/resolve/command_tree.go
+++ b/gapis/resolve/command_tree.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/math/interval"
@@ -140,6 +141,12 @@ func CommandTreeNode(ctx context.Context, c *path.CommandTreeNode, r *path.Resol
 		startID := append(absID, uint64(item.Range.First()))
 		endID := append(absID, uint64(item.Range.Last()))
 		representation = cmdTree.path.Capture.Command(endID[0], endID[1:]...)
+
+		if aliasRepId := getOpenGLAliasRepresentation(c.Indices, &item, cmdTree); aliasRepId != api.CmdNoID {
+			aliasId := append(absID, uint64(aliasRepId))
+			representation = cmdTree.path.Capture.Command(aliasId[0], aliasId[1:]...)
+		}
+
 		return &service.CommandTreeNode{
 			Representation: representation,
 			NumChildren:    item.Count(),
@@ -472,6 +479,7 @@ func (f frame) addGroup(t *commandTree) {
 func addFrameGroups(ctx context.Context, events *service.Events, p *path.CommandTree, t *commandTree, last api.CmdID) {
 	frameCount := 0
 	firstFrame, curFrame := frame{}, frame{}
+
 	for _, e := range events.List {
 		i := api.CmdID(e.Command.Indices[0])
 		switch e.Kind {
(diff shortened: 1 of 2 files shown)


mark-lunarg:
@googlebot I signed it!

gapis/resolve/command_tree.go
```diff
@@ -236,8 +243,11 @@ func (r *CommandTreeResolvable) Resolve(ctx context.Context) (interface{}, error
 			return fmt.Errorf("Fail to mutate command %v: %v", cmd, err)
 		}
 		if filter(id, cmd, s, api.SubCmdIdx([]uint64{uint64(id)})) {
+			var groupnum int
```
pmuetschard, line 246: `groupnum` doesn't appear to be used anywhere.

gapis/resolve/command_tree.go
```diff
@@ -546,3 +557,28 @@ func setRepresentations(ctx context.Context, g *api.CmdIDGroup, drawOrClearCmds
...
+	aliasRepId := api.CmdNoID
+	if item.Name == "OpenGL ES Commands" {
+		parentIndices := indices[:len(indices)-1]
+		pItem, _ := cmdTree.index(parentIndices)
+		parentItem := pItem.(api.CmdIDGroup)
+		aliasRepId = parentItem.Range.Last()
+	} else if strings.HasPrefix(item.Name, "glDraw") || strings.HasPrefix(item.Name, "glMultiDraw") {
+		pindices := indices[:len(indices)-1]
+		pItem, _ := cmdTree.index(pindices)
+		switch parentItem := pItem.(type) {
```
pmuetschard, line 571: Since you only have one case statement, I think it's better to do:
```
if parentItem, ok := pItem.(api.CmdIDGroup); ok {
...
}
```
Same below.

gapis/resolve/command_tree.go
mark-lunarg: Leftover instrumentation, sorry.

gapis/resolve/command_tree.go
mark-lunarg: Thanks -- was looking for something like this!

pmuetschard (approved):

PR #754, opened 2021-04-01
183624452 swarming adb path

hevrard:


hevrard:
There are two commits, the first one gets rid of an irrelevant test script, the second contains meaningful change. You should first look at the change in botutil.py that defines the BotUtil class and moves various methods under it, and then these methods are used in other scripts.

johnplate (approved):

PR #759, opened 2021-04-09
Refactor handling of enabled dynamic states

hevrard:
Various dynamic states may be enabled when a graphics pipeline is
created. These states are defined as:

```
typedef enum VkDynamicState {
    VK_DYNAMIC_STATE_VIEWPORT = 0,
    VK_DYNAMIC_STATE_SCISSOR = 1,
    ... etc ...
    VK_DYNAMIC_STATE_STENCIL_REFERENCE = 8,
    // remaining states are defined by extensions, e.g.:
    // Provided by VK_EXT_line_rasterization
    VK_DYNAMIC_STATE_LINE_STIPPLE_EXT = 1000259000,
    ... etc ...
}
```

So far, we only support the core states, and the code assumes there are nine
possible states from 0 to 8, as we see below.

In practice, upon creation of the graphics pipeline, the list of dynamic state
passed as argument is stored in a map(u32,VkDynamicState):

```
@internal class DynamicData {
    @unused map!(u32, VkDynamicState) DynamicStates
}
```

Upon each draw call, we need to know whether some dynamic states are enabled or
not. To answer this, we could iterate over DynamicStates, but to avoid multiple
iterations over this map, we create a short-lived object with an array of 9
booleans indexed by the dynamic state value, so we can quickly lookup if a
given dynamic state is present or not. This is in draw_commands.api:

```
@internal class DynamicStateSet {
  bool[9] contains
}

sub  ref!DynamicStateSet unpackDynamicState(ref!DynamicData data) {
  obj := new!DynamicStateSet()
  if data != null {
    for _, _, s in data.DynamicStates {
      if as!s32(s) < 9 {
        obj.contains[as!s32(s)] = true
      }
    }
  }
  return obj
}

// This is called at every vkCmdDraw*
sub void readDrawState() {
  ...
  dynamicState := unpackDynamicState(pipeline.DynamicState)
  ...
  // Used like this to check for dynamic state
  if dynamicState.contains[as!u32(VK_DYNAMIC_STATE_VIEWPORT)] {
    ...
  }
  ...
}
```

Note that this approach using an array assumes there are only 9 dynamic states,
whose values range from 0 to 8. Note also that this creates a new, short-lived
DynamicStateSet at every draw call, which probably results in object churning
when we mutate traces with numerous draw calls.

In order to be able to handle dynamic state values from extensions (i.e., that
are not continuous after value 8, but that jumps to arbitrary values like
VK_DYNAMIC_STATE_LINE_STIPPLE_EXT = 1000259000), this change replaces the array
of booleans by a map(VkDynamicState, bool), namely the new Contains map in
DynamicData.

In order to avoid creating a short-lived object at every draw call, the new
Contains is stored in DynamicData and is set once for all upon pipeline
creation. Indeed Vulkan doesn't offer any mean to modify the set of enabled
dynamic state after pipeline creation.

We still need to keep the map(u32, VkDynamicState) DynamicStates since the
state rebuilder expects such a map to know how to re-create a GraphicsPipeline
with the same list of arguments, in the same order, for the set of dynamic
states to enable.

Bug: b/184934061
Test: Manual

### DIFF
--- a/gapis/api/vulkan/api/draw_commands.api
+++ b/gapis/api/vulkan/api/draw_commands.api
@@ -482,25 +482,13 @@ sub void readWriteMemoryInBoundComputeDescriptorSets() {
     lci.BufferBindingOffsets)
 }
 
-@internal class DynamicStateSet {
-  bool[9] contains
-}
-
-sub  ref!DynamicStateSet unpackDynamicState(ref!DynamicData data) {
-  obj := new!DynamicStateSet()
-  if data != null {
-    for _, _, s in data.DynamicStates {
-      if as!s32(s) < 9 {
-        obj.contains[as!s32(s)] = true
-      }
-    }
-  }
-  return obj
-}
-
 @spy_disabled
 sub void readComputeState() {
   readPushConstants(VK_PIPELINE_BIND_POINT_COMPUTE)
+}
+
+sub bool hasDynamicState(ref!GraphicsPipelineObject pipeline, VkDynamicState dynamicState) {
+  return (pipeline.DynamicState != null) && pipeline.DynamicState.Contains[dynamicState]
 }
 
 @spy_disabled
@@ -508,11 +496,10 @@ sub void readDrawState() {
   info := lastDrawInfo()
   pipeline := info.GraphicsPipeline
 
-  dynamicState := unpackDynamicState(pipeline.DynamicState)
   dyn := lastDynamicPipelineState()
 
   // read all viewports
-  if dynamicState.contains[as!u32(VK_DYNAMIC_STATE_VIEWPORT)] {
+  if hasDynamicState(pipeline, VK_DYNAMIC_STATE_VIEWPORT) {
     _ = len(dyn.Viewports)
   } else {
     if pipeline.ViewportState != null {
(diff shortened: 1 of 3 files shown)


hliatis (commented):
This fix regresses the pipeline view. In the sacha willems bloom demo, the dynamically set viewports and scissors are no longer recognized.

Edit: It works for new traces.

purvisa-at-google-com:
Looks good to me. If performance is a concern you could create a custom object that contains an array of 9 bools and a map. Use the bools for the "built in" enums and the map for the extension values. This would avoid the map overhead on common values while still allowing extensibility and a sparse memory footprint.

Consider this comment optional though, I'm happy to submit as-is!

purvisa-at-google-com (approved):

hliatis (approved):

hevrard:
Bumping the trace version breaks the smoketests... I can take a look at this next week.

hevrard:
I decided to not increment the trace version, because:
- this trace version is, I think, meant to be incremented every time we break backward compatibility of traces, but in practice it is never incremented (current value is 3, despite the project being years old with dozens of changes breaking trace backward compatibility)
- Incrementing it breaks smoketests, and would also break various swarming tests, as both types of tests rely on pre-captured traces
- The team has an upcoming series of meetings where I will raise this point to clarify how we should treat trace version and backward compatibility, and potentially refactor the tests accordingly

PR #763, opened 2021-04-13
Debug-log all shell commands

hevrard:
This helps a lot when debugging AGI on certain devices / special
Android builds.

Bug: n/a
Test: manual


paulthomson:
Thanks! Looks like the shell tests now fail though. Also, I guess this will affect all process executions, not just adb, but I guess that is not a bad thing.

paulthomson:
Also, one of the key issues is: I would like to see the stdout/stderr when `adb install` fails, so I know why it fails. Seeing the exact command arguments is also useful, but probably not enough.

PR #766, opened 2021-04-15
Add gapis --android-serial flag to limit device scanning

hevrard:
This flags let the user tell GAPIS to not scan devices other than a
specific one, and thus avoid haing GAPIS install its APK (as part of
device scanning) to all the connected devices.

Bug: b/185349557
Test: core/os/android/adb:LimitToSerialTest, manual


pau-baiget (approved):
Good job! And nice added test :)

PR #767, opened 2021-04-16
Update Perfetto for Frame Lifecycle

silence-do-good:
In some devices, same buffer gets reused across multiple layers. This
breaks some internal assumption. The trace processor fix landed in
perfetto. This change updates perfetto to include the fix.

Since layer name is now displayed along with the buffer id, use right
truncate for the buffer track names.

Bug: 184037965


pmuetschard (commented):
Please fix the presubmit error.

hevrard (approved):
Thanks for this!

PR #768, opened 2021-04-16
Migrate from python2 to python3

johnplate:
This PR migrates AGI from python2 to python3 and updates the build instructions for Windows.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -46,18 +46,18 @@ Start a console, with administrator privilege, and type:
 
 In the same console, install Python 2.7 and MSYS2 as well:
 
-`choco install python2`
+`choco install python`  
 `choco install msys2`
 
 ### Install additional tools
 
-Using the msys64 shell (installed by choco at `C:\tools\msys64\mingw64`):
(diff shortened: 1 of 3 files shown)


BUILDING.md
```diff
@@ -44,21 +44,21 @@ Start a console, with administrator privilege, and type:
 
 `choco install bazel --version 2.0.0`
 
-In the same console, install Python 2.7 and MSYS2 as well:
+In the same console, install Python and MSYS2 as well:
 
-`choco install python2`
+`choco install python`
```
hevrard, line 49: Would it be better to be specific and require python3 here?

hevrard (approved):

BUILDING.md
johnplate: `python` is a specific [chocolatey package](https://community.chocolatey.org/packages/python/3.9.4) that I have successfully tested. `python3` is a different [package](https://community.chocolatey.org/packages/python3/3.9.4) that should be tested with building AGI before we reference it, so for now I would prefer to keep `python`.

PR #770, opened 2021-04-20
Add trace option: name of the process to capture

hevrard:
Some games may start with a first process (or Android activity), and
then fork another process/activity where the actual game rendering
happens. If the starting process also uses Vulkan, then it is likely
to be the first to establish a gapii-gapis connection, such that AGI
won't be able to capture the actual game.

This change adds a "process name" trace option to indicate the name of
the process to capture. This is a new text entry in the UI's
TracerDialog. The value gets propagated as a trace option through
gapis, and is eventually passed to gapii: via an Android system
property on Android, or via an environment variable (refactoring the
existing `GAPID_CAPTURE_PROCESS_NAME` mechanism) on Desktop. The gapii
Spy, upon creation, looks for this process name and only establish a
connection with gapis if the current process name matches the one to
capture, or if the capture process name is empty.

This was motivated by a Unity game which forks a new process whose
name is `com.example.game:UnityKillsMe`, where the game is actually
rendered. This seems to be a way to include Unity games into an
Android app, the `UnityKillsMe` suffix lets Unity take care of killing
the forked activity upon app shutdown.

Bug: b/177363287
Test: manual

### DIFF
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -50,6 +50,7 @@
 
 #include <jni.h>
 #include <sys/prctl.h>
+#include <sys/system_properties.h>
 
 extern "C" jint JNI_OnLoad(JavaVM* vm, void* reserved) {
   GAPID_INFO("JNI_OnLoad() was called. vm = %p", vm);
@@ -64,6 +65,16 @@ const uint32_t kMaxFramebufferObservationWidth = 3840;
 const uint32_t kMaxFramebufferObservationHeight = 2560;
 
 const int32_t kSuspendIndefinitely = -1;
+
+#if TARGET_OS == GAPID_OS_ANDROID
+// Android: system property holding the name of the process to capture.
+// Mirrored in gapii/client/adb.go
+const char* kCaptureProcessNameSystemProperty = "debug.agi.procname";
+#else
+// Desktop: environment variable holding the name of the process to capture.
+// Mirrored in gapis/trace/desktop/trace.go
+const char* kCaptureProcessNameEnvVar = "GAPID_CAPTURE_PROCESS_NAME";
+#endif  // TARGET_OS == GAPID_OS_ANDROID
 
 thread_local gapii::CallObserver* gContext = nullptr;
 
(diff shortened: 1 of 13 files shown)


hevrard:
@johnplate we discussed using a file inside `/sdcard/Android/data/...` instead of the system property, but our code already has primitives to setup system properties, together with cleanup functions, so to keep this change small I reused those. We might want to consider using files in `/sdcard/...` in a future refactoring that would encompass replacing the various system properties we touch.

gapii/cc/spy.cpp
```diff
@@ -97,12 +108,34 @@ Spy::Spy()
...
+  __system_property_get(kCaptureProcessNameSystemProperty, capture_proc_name);
+#else
+  const char* capture_proc_name = getenv(kCaptureProcessNameEnvVar);
+  // Make sure capture_proc_name is not null, since the code below assumes so.
+  if (!capture_proc_name) {
+    capture_proc_name = "";
   }
+#endif
+
+  if (strlen(capture_proc_name)) {
```
johnplate, line 132: No need to count the length, just use `if (capture_proc_name != nullptr && *capture_proc_name != nullptr) {` to check if the string is valid and non-empty. Then you can also remove the 4 lines after the getenv command.

gapii/cc/spy.cpp
```diff
@@ -97,12 +108,34 @@ Spy::Spy()
...
+    this_executable = (!strcmp(this_proc_name.c_str(), capture_proc_name));
```
johnplate, line 133: `this_proc_name` is a std::string, just use it to avoid strlen:
` this_executable = this_proc_name.compare(capture_proc_name) == 0;

johnplate (approved):
LGTM, just two nits.

gapii/cc/spy.cpp
hevrard: I couldn't do `capture_proc_name != nullptr` with the Android code path since that led to:
```
gapii/cc/spy.cpp:132:8: error: comparison of array 'capture_proc_name' not equal to a null pointer is always true [-Werror,-Wtautological-pointer-compare]
```

But I agree that the code was not ideal, it's much easier to do a type cast to work around the compilation warning. I've updated accordingly, with a proper comment.

gapii/cc/spy.cpp
hevrard: Agreed, unfortunately using string.compare() hits a bug in the NDK: https://github.com/android/ndk/issues/1494

I've added a comment.

johnplate: `core::get_process_name` has a bug on Android, it includes a trailing zero in the string, which increases the size of `std:string`, but not `strlen`.
Please fix `core/cc/android/process_name.cpp` with:
```
std::string get_process_name() {
  std::ifstream t("/proc/self/cmdline");
  std::string name;
  auto it = std::istreambuf_iterator<char>(t);
  while (it != std::istreambuf_iterator<char>() && *it != '\0') {
    name += *it++;
  }
  return name;
}
```

johnplate: And close the NDK bug and remove the comment

johnplate:
Great, please perform a final manual test if the string comparison works now before you merge the PR.

gapii/cc/spy.cpp
hevrard: Done

hevrard:
I confirm it works. I also noted that the PR was missing code to have the UI properly save the process name entry as parts of its settings, now fixed. I'll merge once the CI is green.

PR #772, opened 2021-04-22
Calculate peak temporary register pressure

hliatis:
Bug: b/183619276

### DIFF
--- a/gapis/shadertools/cc/staticanalysis.cpp
+++ b/gapis/shadertools/cc/staticanalysis.cpp
@@ -20,6 +20,87 @@
 #include "staticanalysis.h"
 
 #include <cstring>
+#include <map>
+#include <set>
+#include <vector>
+
+struct AnalysisBlock {
+  std::set<uint32_t> def;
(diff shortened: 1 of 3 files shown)


gapis/shadertools/cc/staticanalysis.cpp
```diff
@@ -20,6 +20,90 @@
...
+      case spv::OpArrayLength:
+        return offset != 3;
+
+      case spv::OpCompositeExtract:
+        return offset < 3;
+
+      case spv::OpCompositeInsert:
+        return offset < 4;
+
+      case spv::OpConstantPipeStorage:
```
chrisforbes, line 56: This instruction isnt allowed in Vulkan-flavored SPIRV at all. Did you see it in a shader?

hliatis: No. I was just going through the SPIRV spec quickly. I can remove it.

pmuetschard (approved):

PR #773, opened 2021-04-23
Fix ETC converter initialization

yalcinmelihyasin:
Image Format conversion in AGI done by the converters that are
registered in init time of Go Packages. There are details about this
but simply init function in go packages runs before main when the
packages are loaded.

Recently #651 moved ETC conversion to its own packages similar to ASTC
but different from ASTC, image creation remained in the parent package.
As all the conversion functions called via register map that has registered
during the init, this causes no function from the etc package, which ended up
with it's not being loaded therefore no converter registered.

This is a workaround that just create translator functions to reflect the image
create functions in the parent package and uses those functions in Vulkan/resources
to ensure the package is initialized.

This is suboptimal and only a workaround. I am creating another PR with a refactor
of the image formats. This is only a workaround until the other one can be merged.

b/184954321

### DIFF
--- a/core/image/etc/etc.go
+++ b/core/image/etc/etc.go
@@ -33,21 +33,21 @@ import (
 
 var (
 	// ETC2
-	ETC2_RGB_U8_NORM         = image.NewETC2_RGB_U8_NORM("ETC2_RGB_U8_NORM")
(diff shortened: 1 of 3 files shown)


johnplate (approved):

yalcinmelihyasin:
The next PR for the proper fix is #774

PR #774, opened 2021-04-23
Refactor ETC2 Proto structs

yalcinmelihyasin:
Refactors the ETC proto structs to simplify the code around ETC format image creation

This is a work continuation of #773 

The way that ETC protos structured, end up with lots of constructors that are not really necessary and to avoid duplicate them, actual converters and the image format definitions were in different packages that caused b/184954321

While investigating that bug, I have realized the situation and created an initial fix. This requires more work later that I discuss in the design doc attached to refactor work's bug tracker.

For a summary, all ETC formats are under one struct and the information that buried into their name now lies as fields in proto, which ended up having much more simpler code to interact with it.

Refactor Work Bug Tracking: b/186110226


johnplate (approved):

core/image/image.proto
```diff
@@ -93,25 +86,30 @@ message FmtATC_RGBA_EXPLICIT_ALPHA_AMD {
...
-message FmtETC2_R_S11_NORM {
-}
-message FmtETC2_RG_S11_NORM {
+
+message FmtETC2 {
+  enum ColorMode {
+    R = 0;
+    RG = 1;
+    RGB = 2;
+    RGBA = 3;
```
pmuetschard, line 98: RGBA is redundant. RGBA is RGB + alpha != ALPHA_NONE.
Also, signed = true makes no sense for most formats, similarly srgb = true. Maybe instead of having these extra booleans, you should expand the ColorMode enum to be : R, SIGNED_R, RG, SIGNED_RG, RGB and SRGB. If we did that, the only non-sense would be an alpha mode of anything but NONE for R, SIGNED_R, RG, and SIGNED_RG, but I think I could live with that.

pmuetschard (commented):
Move the deletion of ETC1 into its own commit. A refactor shouldn't also just implicitly get rid of something.

pmuetschard (commented):
I think something went wrong with the rebasing. It still contains the whole moving stuff around that I thought was not necessary. It does say there are conflicts and cannot be merged.

yalcinmelihyasin:
> I think something went wrong with the rebasing. It still contains the whole moving stuff around that I thought was not necessary. It does say there are conflicts and cannot be merged.

I think that you are mixing this with #779. In that one I have removed the indirection as it was not necessary anymore. 

This one on the other hand, unifies the creation of all ETC formats similar to ASTC therefore we have only one newETC function is called with different parameters based on the actual format(like ASTC). 

If you are talking about another stuff please let me know.

pmuetschard:
OK, can you please fix it so that GH no longer complains about conflicts and I can be sure I'm looking at what actually will be merged? I think you might just have to get rid of the merge commit you did.

pmuetschard (approved):

PR #777, opened 2021-04-28
Fix GAPII Swapchain image reuse handling (#740)

### LINKED ISSUE #740: Wrong image shown for PUBG MOBILE LITE frame capture
**Environment information:**
 - AGI version: agi-1.1.0-dev-20210318
 - Host OS: gLinux
 If tracing on Android:
 - Device model: Pixel 4 (flame), but problem seen on Pixel 4 XL 3+ months ago
 - Android Version: Android 11 latest quarterly patch release (RQ2A.210305.006)

**Bug description**
Replaying a GLES frame capture of PUBG MOBILE LITE and clicking on any draw command (or the final vkQueuePresentKHR) shows an 8x8 color image instead of the actual color image contents.  The correct depth-buffer image is shown

**Reproduction steps**
1) Open a trace from gameplay
2) Click on "vkQueuePresentKHR" and/or any glDraw* command under the 2nd vkQueueSubmit hierarchy.  Observe that the color image is 8x8 and that the depth image is more-reasonably sized (e.g. 536x1136 on my Pixel 4).

**Screenshots**

Screenshot from 2021-03-22 17-00-05
Screenshot from 2021-03-22 17-00-11
Screenshot from 2021-03-22 17-00-20

- I have a trace file from just now, as well as a trace file from December.  GitHub won't allow me to upload them (unsupported type and too large).  Here's a pointer to the file in GoogleDocs: https://drive.google.com/file/d/1KODZtyOjWKGDmlFvJ2-ZJ3DiAkCvrytZ/view?usp=sharing
- Here's logcat output:
[agi_210322b_logcat_PUBG.txt](https://github.com/google/agi/files/6185780/agi_210322b_logcat_PUBG.txt)

mark-lunarg:
In some cases `vkGetSwapchainImagesKHR()` can return images that already exist in GAPII's Images object map. This image 
information can be out of date and must be updated to match the parameters specified at swapchain creation time.

Fixes #740

### DIFF
--- a/gapis/api/vulkan/extensions/khr_swapchain.api
+++ b/gapis/api/vulkan/extensions/khr_swapchain.api
@@ -248,30 +248,36 @@ cmd VkResult vkGetSwapchainImagesKHR(
     images := pSwapchainImages[0:count]
     for i in (0 .. count) {
       images[i] = ?
-      if !(images[i] in Images) {
+      if (images[i] in Images) {
+        // When we get images from a swapchain created from an "oldSwapchain" (retired but
+        // not yet destroyed), we may get image handles that are already in the state.
+        Images[images[i]].Info = swapchainObject.Info
+        Images[images[i]].ImageAspect = as!VkImageAspectFlags(VK_IMAGE_ASPECT_COLOR_BIT)
+      } else {
         object := new!ImageObject(
           Device:             device,
           VulkanHandle:       images[i],
           Info:               swapchainObject.Info,
           ImageAspect:        as!VkImageAspectFlags(VK_IMAGE_ASPECT_COLOR_BIT)
         )
-        object.IsSwapchainImage = true
-        width := swapchainObject.Info.Extent.width
-        height := swapchainObject.Info.Extent.height
-        format := swapchainObject.Info.Format
-        layer := new!ImageLayer()
-        level := new!ImageLevel(Width: width,Height:  height,Depth:  1,Layout:  VK_IMAGE_LAYOUT_UNDEFINED)
-        elementAndTexelBlockSize := getElementAndTexelBlockSize(format)
-        // Roundup the width and height in the number of blocks.
-        widthInBlocks := roundUpTo(width, elementAndTexelBlockSize.TexelBlockSize.Width)
-        heightInBlocks := roundUpTo(height, elementAndTexelBlockSize.TexelBlockSize.Height)
-        size := widthInBlocks * heightInBlocks * elementAndTexelBlockSize.ElementSize
-        level.Data = make!u8(size)
         object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT] = new!ImageAspect()
-        object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT].Layers[0] = layer
-        object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT].Layers[0].Levels[0] = level
         Images[images[i]] = object
       }
+      object := Images[images[i]]
+      object.IsSwapchainImage = true
+      width := swapchainObject.Info.Extent.width
+      height := swapchainObject.Info.Extent.height
+      format := swapchainObject.Info.Format
+      layer := new!ImageLayer()
+      level := new!ImageLevel(Width: width,Height:  height,Depth:  1,Layout:  VK_IMAGE_LAYOUT_UNDEFINED)
+      elementAndTexelBlockSize := getElementAndTexelBlockSize(format)
+      // Roundup the width and height in the number of blocks.
+      widthInBlocks := roundUpTo(width, elementAndTexelBlockSize.TexelBlockSize.Width)
+      heightInBlocks := roundUpTo(height, elementAndTexelBlockSize.TexelBlockSize.Height)
+      size := widthInBlocks * heightInBlocks * elementAndTexelBlockSize.ElementSize
+      level.Data = make!u8(size)
+      object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT].Layers[0] = layer
+      object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT].Layers[0].Levels[0] = level
       swapchainObject.SwapchainImages[i] = Images[images[i]]
     }
     pSwapchainImageCount[0] = count


gapis/api/vulkan/extensions/khr_swapchain.api
```diff
@@ -248,30 +248,31 @@ cmd VkResult vkGetSwapchainImagesKHR(
...
-        // Roundup the width and height in the number of blocks.
-        widthInBlocks := roundUpTo(width, elementAndTexelBlockSize.TexelBlockSize.Width)
-        heightInBlocks := roundUpTo(height, elementAndTexelBlockSize.TexelBlockSize.Height)
-        size := widthInBlocks * heightInBlocks * elementAndTexelBlockSize.ElementSize
-        level.Data = make!u8(size)
-        object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT] = new!ImageAspect()
-        object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT].Layers[0] = layer
-        object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT].Layers[0].Levels[0] = level
-        Images[images[i]] = object
+      if (images[i] in Images) {
```
hevrard, line 251: Add a short comment here to clarify this case, e.g.:

```
// When we get images from a swapchain created from an "oldSwapchain" that is
// retired but destroyed, here we may get image handles that are already in the state.
```

pmuetschard (commented):
I'm not certain this is correct. If this fixes tracing an app, I think the fix is hiding an underlying issue. We remove all the image in the swapchain destroy

hevrard:
@pmuetschard we discussed that in our meeting with @ianelliottus and @mark-lunarg : what happens here is that the previous swapchain is not destroyed, but only "retired" by being used as an "oldSwapchain" when creating a new swapchain. Given that the retired swapchain is not destroyed, AGI does not removes these images from the state.

mark-lunarg:
@pmuetschard, in these cases the swapchain has not yet been destroyed, it has only been retired.

gapis/api/vulkan/extensions/khr_swapchain.api
mark-lunarg: Added, thanks!

hevrard (approved):

ianelliottus:
tl;dr: swapchain images from "retired" swapchains can and are reused by Android.  Therefore, a capture-time change, like this, are needed.

Yes, the Vulkan spec allows the presentation engine (PE) part of Vulkan to reuse images from an oldSwapchain for a recreated swapchain.  I've seen the Android PE do this (which is implemented in the Android Vulkan loader).  If the oldSwapchain has been used by ANGLE, ANGLE queues it to be destroyed later.  It's possible that ANGLE doesn't destroy it until the app exits.  It depends on what the app does.  If desired, I'm happy to chat about some of the reasons swapchains were designed this way.

What Mark found is that ANGLE initially created an 8x8 swapchain, and then "retired" this one by making it the oldSwapchain when a newer, much larger swapchain was created.  The oldSwapchain wasn't destroyed, but the PE reused some of the images.

pmuetschard (commented):
OK, I see what you are trying to fix, however, this is not the correct way to do it. If an application calls `vkGetSwapchainImagesKHR` and creates `VkImageView`s for the returned images, then calls `vkGetSwapchainImagesKHR` again, we are stranding those image view's references to the images. We cannot blindly destroy and re-create objects that something else in the state may hold a reference to. What we need to do is if the image already exists, update it to match the current swapchain. I'm assuming that if the implementation re-uses an image, the application can re-use the image view, too, no?

ianelliottus:
> What we need to do is if the image already exists, update it to match the current swapchain. I'm assuming that if the implementation re-uses an image, the application can re-use the image view, too, no?

Let me clarify.  The image handle is reused, and the underlying memory may be reused.  I believe it should be treated as a new VkImage, with the need to create a new image view.  That's what ANGLE does.

I'd have to look more at the spec to refresh my memory if you're not convinced.

mark-lunarg:
@pmuetschard, this does happen as you'd suggest, but I checked `DestroyImage()` in **image.api** and didn't notice any such cleanup so I thought it was OK.  `ImageView `objects can also exist even though their underlying `image `has been deleted, though they become Invalid and the app should clean them up (these particular `ImageViews `do get deleted later).  So, I will do the image-exists check, and if it does exist I'll avoid creating a new one and just reset all the fields, leaving the `ImageViews `intact?

pmuetschard:
> So, I will do the image-exists check, and if it does exist I'll avoid creating a new one and just reset all the fields, leaving the ImageViews intact?

I think that should work.

What I'm talking about is our `ImageViewObject`. It has a reference to a `ImageObject`. If an app calls `vkGetSwapchainImagesKHR` multiple times and the second and subsequent time, we remove the `ImageObject` from our state and create a new one, the `ImageViewObject` will not get its reference updated and thus have a dangling pointer. It is valid for an app to destroy a `vkImage` before the `vkImageView`, which results in us removing the `ImageObject` and the same dangling pointer, however that is fine, because the app may do nothing with that `vkImageView`, but destroy it.

mark-lunarg:
@pmuetschard, the first version of the suggested changes is up.  I'm unsure of all the implications related to GO garbage collection, so please take a close look.

gapis/api/vulkan/extensions/khr_swapchain.api
```diff
@@ -248,35 +248,41 @@ cmd VkResult vkGetSwapchainImagesKHR(
...
+      size := widthInBlocks * heightInBlocks * elementAndTexelBlockSize.ElementSize
+      level.Data = make!u8(size)
+      object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT].Layers[0] = layer
+      object.Aspects[VK_IMAGE_ASPECT_COLOR_BIT].Layers[0].Levels[0] = level
       swapchainObject.SwapchainImages[i] = Images[images[i]]
     }
     pSwapchainImageCount[0] = count
   }
-
+
```
hevrard, line 285: remove this trailing whitespace

hevrard (approved):
I think this looks aligned to what Pascal was describing.

gapis/api/vulkan/extensions/khr_swapchain.api
mark-lunarg: Sorry, gone now.

PR #778, opened 2021-04-29
Integrate Perfetto's existing power rail data into AGI.

stellama0208:
If a device has the capability to capture power rail data, give users the option
to profile it in System Profiler. And for a given perfetto trace, if it contains 
power.rail.* information, let AGI visualize it in counter track format.

Bug: b/177067971, #1.
Test: will need certain devices for testing, the device requirement is in the bug ticket's #8.


hevrard (approved):
LGTM, I confirm it works. Maybe move the "collect power rails" tickbox to below the Poll Rate?
powerrail

stellama0208:
> LGTM, I confirm it works. Maybe move the "collect power rails" tickbox to below the Poll Rate?
> powerrail

Sounds good! Done that with a new commit. :)

PR #779, opened 2021-04-29
Fix ETC converter initialization

yalcinmelihyasin:
Image Format conversion in AGI done by the converters that are
registered in init time of Go Packages. There are details about this
but simply init function in go packages runs before main when the
packages are loaded.

Recently #651 moved ETC conversion to its own packages similar to ASTC
but different from ASTC, image creation remained in the parent package.
As all the conversion functions called via register map that has registered
during the init, this causes no function from the etc package, which ended up
with it's not being loaded therefore no converter registered.

This is a workaround to just add ETC package to resources.go so that ETC package 
can be initialised.

This is suboptimal and only a workaround. I am creating another PR with a refactor
of the image formats. This is only a workaround until the other one can be merged.

### DIFF
--- a/gapis/api/vulkan/resources.go
+++ b/gapis/api/vulkan/resources.go
@@ -22,6 +22,7 @@ import (
 	"github.com/google/gapid/core/data/id"
 	"github.com/google/gapid/core/image"
 	"github.com/google/gapid/core/image/astc"
+	"github.com/google/gapid/core/image/etc"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/stream/fmts"
 	"github.com/google/gapid/gapis/api"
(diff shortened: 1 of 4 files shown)


hevrard (approved):
LGTM, let's make sure it passes windows CI before merging.

pmuetschard (dismissed):
Simply add an unused import to the etc.go file in the image package, rather than adding additional round trips:
```
import (
_ "foo/bar"
)
```
This is a common practice for situations like this (

yalcinmelihyasin:
> Simply add an unused import to the etc.go file in the image package, rather than adding additional round trips:
> 
> ```
> import (
> _ "foo/bar"
> )
> ```
> 
> This is a common practice for situations like this (

Oh I didn't know that this was an option.

johnplate (approved):
LGTM with Pascal's comment fixed

yalcinmelihyasin:
> Simply add an unused import to the etc.go file in the image package, rather than adding additional round trips:
> 
> ```
> import (
> _ "foo/bar"
> )
> ```
> 
> This is a common practice for situations like this (

This does not work. 

For this to work, image needs a dependency on etc2 package which already depends on image package. So, it's a circular dependency. Instead I've added it to vulkan package similar to ASTC.

gapis/api/vulkan/resources.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/google/gapid/core/data/id"
 	"github.com/google/gapid/core/image"
 	"github.com/google/gapid/core/image/astc"
+	_ "github.com/google/gapid/core/image/etc"
```
hevrard, line 25: Add a comment that says **why** we need this here, e.g.:

```
// Required such that core/image/etc:init() registers ETC formats
```

hevrard (approved):
Just add a comment saying the reason behind the unused import, because this would be non-trivial to anyone re-reading this code in a few months time!

hevrard (approved):

PR #782, opened 2021-05-06
Add option to load validation layer at capture time on Android

hevrard:
This change adds a boolean "LoadValidationLayer" option to trace
options. When true, we load the Vulkan validation layer (VVL) below
our spy layer. This is useful for AGI developers to debug capture-time
issues.

This is only supported on Android, because we can rely on the VVL
which is present in our own APK. On desktop, we have no guarantee that
the VVL is available anywhere.

The const that defines the vulkan validation layer name is moved to
core/vulkan/loader package such that it can be easily accessed in both
vulkan transform and gapii client code.

Bug: n/a
Test: manual


hevrard:
screenshot

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -694,6 +698,8 @@ private void updateOnApiChange(
             isPerfetto ? Messages.CAPTURE_TRACE_PERFETTO : Messages.CAPTURE_TRACE_GRAPHICS);
         withoutBuffering.setEnabled(!isPerfetto);
         withoutBuffering.setSelection(!isPerfetto && trace.getWithoutBuffering());
+        loadValidationLayer.setEnabled(!isPerfetto && getSelectedDevice().isAndroid());
```
purvisa-at-google-com, line 701: I think we need to document that this new option doesn't work when doing a perfetto trace. Ideally even deselect/gray the box in the UI to communicate this.

hevrard: Good point, and that's what this line does. Here loadValidationLayer is a `Button`, and `setEnabled()` grays it out when the trace is Perfetto, or the device is not Android.

purvisa-at-google-com: Ah sorry silly me!

mark-lunarg:
👍

johnplate (approved):

PR #783, opened 2021-05-07
[WIP] Be robust to unexpected CmdSubmissionKey occurences

hevrard:
A certain app leads to the value of `c` being outside the range of
`indices`. Be robust to this case. That's probably not the proper fix.

Bug: b/18728916


hevrard:
The proper fix is https://github.com/google/agi/pull/789

PR #784, opened 2021-05-08
Support render VK_FORMAT_R16G16_SFLOAT images in FrameBuffer and Texture view.

stellama0208:
Create a specific ArrayImage model for VK_FORMAT_R16G16_SFLOAT typed images, so
that 1. the image will get rendered with correct settings, 2. the image data buffer
will get decoded in the right way.

Bug: b/177572847.
Test: test with a frame capture that records VK_FORMAT_R16G16_SFLOAT typed textures.

### DIFF
--- a/gapic/src/main/com/google/gapid/image/ArrayImage.java
+++ b/gapic/src/main/com/google/gapid/image/ArrayImage.java
@@ -358,6 +358,134 @@ public abstract class ArrayImage implements com.google.gapid.image.Image {
   }
 
   /**
+   * An {@link ArrayImage} that represents an R16G16 image with 16bit float color channels.
+   */
+  public static class RG16FloatImage extends ArrayImage {
+    private final FloatBuffer buffer; // Stores {Red, Green} pairs.
+    private final PixelInfo info;
+
+    public RG16FloatImage(Image.Key key, int width, int height, int depth, byte[] data) {
+      super(width, height, depth, 4, data, GL30.GL_RG16F, GL30.GL_RG, GL30.GL_HALF_FLOAT);
+      this.buffer = FloatBuffer.allocate(data.length / 2);
+      for (int i = 0; i + 3 < data.length; i+=4) {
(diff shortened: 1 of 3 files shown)


yalcinmelihyasin:
I think that it is better to have this logic in core/image in server side. Because all the other similar code lies there. We can wait for @pmuetschard's view on it but I would be favor in moving it to core/image

stellama0208:
Thanks for everyone's pointer, sorry I didn't know there's related logic in gapis to deal with format converting before, I will find a more proper space to inject the fix.

> I'm a bit confused why we need special treatment in the UI for this specific format and not say, a 1, 3, or 4 channel 16bit float as well? Still, requesting a RG 16bit-float format as RGBA 32bit float should be handled correctly on the server, so I think this is trying to circumvent a bug on the server rather than fixing it. I suspect that the bug could be with the implicit tone-mapping done on the server for float formats. Is it interfering with the implicit channel expansion?

And thanks Pascal for pointing out possible errors, following this trajectory, I checked and found that for `RG_16F`, server indeed execute an implicit tone-mapping and implicit channel expansion to transform it into `RGBA_U8` format. This part of work actually works fine as intended, I could see reasonable bitstream output. I found the gap actually comes from the client doesn't know this mapping happens, so it wrongly think it should receive a `RGBA 32F` image. As we are here, I think we could either:
1. At client, dispatch properly so that the client knows it will receive a transformed `RGBA_U8` image for `RG_16F` image.
2. At server, don't use the implicit converter, register a specific one to transform `RG_16F` to `RGBA_32F`.

I will take a look at both approaches for now to see which implementation is more reasonable, but at the meantime, @pmuetschard if you have a clear preference for one of the approaches, or another way to handle it, feel free to let me know, thanks!

pmuetschard:
Hmm that's odd. The *_16F formats should not go via RGBA_U8 to get to *_32F. There is/should be a direct float->float converter that expands the range.

stellama0208:
Close this PR because #792 contains a more relevant solution.

PR #788, opened 2021-05-12
Avoid out-of-index array query when command and driver report different renderpass count

stellama0208:
It was observed recently in some game that for a given renderpass record
identified by {submission, commandBuffer, renderPass, renderTarget}, the
record count reported by command buffer walk-through may be less than that
reported by vulkan driver. This patch will do an extra check at slice
association stage, so that when this unexpected scenario happens, it will not
make an out-of-index array query that paralyze gapis totally, but to log a
warning message and skip this GPU slice.

There might be more investigation needed to see why the discrepancy happens
(may need studio or driver team's help), but this patch will serve as a
quick fix to unblock our users and let them see other data.

Bug: b/187289167.
Test: test with the game provided in the bug ticket.


stellama0208:
Had some thought, and found the bug should root from miss capturing NextSubpass command's info. Will create another PR to solve this problem.

PR #789, opened 2021-05-13
Mark down subpasses's info when preparing GPU slices' association map.

stellama0208:
For a Surface slice reported by adreno driver, it will get connencted
to a representative command in the command list, and the command could be
either a vkBeginRenderPass or a vkCmdNextSubpass. Before this PR, AGI
only captures the vkBeginRenderPass when creating the command-slice
association map, and thus the renderpass count recorded by command buffer
walk-through will be less than that reported by vulkan driver, which would
later result in out-of-index array query panic and paralyze the AGI server.
This patch will add in the capture of vkCmdNextSubpass command to solve
the problem.

Bug: b/187289167.
Test: test with the apk in the ticket, or a capture containing subpasses.


hevrard (approved):

hevrard:
Typo in the commit message's first line (you can edit this when doing the "squash and merge" on github): connencted -> connected

PR #792, opened 2021-05-19
Fix a bug when setting the end of the framebuffer histogram's default range.

stellama0208:
AGI will calculate and set default histogram range in framebuffer view and the general
intention is to set the range's end to the bin at 99% percentile in the histogram.

However, different channels will reach 99% percentile at different bin, and previously,
because of an early return, when iterating through the bins, if AGI found any of the
channels already reaches 99% percentile, it will immediately return that bin as the end
of the whole range. This will cause trouble in cases where different channels reaches
99% at very different bins. e.g. for a R16G16 image, the blue channel is always 0,
so 100% of the pixels falls at the value 0, and thus blue channel's 99% percentile bin reaches
early at bin 0, if we use this as the default histogram range's end, it will erase detail
for the red and yellow channel.

Ideally the range's end should be the last 99% percentile bin across all channels, so
that the main distribution of all channels will get displayed.This PR makes sure the range
starts at the *first* 1% percentile bin, and ends at the *last* 99% percentile bin.

Bug: b/177572847.

### DIFF
--- a/gapic/src/main/com/google/gapid/image/Histogram.java
+++ b/gapic/src/main/com/google/gapid/image/Histogram.java
@@ -18,12 +18,15 @@ package com.google.gapid.image;
 import static java.util.Arrays.stream;
 import static java.util.stream.Collectors.toSet;
 
+import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.gapid.image.Image.PixelInfo;
 import com.google.gapid.proto.stream.Stream;
 import com.google.gapid.proto.stream.Stream.Channel;
 import com.google.gapid.util.Range;
 
+import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.DoubleStream;
 import java.util.stream.IntStream;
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/image/Histogram.java
```diff
@@ -87,13 +90,25 @@ public Range getInitialRange(double snapThreshold) {
...
+   * @param lastBin if true, for all the percentile bins from different channels, choose the last
+   * bin, otherwise choose the first bin.
    * @param high if true, return the upper limit on the percentile's bin, otherwise the lower limit.
    * @return the absolute pixel value at the specified percentile in the histogram.
    */
-  private double getPercentile(int percentile, boolean high) {
-    int bin = bins.getPercentileBin(percentile, channels);
-    return (bin < 0) ? mapper.limits.max :
-        getValueFromNormalizedX((bin + (high ? 1 : 0)) / (double)bins.count());
+  private double getPercentile(int percentile, boolean lastBin, boolean high) {
```
pmuetschard, line 98: Is there ever a reason `lastBin != high`? In all our usage we set them to the same value, so I think you can just use one parameter.

pmuetschard (approved):

gapic/src/main/com/google/gapid/image/Histogram.java
stellama0208: Yes indeed, they are the same in all usage cases. Sounds good, just combined them and updated the annotations!

PR #797, opened 2021-05-25
Rename GPU activity to GPU queues

hevrard:
The term "GPU activity" appears nowhere in the results, so it's
confusing to use it in the config menu. As both Adreno and Mali return
a "GPU Queue" track when GPU activity is ticked, lets use this term to
in the config menu as well.

Bug: b/186401470
Test: manual


pau-baiget (approved):

pmuetschard (changes requested):
Let's discuss on the bug.

PR #798, opened 2021-05-25
Support opening Fuchsia traces with Perfetto.

pmuetschard:
Perfetto's trace processor can handle Fuchsia profile traces (FXT files).


hevrard (approved):
The CI failure looks like a flake

PR #799, opened 2021-05-26
Avoid the need to restart AGI after installing ANGLE

hevrard:
Before this change, if you start AGI with a device that has no ANGLE,
then use AGI to install ANGLE, then try and start an ANGLE trace,
you'll get a "ANGLE requested but no ANGLE package found" error. This
is because installing ANGLE doesn't update the device info, such that
AGI still thinks ANGLE is not present on the device.

The way device info is stored is non-trivial: there's a background
task (see monitorAndroidDevices()) that monitors devices and store
them in a "registry" that is queried when we list devices. It sounds
like the proper way to update device info is to remove the device from
the registry, and let the monitor re-scan it accordingly. It sounds
relevant to do it after any APK install that may change the device
info.

On the UI side, we also need to make sure to refresh the device info
after installing ANGLE.

Bug: b/186611374
Test: manual, making sure to run AGI on a device with no ANGLE on it.

### DIFF
--- a/core/os/android/adb/device.go
+++ b/core/os/android/adb/device.go
@@ -350,6 +350,15 @@ func scanDevices(ctx context.Context) error {
 	return nil
 }
 
+// rescanDevice force-refresh the device info for a particular device.
+func rescanDevice(ctx context.Context, d bind.Device) error {
+	registry.RemoveDevice(ctx, d)
+	cacheMutex.Lock()
+	delete(cache, d.Instance().Serial)
+	cacheMutex.Unlock()
+	return scanDevices(ctx)
+}
+
 func parseDevices(ctx context.Context, out string) (map[string]bind.Status, error) {
 	a := strings.SplitAfter(out, "List of devices attached")
 	if len(a) != 2 {
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -183,7 +183,14 @@ func (b *binding) InstallAPK(ctx context.Context, path string, reinstall bool, g
 	// Because adb root/unroot is outside the program's state there's not much we can do to prevent this
 	// race conditon though. Someone could always run adb from a terminal at a bad time...
 	b.Unroot(ctx)
-	return b.Command("install", args...).Run(ctx)
+	if err := b.Command("install", args...).Run(ctx); err != nil {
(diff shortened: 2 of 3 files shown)


core/os/android/adb/device.go
```diff
@@ -350,6 +350,15 @@ func scanDevices(ctx context.Context) error {
 	return nil
 }
 
+// rescanDevice force-refresh the device info for a particular device.
+func rescanDevice(ctx context.Context, d bind.Device) error {
+	registry.RemoveDevice(ctx, d)
+	cacheMutex.Lock()
+	delete(cache, d.Instance().Serial)
+	cacheMutex.Unlock()
+	return scanDevices(ctx)
```
pmuetschard, line 359: This seems a bit drastic. Will removing and re-adding a device change its identifier? If so, that will cause all sorts of other issues.

hevrard: By installing or updating ANGLE, we effectively change the device configuration, and hence its ID.

A device instance ID is computed via `core/os/device/instance.go:GenID()`:

```
func (i *Instance) GenID() {
	key := fmt.Sprintf("%v:%v:%v", i.Serial, i.Name, i.Configuration.String())
	i.ID = NewID(id.OfString(key))
}
```

And the configuration contains ANGLE details, from `core/os/device/device.proto`:

```
message Configuration {
...
  // ANGLE info.
  ANGLE angle = 6;
}

...

message ANGLE {
  string package = 1;
  int32 version = 2;
}
```

So in the current situation I believe it is necessary to remove the previous device instance and rescan the device entirely to get a fresh set of information.

The idea behing this change is that whenever we install an APK on a device, we might be changing some parts of the device configuration, so we rescan it, and the effect should be similar to unplugging the device, install the APK on it separately, and plug it in again.

Another possible approach could be to not take the ANGLE version into consideration for computing the ID, such that we can just update ANGLE details while keeping the same ID, but I get the feeling it goes against the current approach of computing the ID as a hash of all configuration details.

damian-cell (approved):

damian-cell (approved):

damian-cell (approved):

hevrard:
Partly superseded by #838

PR #801, opened 2021-05-28
Clean up the device interface.

pmuetschard:
- Introduces a new `Desktop` device interface that contains all the functionality for the local host or remote desktop like devices.
- Cleans up the "simple" device implementation to no longer assume it is used only for the local host, which it wasn't.
- Removes unused/broken device functionality from the Android device implementation.
- Removes Android functionality from the common interface.


hevrard (approved):
Just need to fix presubmit.

PR #807, opened 2021-06-03
Disable UI support to edit shader source

matled:
Pushing the shader changes doesn't appear to work yet. Until we fix the issue,
hide the button and make the source code not editable.

This introduces a new flag to enable features that are under development. This
is intended for minor features that need to stay disabled while they are
implemented or issues are resolved.

Bug: 188434910
Test: manual

### DIFF
--- a/gapic/src/main/com/google/gapid/views/ShaderView.java
+++ b/gapic/src/main/com/google/gapid/views/ShaderView.java
@@ -42,6 +42,7 @@ import com.google.gapid.rpc.Rpc;
 import com.google.gapid.rpc.RpcException;
 import com.google.gapid.rpc.SingleInFlight;
 import com.google.gapid.rpc.UiCallback;
+import com.google.gapid.util.Experimental;
 import com.google.gapid.util.Loadable;
 import com.google.gapid.util.Messages;
 import com.google.gapid.widgets.LoadablePanel;
@@ -70,6 +71,7 @@ import org.eclipse.swt.widgets.TabItem;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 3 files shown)


matled:
Notes on some choices in this change:
(1) I chose to use enableAll instead of adding a custom flag. We intend to add a few more features that are not ready yet. I'm not sure each of them should get its own flag, as they will probably only be enabled by developers while they're trying to get the feature ready for use.
(2) I chose to use `Optional<Button>` instead of `setVisible(false)`. Making the button invisible still takes up space in the layout.

gapic/src/main/com/google/gapid/views/ShaderView.java
```diff
@@ -153,16 +155,24 @@ public ShaderView(Composite parent, Models models, Widgets widgets) {
...
-      if (shaderResource != null && shaderMessage != null) {
-        models.analytics.postInteraction(View.ShaderView, ClientAction.Edit);
-        models.resources.updateResource(shaderResource, API.ResourceData.newBuilder()
-            .setShader(shaderMessage.toBuilder().setSource(spirvViewer.getDocument().get()))
-            .build());
-      }
-    });
-    pushButton.setLayoutData(new GridData(SWT.RIGHT, SWT.BOTTOM, false, false));
-    pushButton.setEnabled(false);
+    if (!Experimental.enableAll.get()) {
```
hevrard, line 160: Let's have a dedicated flag for this, e.g. `enableUnstableFeatures`, rather than relying on `enableAll`.

gapic/src/main/com/google/gapid/views/ShaderView.java
```diff
@@ -282,7 +292,7 @@ private void clear() {
     if (!pinned) {
       shaderMessage = null;
     }
-    pushButton.setEnabled(false);
+    pushButton.ifPresent(b -> b.setEnabled(true));
```
hevrard, line 297: I think this should be `setEnabled(false)` here?

hevrard (changes requested):
Please capitalize the commit message title: `Disable UI ...`

hevrard:
I reviewed before seeing your comment on github.

For the flag, let's have a single `enableUnstableFeatures` flag for all the ones we'll be hiding, and let's make sure that enabling all experimental features also activates this one.

hevrard (approved):
LGTM!

PR #810, opened 2021-06-04
Show async process event tracks.

pmuetschard:
These are similar to thread tracks, except the events are async, which requires combining the data from multiple tracks, since slices in a single track cannot overlap.

Bug: http://b/188813975


hevrard (approved):
LGTM, I would have appreciated a bit more details in the commit messages.

Suggested a couple of comments that I think would help the next dev who'll look at this part in the future.

PR #812, opened 2021-06-04
Create group for draw cmds without state setup

### LINKED ISSUE #800: Inconsistent vkDraw* hierarchy in AGI command panel
**Environment information:**
 - AGI version: 1.1.0:developer
 - Host OS: Linux
 If tracing on Android:
 - Device model: Pixel 4 XL
 - Android Version: 11

**Bug description**
When viewing an OpenGL on ANGLE trace with consecutive `glDrawElementsBaseVertexOES` commands, the single `vkCmdDrawIndexed` command for the second draw is not organized under a DrawIndexed node like the first. This is inconsistent.

**Reproduction steps**
Capture gameplay from "Fate/Grand Order", inspect glDrawElementsBaseVertexOES calls from first RenderPass.

**Stacktrace**
N/A

**Screenshots**
draw_tree

N/A

mikes-lunarg:
For consecutive vkCmdDraw* commands, create a draw group in the command tree for the second draw (without any state setup) so it matches how other draws (with state setup) are displayed. There will only be a single draw command in the group for this case.

Fixes #800

### DIFF
--- a/gapis/api/vulkan/vulkan.go
+++ b/gapis/api/vulkan/vulkan.go
@@ -455,19 +455,21 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
 
 			// Handle draw commands grouping.
 			cmdName := cb.CommandReferences().Get(uint32(i)).Type().String()
-			isDrawCmd := strings.HasPrefix(cmdName, "cmd_vkCmdDraw") || strings.HasPrefix(cmdName, "cmd_vkCmdDispatch")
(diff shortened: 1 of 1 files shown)


mikes-lunarg:
Pushed update to group vkCmdClearAttachments like a draw

ianelliottus (commented):
LGTM, but I'd like Hugues or somebody else from the AGI team to review this.

hevrard (approved):
Thanks for this!

PR #818, opened 2021-06-08
Add support for VK_EXT_vertex_attribute_divisor

hevrard:
This extension lets us add a "divisor" to some vertex attribute
bindings used in instanced rendering. See chapter "Vertex Attribute
Divisor in Instanced Rendering" in the Vulkan spec, which has a bit of
pseudo code at the end that recaps how divisor values may be used.

Bug: b/190150915


yalcinmelihyasin (commented):
Fine for me but please wait for the Pascal's input.

hevrard:
@pmuetschard thanks for this!

I had also forgotten to add the new API file that defines the extension file, now fixed.

hevrard:
@pmuetschard PTAL

purvisa-at-google-com (approved):
What's here looks correct to me. I'm in a bad position to say if something has been forgotten or missed though. Up to you if you want to wait for Pascal or preemptively submit. Guess this CL is low danger for breaking things that don't use the vertex divisor feature.

matled:
Thanks for adding me! Similar to Alan, I can't spot any issue but also don't have any experience, good to see for me as a reference how an extension is added.

pmuetschard (commented):
It is now missing the state reconstruction bits. You need to pass in the extension struct to the vkCreateDevice if it's present. It should be pretty obvious, when you look at the code, where to put it. Look at some of my recent commits, where I've done similar things (i.e. look at the history of state_rebuilder.go)

PR #820, opened 2021-06-08
JNI_OnLoad is gone.

pmuetschard:


hevrard (approved):
Thanks for the cleanup!

PR #823, opened 2021-06-10
Hide the report tab

matled:
The report tab is using an old version of the vulkan validation layer,
which are buggy and may report false positives. Hide the report tab
until this issue is fixed.

Bug: 188416598
Test: manual


hevrard (approved):
Thanks for this!

PR #825, opened 2021-06-10
Cleanup gapis build-time config options and document them in DEVDOC

hevrard:
Cleanup details:

- Remove all obsolete configs that do not appear anywhere else in the
  code.

- Also remove DebugDependencyGraph since
  dependencyGraphBuilder.debug() was never used, and seemed to be a
  leftover from some debugging session

- Set DumpValidationTrace to false, as we don't want this to be set in
  our dev-release builds

- Add a proper documentation for each config option

Bug: n/a

### DIFF
--- a/gapis/config/config.go
+++ b/gapis/config/config.go
@@ -27,10 +27,12 @@ const (
 	// dependency graph.
 	DisableDeadCodeElimination = false
 
-	// DeadSubCmdElimination prevents the elimination of subcommands in DCE.
+	// DeadSubCmdElimination prevents the elimination of subcommands in dead
+	// code elimination.
 	DeadSubCmdElimination = false
 
-	// DebugDeadCodeElimination activates various debug logs related to DCE.
+	// DebugDeadCodeElimination activates various debug logs related to dead
+	// code elimination.
 	DebugDeadCodeElimination = false
 
 	// DumpReplayProfile dumps the perfetto trace of a replay profile.
@@ -40,7 +42,7 @@ const (
 	DumpValidationTrace = false
 
 	// AllInitialCommandsLive forces all initial commands to be considered as
-	// live when computing DCE.
+	// live when computing dead code elimination.
 	AllInitialCommandsLive = false
 
 	// LogExtrasInTransforms logs all commands' extras together with transforms.
@@ -60,7 +62,7 @@ const (
 	LogTransformsToFile = false
 
 	// LogTransformsToCapture creates a gfxtrace that stores the commands
-	// obtained after each transformation has been applied.
+	// obtained after all transforms have been applied.
 	LogTransformsToCapture = false
 
 	// LogInitialCmdsToCapture creates a gfxtrace that stores initial commands.
--- a/gapis/config/config.go
+++ b/gapis/config/config.go
@@ -16,25 +16,53 @@
 package config
 
 const (
-	DebugReplay                = false
-	DebugReplayBuilder         = false
+	// DebugReplay activates various debug logs related to replay.
+	DebugReplay = false
+
+	// DebugReplayBuilder activates various debug logs and checks related to the
+	// creation of a replay payload.
+	DebugReplayBuilder = false
+
+	// DisableDeadCodeElimination prevents the early computation of the
+	// dependency graph.
 	DisableDeadCodeElimination = false
-	DeadSubCmdElimination      = false
-	DebugDeadCodeElimination   = false
-	DebugDependencyGraph       = false
-	DumpReplayProfile          = false
-	DumpValidationTrace        = true
-	AllInitialCommandsLive     = false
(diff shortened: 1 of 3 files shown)


gapis/config/config.go
```diff
@@ -16,25 +16,53 @@
...
-	// it will only print the last mapping).  Only works for Vulkan.
-	LogMappingsToFile        = false
-	LogTransformsToFile      = false
-	LogTransformsToCapture   = false
-	LogInitialCmdsIssues     = false
-	LogInitialCmdsToCapture  = false
-	SeparateMutateStates     = false
-	CheckRebuiltStateMatches = false
+
+	// DeadSubCmdElimination prevents the elimination of subcommands in DCE.
```
matled, line 30: nit: Using the DCE in the comment caused me to stumble a bit while reading and trying to figure out what DCE means. From the other options it becomes obvious that it is "dead code elimination". 

Options:
(1) Always spell out "dead code elimination" instead of DCE. This is somewhat my preference, as it keeps the flow while reading and I don't see much downside of the slightly more verbose comment.
(2) Introduce DCE as an abbreviation above, e.g. "prevents the early computation of the dependency graph, which is used to perform dead code elimination (DCE)."

Option (1) would apply to all uses of DCE here. I also prefer option 1 as the reader may just arrive here looking at a specific option, and doesn't read from top to bottom, i.e. even if we introduce the abbreviation, they may miss it.

gapis/config/config.go
```diff
@@ -16,25 +16,53 @@
...
+	// original trace handles to replay client handles (if handles are reused in
+	// the trace it will only print the last mapping).  Only works for Vulkan.
+	LogMappingsToFile = false
+
+	// LogTransformsToFile logs all commands seen by each transform into a
+	// separate file for each transform.
+	LogTransformsToFile = false
+
+	// LogTransformsToCapture creates a gfxtrace that stores the commands
+	// obtained after each transformation has been applied.
```
matled, line 63: (1) From my understanding, this does not include the initial state setup. This may be worth pointing out as a note?
(2) Also, there are some setup/tear-down commands included. Maybe worth pointing out similarly?
(3) nit: "after all transformations have been applied"

E.g.
LogTransformsToCapture creates a gfxtrace that stores the commands after all transformations have been applied.
Note that transformations introduce additional commands for [not sure exactly, allocate memory? preserve state?].
Note also that for mid-execution captures, the resulting trace will not include the initial state setup (see `LogInitialCmdsToCapture`). In general it is therefore an invalid trace and cannot be replayed.

matled (approved):
Thanks a lot for adding documentation here! This is super helpful.


Note: This PR performs quite a few different things by (1) adding documentation (2) changing defaults and (3) removing dead code. In general, I think it is preferable to separate concerns to make it easier to review each individually and also revert them individually if necessary. E.g. without reading the description I would probably not even notice that the default of one value changed. Happy to leave this PR as-is, just leaving this comment as feedback for future PRs.

yalcinmelihyasin (commented):
Except the mattias's comments, lgtm.

gapis/config/config.go
hevrard: Good point, I've spelled it out entirely here. There's a bit of a balance between avoiding acronyms and using ones that are common enough, but I've very much on the side of avoiding them for the sake of clarity and accessibility to our code.

Note that you may see DCE in other parts of the code, I didn't want to replace all occurrences as that's probably a wide change.

gapis/config/config.go
hevrard: (1) I believe it does, I just tried it to confirm. It may depend on the way you reach `gapis/api/vulkan/replay.go:API.Replay()`, but I think most if not all of the time you'll have the initial commands by then.

(2) The whole point of transforms is to add/remove commands (except maybe for `transform_file_log`), so I think we can assume developers will be aware that transforms have added/removed commands.

Transforms add commands for various reasons. To take an example, transform_read_framebuffer will add the necessary commands to read the framebuffer at a given point in the trace. These are not really documented apart from (1) being stored in a file name like `gapis/api/vulkan/transform_*.go`, and (2) a small comment at the top of the file, e.g.:

```
// readFramebuffer implements a transform that adds the necessary
// Vulkan commands to retrieve a framebuffer attachement.
type readFramebuffer struct {
```

(3) Fixed, thanks!

hevrard:
> Note: This PR performs quite a few different things by (1) adding documentation (2) changing defaults and (3) removing dead code. In general, I think it is preferable to separate concerns to make it easier to review each individually and also revert them individually if necessary.

Agreed! Github makes it harder than Gerrit to have a stack of sequential commits open for review, but I agree it's good to try and keep PRs as small as feasible, and especially separate changes that are only refactoring (no behavioral change, like adding documentation) from behavior-changing ones (like changing some defaults). I'll aim at a better job next time :)

PR #829, opened 2021-06-15
Add license for Etc2Comp

hevrard:


gapic/res/text/licenses.html
```diff
@@ -768,6 +768,23 @@ <h3>Eclipse OSGi</h3>
 property of their respective owners and are hereby recognized.
 </pre>
 
+<h3>Etc2Comp</h3>
+<pre>
+Copyright 2015 Etc2Comp Authors.
```
matled, line 773: How did you determine the year? The commit we pinned in workspace.bzl is from 2017-04-24. Should this be 2017 instead?

matled (approved):

gapic/res/text/licenses.html
hevrard: I took the year used by Etc2Comp itself, see the bottom of their main README.md: 

https://github.com/google/etc2comp#copyright

This is validated by the copyright date of their top-level CMakeFile.txt, which is probably here since the start of this project.

PR #832, opened 2021-06-17
In perftab, add recommended presets based on vendor provided counter type metadata.

stellama0208:
- Before, the presets(quick filters) in perftab are only created by users.
This PR add in some extra presets based on vendor's counter type grouping
metadata. In this way, when a user opens perftab for the first time, they
have some basic recommended presets to play with.
- Organize the functional buttons in perftab.Put them in two rows, leaving 
the preset buttons in the second row. Also add separators in preset buttons row.

Bug: b/190540726, b/190542146.


hevrard (approved):

hevrard:
@stellama0208 there's just a quick formatting fix to do to pass presubmit.

PR #833, opened 2021-06-17
Fix retired swapchain image deletion

hevrard:
The deletion of retired swapchain images could actually remove a
non-swapchain image from the global Images map that tracks Vulkan
images as part of the Vulkan state.

Consider these commands, and how our code was working before this
change:

```
vkCreateSwapchainKHR(oldSwapchain:0x123) # retires swapchain 0x123
                                         # which has img 0x456
					 # that is not acquired

vkCreateImage() # Returns a fresh new image with handle 0x456

vkDestroySwapchainKHR(0x123) # This removes img handle 0x456 from
                             # Images, and that is incorrect
```

This change fixes how retired swapchain images are handled, and
properly deletes them from Images only when it is relevant. See
code comments for details.

Bug: b/190829522
Test: manual, on an app that systematically triggered an error.

### DIFF
--- a/gapis/api/vulkan/extensions/khr_swapchain.api
+++ b/gapis/api/vulkan/extensions/khr_swapchain.api
@@ -119,6 +119,18 @@ cmd VkResult vkCreateSwapchainKHR(
   if pCreateInfo == null { vkErrorNullPointer("VkSwapchainCreateInfoKHR") }
   create_info := pCreateInfo[0]
 
+  if create_info.oldSwapchain in Swapchains {
+    old := Swapchains[create_info.oldSwapchain]
+    old.Retired = true
+    // Delete images from the oldSwapchain that is being retired, unless these
+    // images are acquired, in which case they may still be used and presented.
+    for _, i, img in old.SwapchainImages {
+      if !(old.ImagesAcquired[i]) {
+        delete(Images, img.VulkanHandle)
+      }
+    }
+  }
+
   queueFamilyIndices := create_info.pQueueFamilyIndices[0:create_info.queueFamilyIndexCount]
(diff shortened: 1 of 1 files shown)


matled (commented):
LGTM

Thanks for the detailed PR description and comments!

Not approving as I don't have much context.

gapis/api/vulkan/extensions/khr_swapchain.api
```diff
@@ -193,7 +205,7 @@ sub bool imageUsedByAnotherSwapchain(
     for _, _, s in Swapchains {
       if (s != null) && (s != swapchainObject) {
         for _, _, i in s.SwapchainImages {
-          if i == imageObject {
+          if i.VulkanHandle == imageObject.VulkanHandle {
```
yalcinmelihyasin, line 208: What is the reason for this change?

yalcinmelihyasin (approved):
LGTM

Only one minor question.

gapis/api/vulkan/extensions/khr_swapchain.api
hevrard: The code used to do a "deep equality" test, whereas we only want to check whether the handles are the same: this is particularly important as if an image handle from a retired swapchain image has been "recycled" for another swapchain image, then the image objects may have the same Vulkan handle, but be different otherwise.

yalcinmelihyasin: I see. Makes sense.

ianelliottus (approved):
I'm not familiar with this code, but I've tried to understand what it's doing and it LGTM

PR #834, opened 2021-06-17
Fix display of representations for glClear commands (#828)

### LINKED ISSUE #828: Correct thumbnail for low-level glClear*'s
**Bug description**

After changing ANGLE to create a glDraw*-like hierarchy for glClear*, I observed that the thumbnail behavior isn't correct.  It seems like it needs the same sort of hierarchy fix that Mark did for glDraw*.

mark-lunarg:
In an ANGLE AGI trace, the thumbnail/resources of the containing OpenGL group and the OpenGL glClear displayed incorrect info.
Created aliased representations for them to the correct thumbnail/resource data.

Test: Manual

Fixes #828.

### DIFF
--- a/gapis/resolve/command_tree.go
+++ b/gapis/resolve/command_tree.go
@@ -580,7 +580,7 @@ func getOpenGLAliasRepresentation(indices []uint64, item *api.CmdIDGroup, cmdTre
 		pItem, _ := cmdTree.index(parentIndices)
 		parentItem := pItem.(api.CmdIDGroup)
 		aliasRepId = parentItem.Range.Last()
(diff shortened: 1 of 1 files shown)


mark-lunarg:
This change was tested with depth/stencil clears, but still looking for an app that clears a color attachment.

ianelliottus:
I think all of the apps that I've seen that do glClear's for color buffers do them at the start of the render pass.

I'm off to a Dr. appointment.  I may not look at this until tomorrow.  Probably best to get Hugues to review this (since he knows the code base better).

ianelliottus (approved):
LGTM, thanks!

PR #837, opened 2021-06-21
Verify auth token for streaming RPCs

matled:
Also, use the auth token in the integration test service_test.go.
TestProfile is using a streaming RPCs and fails without adding the
StreamClientInterceptor to the client in the test.

Bug: 191146135

### DIFF
--- a/core/app/auth/auth.go
+++ b/core/app/auth/auth.go
@@ -111,7 +111,7 @@ func UnaryServerInterceptor(token Token) grpc.UnaryServerInterceptor {
 	}
 }
 
-// StreamServerInterceptor returns a grpc.UnaryServerInterceptor that checks
+// StreamServerInterceptor returns a grpc.StreamServerInterceptor that checks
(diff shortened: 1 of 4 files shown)


core/app/auth/auth.go
```diff
@@ -111,9 +111,29 @@ func ServerInterceptor(token Token) grpc.UnaryServerInterceptor {
 	}
 }
 
-// ClientInterceptor returns a grpc.UnaryClientInterceptor that adds the given
-// auth token to outgoing RPC calls.
-func ClientInterceptor(token Token) grpc.UnaryClientInterceptor {
+// StreamServerInterceptor returns a grpc.UnaryServerInterceptor that checks
```
hevrard, line 114: Typo in comment: `grpc.UnaryServerInterceptor` -> `grpc.StreamServerInterceptor`

hevrard (approved):
LGTM with just a typo to fix

PR #838, opened 2021-06-22
Improve ANGLE installation UI

johnplate:
- Disable Ok button in Tracer dialog for ANGLE traces if ANGLE is not detected.
- Remove cached device from server after installing apk, so that ANGLE can be detected when refreshing devices.
- Changed ANGLE install/update UI to buttons, see screen shots below.

This fixes b/191636142 AGI does not force installing ANGLE when using new device for first time
and b/183425312 Error trying frame trace after AGI installs ANGLE

Install ANGLE
Update ANGLE

### DIFF
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -1515,7 +1515,7 @@ public class TracerDialog {
     private static enum Status {
       DOWNLOADING("Downloading APK..."),
       INSTALLING("Installing APK..."),
-      DONE("All done."),
+      DONE("All done. AGI restart required."),
       FAILED("Download/Install failed.");
 
       public final String message;
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -520,7 +520,7 @@ public class TracerDialog {
         traceTarget.addBoxListener(SWT.Modify, e -> updateEmptyAppWithRenderStageWarning(models.settings));
 
         RowLayout angleBarLayout = new RowLayout();
-        angleBarLayout.fill = true;
+        angleBarLayout.center = true;
         angleBarLayout.spacing = 10;
         angleBar = withLayoutData(
             createComposite(this, angleBarLayout),
@@ -757,10 +757,10 @@ public class TracerDialog {
           if (version == 0 || version < settings.preferences().getLatestAngleRelease().getVersion())
           {
             if (version == 0) {
-              angleMessage.setText("ANGLE was not found on this device, but is required for selected trace type:");
+              angleMessage.setText("ANGLE is required for this trace type (AGI restart required):");
               installAngle.setText("Install ANGLE");
             } else {
-              angleMessage.setText("A new recommended ANGLE version is available for this device:");
+              angleMessage.setText("A recommended update to ANGLE is available (AGI restart required):");
               installAngle.setText("Update ANGLE");
             }
             angleMessage.requestLayout();
@@ -920,7 +920,11 @@ public class TracerDialog {
 
         if (InstallAngleDialog
             .showDialogAndInstallApk(getShell(), models, theme, dev.device, url) == Window.OK) {
-          deviceLoader.notifyListeners(SWT.MouseDown, new Event());
+          Composite comp = this;
+          while (comp.getParent() != null) {
+            comp = comp.getParent();
+          }
+          comp.getShell().close();
         }
       }
 
@@ -1429,6 +1433,10 @@ public class TracerDialog {
     @Override
     public void create() {
       super.create();
+      Button ok = getButton(IDialogConstants.OK_ID);
+      if (ok != null) {
+        ok.setText("Close AGI");
+      }
       updateButtons();
     }
 
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -87,6 +87,7 @@ import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.layout.RowLayout;
 import org.eclipse.swt.program.Program;
 import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Combo;
@@ -331,7 +332,9 @@ public class TracerDialog {
       private final Label fileLabel;
       private final Label requiredFieldMessage;
       private final Label emptyAppWarning;
-      private final Link newAngleAvailableMessage;
+      private final Composite angleBar;
+      private final Label angleMessage;
+      private final Button installAngle;
 
       protected String friendlyName = "";
       protected boolean userHasChangedOutputFile = false;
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -903,7 +920,7 @@ private void downloadAndInstallAngle(Theme theme, Settings settings) {
 
         if (InstallAngleDialog
             .showDialogAndInstallApk(getShell(), models, theme, dev.device, url) == Window.OK) {
-          newAngleAvailableMessage.setVisible(false);
+          deviceLoader.notifyListeners(SWT.MouseDown, new Event());
```
stellama0208, line 923: Hey John, I think the race condition sneaks in here. 
It's basically because we allow users to start tracing before device got actually refreshed. I could reproduce the gapir problem if I am trying to update ANGLE and I click "start tracing" very fast right after I close the install window.

By the current approach, the timeline looks like: 1. ANGLE got installed -> 2. user close ANGLE installing window -> 3. at this line, listener notifies `deviceLoader` to run -> 4. idle wait of 300ms(see Line365 of this file) -> 5. actually run `refreshDevices` runnable which also takes quite some time. If the user start tracing before step 5 finishes, the problem will occur. 

I think to solve it, we should mandate the `refreshDevices` to be the last step of the ANGLE installation process, so that everything is done before users closing the ANGLE installing window. To do that, I think we should basically modify `InstallAngleDialog` class: adding one more status, and pass in the `refreshDevices` runnable to be executed as the final step.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -742,16 +750,25 @@ private void updateOnApiChange(
...
-          if (dev.device.getConfiguration().getAngle().getVersion() == 0 ||
-              dev.device.getConfiguration().getAngle().getVersion() <
-              settings.preferences().getLatestAngleRelease().getVersion()) {
-            showAngleWarning = true;
+          int version = dev.device.getConfiguration().getAngle().getVersion();
+          if (version == 0 || version < settings.preferences().getLatestAngleRelease().getVersion())
+          {
+            if (version == 0) {
+              angleMessage.setText("ANGLE was not found on this device, but is required for selected trace type:");
+              installAngle.setText("Install ANGLE");
```
stellama0208, line 761: It would be great if we could keep the original color change, maybe yellow for ANGLE update and red for ANGLE install?

stellama0208 (commented):
Looks mostly good and verified the user journey flow by manual testing. After we fix the race condition I think the PR is good.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -498,20 +501,34 @@ protected void configureDialog(DirectoryDialog dialog) {
...
-        adbWarning.setVisible(!models.settings.isAdbValid());
+        emptyAppWarning.setForeground(getDisplay().getSystemColor(SWT.COLOR_DARK_YELLOW));
+        emptyAppWarning.setVisible(false);
+        device.getCombo().addListener(SWT.Selection, e -> updateEmptyAppWithRenderStageWarning(models.settings));
+        api.getCombo().addListener(SWT.Selection, e -> updateEmptyAppWithRenderStageWarning(models.settings));
+        traceTarget.addBoxListener(SWT.Modify, e -> updateEmptyAppWithRenderStageWarning(models.settings));
 
-        newAngleAvailableMessage = createLink(this, "A new ANGLE version is available for this " +
-            "device. Click to <a>download and install</a> the APK.",
+        RowLayout angleBarLayout = new RowLayout();
```
matled, line 522: Maybe the text message would be better placed above the button instead of next to it? I think this is a bit more common (I haven't found a good reference/guideline for this kind of UI design yet, just based on my intuition and a quick Google image search).

If it is next to it, it should probably also aligned in the middle vertically.

matled (commented):
Thanks for improving this!

gapic/src/main/com/google/gapid/views/TracerDialog.java
johnplate: Hi Stella, thank you for the review!

Actually after installing ANGLE the user is not able to start tracing before the device got refreshed, because the OK button will only be enabled if it detects ANGLE in the currently selected device, which can only happen after the refresh is complete. Also the trace would fail with "ANGLE not detected" if it would be run with the old device handle.

There must be another copy of the old device handle somewhere, which is used for opening the trace. This fails even when you don't take a new trace. It is enough to open an existing trace after installing ANGLE and cancelling the trace dialog.

gapic/src/main/com/google/gapid/views/TracerDialog.java
johnplate: Yes, Pau had the same concern that the message is less highlighted with the previous color, so we should keep the regular color for now until the theme issues are fixed.

gapic/src/main/com/google/gapid/views/TracerDialog.java
johnplate: We have the convention of allocating space for every hidden UI element, which is why there is already a lot of typically empty vertical space above the dialog button bar. This looks a bit weird already, and I don't want to keep it at a minimum, so I prefer the horizontal placemant. I will try to align it in the middle.

gapic/src/main/com/google/gapid/views/TracerDialog.java
johnplate: I was not able to solve the issue and was even able to reproduce it with enabled logging. Considering [Pascal's comment](https://github.com/google/agi/pull/799#discussion_r640063420) on Hughes identical solution I have reverted it and force the user instead after installing ANGLE to close and restart AGI.

gapic/src/main/com/google/gapid/views/TracerDialog.java
johnplate: The alignment is fixed. Thanks for paying attention to the details!

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -742,16 +750,25 @@ private void updateOnApiChange(
...
-              settings.preferences().getLatestAngleRelease().getVersion()) {
-            showAngleWarning = true;
+          int version = dev.device.getConfiguration().getAngle().getVersion();
+          if (version == 0 || version < settings.preferences().getLatestAngleRelease().getVersion())
+          {
+            if (version == 0) {
+              angleMessage.setText("ANGLE was not found on this device, but is required for selected trace type:");
+              installAngle.setText("Install ANGLE");
+            } else {
+              angleMessage.setText("A new recommended ANGLE version is available for this device:");
```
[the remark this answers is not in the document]
johnplate: I have updated the text, but kept "recommended" to implicitely show that we want users to update before they file bugs. :) Also added restart warning.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -1411,6 +1433,10 @@ protected Control createDialogArea(Composite parent) {
     @Override
     public void create() {
       super.create();
+      Button ok = getButton(IDialogConstants.OK_ID);
+      if (ok != null) {
+        ok.setText("Close AGI");
```
stellama0208, line 1438: I am a bit afraid of the current "restart" requirement is not noticeable and mandatory to users enough. Technically, if a user doesn't notice, they could close this dialog by the "X" icon at the upper right/left corner and continue without restarting. If we can't control the "X" icon, maybe at least we should move the "restart required for ANGLE installation/update" string to this dialog as well.

gapic/src/main/com/google/gapid/views/TracerDialog.java
stellama0208: Considering that we change to texts to warning colors when other requirements are not fulfilled(Not all field finished; invalid adb; empty app), I am a bit concerned that users will find it inconsistent and think this is not something mandatory(although the continue button will get disabled).  
To make sure the texts are readable, we could 1. mandate foreground color and background color together, and 2. define the color ourselves rather than get the system default color. But I am fine that the theme thing is done in another PR, just that we should remember to do it and not let it get lost.

gapic/src/main/com/google/gapid/views/TracerDialog.java
johnplate: Actually even if they close the install dialog, the OK button of the trace dialog would still be disabled, because AGI doesn't notice the new installation until the server is restarted or the device is un- & re-plugged.

I have now added a close event handler, which always returns ok after successful installation, doesn't matter how the dialog is closed. This also works with the escape key. It also prevents the shell from being closed during APK installation, which was also possible before even though it shouldn't.

gapic/src/main/com/google/gapid/views/TracerDialog.java
johnplate: You are right about inconsistent colors in general. In this case I don't see a grave inconsistency, since the angle message bar is a label and button, which is visually very different from the warning messages.

Fixing the theme problem is a bigger issue since SWT inherits OS constraints. That won't get lost.

stellama0208 (approved):

gapic/src/main/com/google/gapid/views/TracerDialog.java
stellama0208: Thanks for handling it! I think right now the flow guidance for the users is very clear!

PR #840, opened 2021-06-23
Create groups for vkCmd(Begin|End)Query

### LINKED ISSUE #826: Handle gl{Begin|End}Query like glClear with only 1 Vulkan command
**Bug description**
I recently changed ANGLE to treat glBeginQuery* and glEndQuery* like glDraw*, glClear*, and glDispatch*.  That is, they are no longer buried under other GLES commands, but are top-level commands.  They will typically have only one Vulkan command under them, vkCmd{Begin|End}Query.  This is similar to glClear*.  I'd like Mike to treat these single Vulkan commands like glClear*

mikes-lunarg:
Aligns with recent ANGLE change to split glBeginQuery and glEndQuery
into their own heirachy. Fixes #826.

Test: manual, on trace containing glBeginQuery/glEndQuery.

### DIFF
--- a/gapis/api/vulkan/vulkan.go
+++ b/gapis/api/vulkan/vulkan.go
@@ -455,24 +455,26 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
 
 			// Handle draw commands grouping.
 			cmdName := cb.CommandReferences().Get(uint32(i)).Type().String()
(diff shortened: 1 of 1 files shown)


mikes-lunarg:
Tested against TOT ANGLE (https://github.com/google/angle/commit/15623ada6083f92c280333df810a8fc98d9bdede)
826_test

ianelliottus:
Awesome, thanks Mike!

hevrard (approved):

PR #841, opened 2021-06-25
Use ANDROID_SDK_ROOT rather than ANDROID_SDK_HOME

hevrard:
ANDROID_SDK_ROOT is the recommended environment variable for defining
the path to the SDK, see
https://developer.android.com/studio/command-line/variables

Bug: b/192053630

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Settings.java
+++ b/gapic/src/main/com/google/gapid/models/Settings.java
@@ -345,7 +345,7 @@ public class Settings {
       return;
     }
 
-    String[] sdkVars = { "ANDROID_HOME", "ANDROID_SDK_HOME", "ANDROID_ROOT", "ANDROID_SDK_ROOT" };
+    String[] sdkVars = { "ANDROID_HOME", "ANDROID_ROOT", "ANDROID_SDK_ROOT" };
     for (String sdkVar : sdkVars) {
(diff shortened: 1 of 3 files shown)


matled (approved):

gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -345,7 +345,7 @@ private static void tryFindAdb(SettingsProto.Preferences.Builder current) {
       return;
     }
 
-    String[] sdkVars = { "ANDROID_HOME", "ANDROID_SDK_HOME", "ANDROID_ROOT", "ANDROID_SDK_ROOT" };
```
pmuetschard, line 348: Why did you remove it from here? This is not an improvement. This is the list of places to search for adb, so removing a location helps noone.

hevrard: I removed it because the Android documentation at https://developer.android.com/studio/command-line/variables says:

> Note: Starting with Android Studio 4.2, the ANDROID_SDK_HOME environment variable is deprecated and has been replaced with ANDROID_PREFS_ROOT.

Even before Android studio 4.2, `ANDROID_SDK_HOME` seemed to only be related to the Emulator, I don't think we could expect `adb` to be located there.

Is there a good reason to look for `adb` in this location?

PR #843, opened 2021-06-25
Resize freezing investigation

stellama0208:
This PR is only opened for reference, to explain and support some investigation result used in another real PR, never mind! Basically: 
- the first commit disables scroll bar and all related logic in AGI, the bug disappears.
- the second commit has only one line, which simply enables scroll bar config, the bug re-occurs.
- With these two observations, it looks like even simply having the scroll bar enabled, and not doing anything with it, is already enough to cause a problem.


stellama0208:
Close since this PR is only for reference.

PR #844, opened 2021-06-25
Temporarily disable ImageComponent's scroll bar to avoid a freeze bug on macOS.

stellama0208:
- Affected Panels: FrameBuffer and Texture(singular).
- Disabled Function: the scroll bar was originally used to quickly jump
to another part of the image, if the image got zoomed too big. After this
function gets disabled, users could still drag to navigate around the image,
and scroll to zoom in/out of the image.
- The original bug only happens on macOS platform, it's likely that there's
some compatibility issue between SWT and macOS Big Sur that causes it. More info
in PR #843.
- Weighing between the small impact of the disabled fuction, and the big possible
bad freezing experience for AGI mac users, this PR decides to hide the scroll bar
temporarily for the coming AGI frame profiler launch.

Bug: b/189382432.
Test: Manual test with AGI on macOS, have the FrameBuffer and Texture view
shown together, try resize the window, AGI client will freeze before this PR.

### DIFF
--- a/gapic/src/main/com/google/gapid/widgets/ImagePanel.java
+++ b/gapic/src/main/com/google/gapid/widgets/ImagePanel.java
@@ -696,7 +696,6 @@ public class ImagePanel extends Composite implements Loadable {
     private final Consumer<AlphaWarning> showAlphaWarning;
     private final boolean naturallyFlipped;
 
-    private final ScrollBar scrollbars[];
     private final ScenePanel<SceneData> canvas;
     protected final SceneData data;
     private Image[] images = {};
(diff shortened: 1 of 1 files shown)


hevrard (approved):

johnplate (approved):

pmuetschard (commented):
Since this is a Mac only issue, can you please only disable it on Mac? (`OS.isMac`)

stellama0208:
> Since this is a Mac only issue, can you please only disable it on Mac? (`OS.isMac`)

Good point! Just did it with a new commit.

pmuetschard (approved):

PR #848, opened 2021-06-30
Fuchsia device discovery

rosasco-wk:
- Fuchsia device discovery and test.
- file.Path.IsExecutable() func and test.

### DIFF
--- a/core/os/file/path.go
+++ b/core/os/file/path.go
@@ -17,12 +17,12 @@ package file
 import (
 	"encoding/json"
 	"io/ioutil"
-	"log"
 	"net/url"
 	"os"
 	"os/exec"
 	"os/user"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 )
@@ -92,17 +92,11 @@ func (p Path) IsDir() bool {
 // IsExecutable returns true if the file at p.value exists, is executable
 // and is not a directory.
 func (p Path) IsExecutable() bool {
-	if !p.IsDir() {
-		fileInfo, err := os.Stat(p.value)
-		if err == nil {
-			bits := fileInfo.Mode().Perm()
-			return (bits & 0111) != 0
-		} else {
-			log.Print(err)
-			return false
-		}
+	if runtime.GOOS == "windows" {
+		return true
 	}
-	return false
+	info := p.Info()
+	return info != nil && !info.IsDir() && (info.Mode().Perm()&0111) != 0
 }
 
 // System returns the full absolute path using the system separator.
--- a/core/os/file/path.go
+++ b/core/os/file/path.go
@@ -17,6 +17,7 @@ package file
 import (
 	"encoding/json"
 	"io/ioutil"
+	"log"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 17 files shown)


hevrard:
@rosasco-wk the CI is failing, you can start by fixing CI-presubmit by testing it locally with:

```
./kokoro/presubmit/presubmit.sh
```

core/os/file/path.go
```diff
@@ -88,6 +89,22 @@ func (p Path) IsDir() bool {
 	return info != nil && info.IsDir()
 }
 
+// IsExecutable returns true if the file at p.value exists, is executable
+// and is not a directory.
+func (p Path) IsExecutable() bool {
```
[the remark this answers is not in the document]
pmuetschard: To close the loop, I ran this:
```
info, _ := os.Stat("c:\\dev\\go\\bin\\go.exe")
fmt.Println(info != nil && !info.IsDir() && (info.Mode().Perm() & 0111) != 0)
fmt.Println(info.Mode().Perm())
```
And it prints:
```
false
-rw-rw-rw-
```

So we should add `if runtime.GOOS == "windows" { return true }` to the top.

core/os/file/path.go
```diff
@@ -88,6 +89,16 @@ func (p Path) IsDir() bool {
...
+	if runtime.GOOS == "windows" {
+		return true
+	}
+	info := p.Info()
+	return info != nil && !info.IsDir() && (info.Mode().Perm()&0111) != 0
```
pmuetschard, line 99: I was going to ask you to update the doc comment to say that on Windows it always returns true, but then realized we should be a bit better and do this:

```suggestion
	info := p.Info()
	if info == nil || info.IsDir() {
		return false
	}
	return runtime.GOOS == "windows" || (info.Mode().Perm()&0111) != 0
```

Sorry for the back-and-forth. Please do update the comment to indicate that on Windows all files are considered executable.

core/os/file/path_test.go
```diff
@@ -50,3 +53,32 @@ func TestPathContains(t *testing.T) {
...
 }
+
+func TestIsExecutable(t *testing.T) {
+	ctx := log.Testing(t)
+	var (
+		tmpdir  string
+		err     error
+		tmpfile *os.File
+	)
+	tmpdir, err = ioutil.TempDir("/tmp", "path_test-")
```
pmuetschard, line 64: use `""` instead of `"/tmp"`, so that `ioutil.TempDir` uses the default temporary directory.

core/os/file/path_test.go
```diff
@@ -50,3 +53,32 @@ func TestPathContains(t *testing.T) {
...
+
+	path := file.Abs(tmpdir)
+	assert.For(ctx, "Temp directory").ThatBoolean(path.IsExecutable()).IsFalse()
+
+	tmpfile, err = ioutil.TempFile(tmpdir, "path_test_file-")
+	tmpfile.Close()
+
+	err = os.Chmod(tmpfile.Name(), 0600)
+	assert.For(ctx, "Chmod temp file").ThatError(err).Succeeded()
+	assert.For(ctx, "Chmod temp file").ThatBoolean(path.IsExecutable()).IsFalse()
```
pmuetschard, line 75: This needs to assert `IsTrue` on Windows.

pmuetschard (approved):

PR #849, opened 2021-06-30
Fixes missing device id in Perfetto data.

yalcinmelihyasin:
Sometimes Perfetto returns deviceID 0 for the QC devices which leads
to missing data.

There is no guaranteed way to fix this but for the Android we can assume
that, if there is only one device id is used for the entire trace,
it's very likely that missing deviceID is the same one rather than a
different device. If there are multiple devices then it does not do anything.

### DIFF
--- a/gapis/trace/android/adreno/profiling_data.go
+++ b/gapis/trace/android/adreno/profiling_data.go
@@ -98,7 +98,7 @@ func fixContextIds(contextIDs []int64) {
 	// device id, where there is only one device id.
 
 	zeroIndices := make([]int, 0)
-	deviceID := int64(0)
+	contextID := int64(0)
 
 	for i, v := range contextIDs {
 		if v == 0 {
@@ -106,12 +106,12 @@ func fixContextIds(contextIDs []int64) {
 			continue
 		}
 
-		if deviceID == 0 {
-			deviceID = v
+		if contextID == 0 {
+			contextID = v
 			continue
 		}
 
-		if deviceID != v {
+		if contextID != v {
 			// There are multiple devices
 			// We cannot know which one to fill
 			return
@@ -121,7 +121,7 @@ func fixContextIds(contextIDs []int64) {
 	for _, v := range zeroIndices {
 		// If there is only one device in entire trace
 		// We can assume that we possibly have only one device
-		contextIDs[v] = deviceID
+		contextIDs[v] = contextID
 	}
 }
 
--- a/gapis/trace/android/adreno/profiling_data.go
+++ b/gapis/trace/android/adreno/profiling_data.go
@@ -91,13 +91,15 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
 }
 
 func fixContextIds(contextIDs []int64) []int64 {
-	// This is a temporary work to workaround a QC bug
+	// This is a workaround a QC bug(b/192546534)
 	// that causes first deviceID to be zero after a
 	// renderpass change in the same queue submit.
+	// So, we fill the zero devices with the existing
+	// device id, where there is only one device id.
 
 	zeroIndices := make([]int, 0)
 	deviceID := int64(0)
-	multipleDevice := false
+	multipleDevices := false
 
 	for i, v := range contextIDs {
 		if v == 0 {
@@ -109,13 +111,12 @@ func fixContextIds(contextIDs []int64) []int64 {
 			if deviceID == 0 {
 				deviceID = v
 				continue
-			} else {
-				multipleDevice = true
 			}
+			multipleDevices = true
 		}
 	}
 
-	if multipleDevice {
+	if multipleDevices {
 		// We cannot know which deviceID we should fill
 		return contextIDs
 	}
--- a/gapis/trace/android/adreno/profiling_data.go
+++ b/gapis/trace/android/adreno/profiling_data.go
@@ -90,6 +90,45 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
 	}
 }
 
+func fixContextIds(contextIDs []int64) []int64 {
+	// This is a temporary work to workaround a QC bug
+	// that causes first deviceID to be zero after a
+	// renderpass change in the same queue submit.
+
+	zeroIndices := make([]int, 0)
+	deviceID := int64(0)
+	multipleDevice := false
+
+	for i, v := range contextIDs {
+		if v == 0 {
+			zeroIndices = append(zeroIndices, i)
+			continue
+		}
+
+		if deviceID != v {
+			if deviceID == 0 {
+				deviceID = v
+				continue
+			} else {
+				multipleDevice = true
+			}
+		}
+	}
+
+	if multipleDevice {
+		// We cannot know which deviceID we should fill
+		return contextIDs
+	}
+
+	for _, v := range zeroIndices {
+		// If there is only one device in entire trace
+		// We can assume that we possibly have only one device
+		contextIDs[v] = deviceID
+	}
+
+	return contextIDs
+}
+
 func processGpuSlices(ctx context.Context, processor *perfetto.Processor, capture *path.Capture, handleMapping *map[uint64][]service.VulkanHandleMappingItem, syncData *sync.Data) (*service.ProfilingData_GpuSlices, error) {
 	slicesQueryResult, err := processor.Query(slicesQuery)
 	if err != nil {
@@ -120,6 +159,7 @@ func processGpuSlices(ctx context.Context, processor *perfetto.Processor, captur
 	// Grab all the column values. Depends on the order of columns selected in slicesQuery
 
 	contextIds := slicesColumns[0].GetLongValues()
+	contextIds = fixContextIds(contextIds)
 	extractTraceHandles(ctx, &contextIds, "VkDevice", handleMapping)
 
 	renderTargets := slicesColumns[1].GetLongValues()
--- a/gapis/trace/android/adreno/profiling_data.go
+++ b/gapis/trace/android/adreno/profiling_data.go
@@ -90,7 +90,7 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
 	}
 }
 
-func fixContextIds(contextIDs []int64) []int64 {
+func fixContextIds(contextIDs []int64) {
 	// This is a workaround a QC bug(b/192546534)
 	// that causes first deviceID to be zero after a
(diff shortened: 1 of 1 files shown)


gapis/trace/android/adreno/profiling_data.go
```diff
@@ -90,6 +90,45 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
...
+	for i, v := range contextIDs {
+		if v == 0 {
+			zeroIndices = append(zeroIndices, i)
+			continue
+		}
+
+		if deviceID != v {
+			if deviceID == 0 {
+				deviceID = v
+				continue
```
hevrard, line 111: I think you can remove this `continue`

gapis/trace/android/adreno/profiling_data.go
```diff
@@ -90,6 +90,45 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
 	}
 }
 
+func fixContextIds(contextIDs []int64) []int64 {
+	// This is a temporary work to workaround a QC bug
```
matled, line 94: If this is temporary, we should put some context here to make it easy to remove when appropriate. Do we have a tracking bug for the underlying fix? Is there any kind of timeline for when this would be fixed? I assume this would still be an issue on older/not updated devices for a while? How would we validate that the fix is not necessary anymore? 

The answers should not necessarily be in the comment here, but probably a link to a resource that has more of this context to follow up.

hevrard (approved):

gapis/trace/android/adreno/profiling_data.go
```diff
@@ -90,6 +90,45 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
...
+	// that causes first deviceID to be zero after a
+	// renderpass change in the same queue submit.
+
+	zeroIndices := make([]int, 0)
+	deviceID := int64(0)
+	multipleDevice := false
```
matled, line 100: nit: multipleDevices

gapis/trace/android/adreno/profiling_data.go
```diff
@@ -90,6 +90,45 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
 	}
 }
 
+func fixContextIds(contextIDs []int64) []int64 {
+	// This is a temporary work to workaround a QC bug
+	// that causes first deviceID to be zero after a
+	// renderpass change in the same queue submit.
+
+	zeroIndices := make([]int, 0)
```
matled, line 98: Using a separate list of indices here looks like it may be more complicated than necessary?

Did you consider just doing two full passes:
1. Detect if there are multiple non-zero ids.
2. If there is only one non-zero id, just do a full second pass.

Depending on the size of the input, this may not even be less efficient. (Re)allocating the zeroIndices also has some cost. (No idea if performance is a concern here.)

gapis/trace/android/adreno/profiling_data.go
```diff
@@ -90,6 +90,45 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
 	}
 }
 
+func fixContextIds(contextIDs []int64) []int64 {
```
matled, line 93: Could you add a function comment on what this is intended to do?

For example something like this:
// fixContextIds applies heuristics to fill in missing context ids.
// This is to work around a bug in certain devices (QC, see b/...) reporting a device id of 0.
// If only one non-zero device id is found, all zero device ids are replaced by that id.
// Otherwise the contextIds are left unchanged.

johnplate (commented):
First some questions:

- I'm a bit confused about the variable names and the content. Do we have context IDs or device IDs?
- Is it possible at all and supported by AGI that there are multiple devices in contextIDs?

 I propose the following improvements:

- Do we have a cached info, if the device is QC? If yes, please only call the function for QC, since it is a workaround for QC.
- It would be more effecient to return immediately when you find a second device ID, instead of finishing the first loop.
- The function signature intuitively suggests that the passed array is not modified and a new array is returned. I would not return anything from a function which modifies the passed array in-place.

hevrard (commented):
Looks like there are lots of relevant comments to address.

hevrard:
As an aside: the CI-linux-test failure is due to a flake, this should be fixed by https://github.com/google/agi/pull/852

gapis/trace/android/adreno/profiling_data.go
yalcinmelihyasin: Performance is not a concern here. It's all under 100 entry and all of them are probably in cache anyway.

And it does pretty much the same thing. I don't see a reason to change.

gapis/trace/android/adreno/profiling_data.go
yalcinmelihyasin: Yes that's true, I deleted the else instead.

gapis/trace/android/adreno/profiling_data.go
yalcinmelihyasin: 👍

gapis/trace/android/adreno/profiling_data.go
yalcinmelihyasin: I have merged the two comment about comments and answering them here:

I generally do not favor explaining the code. It either leads to outdated comment, or extra maintenance of the comment. I think code itself should be, and in this case I believe it is, explaining itself.

Here however there is an extra bit of it that makes people aware the bug and the logic to workaround it. So I added the information about why and how we do the workaround.

yalcinmelihyasin: And for being a bit more honest. I removed the "temporary" from the comment. Realistically speaking, we will never remove this if this is not a perfetto but a QC bug because there will always be devices with old drivers.

yalcinmelihyasin:
- ContextID is deviceID. I know it's confusing and I don't know why it has named like that. But if you look at the code, in perfetto table it's a contextId but when we map it from replay to trace, it's handle type for contextID is VkDevice.

- So, in theory, there could be an application that does, two different computation with two different VkDevices and share their result with each other. In practice, however, I don't think that it would ever happen for an Android Application. Maybe I was a bit too defensive there. Let me know, if you think that removing it is better idea.

- The entire file runs only for QC devices. There is another of this file for Mali. Which is a problem itself because it's mostly a code duplication. I don't know why it happened like that. I think we should refactor and they share the same common code and only do the different bits in other files. But that's a low priority.

- That's true. I will do that.
- That's also true. I think it's a leftover. I will fix it.

> First some questions:
> 
> * I'm a bit confused about the variable names and the content. Do we have context IDs or device IDs?
> * Is it possible at all and supported by AGI that there are multiple devices in contextIDs?
> 
> I propose the following improvements:
> 
> * Do we have a cached info, if the device is QC? If yes, please only call the function for QC, since it is a workaround for QC.
> * It would be more effecient to return immediately when you find a second device ID, instead of finishing the first loop.
> * The function signature intuitively suggests that the passed array is not modified and a new array is returned. I would not return anything from a function which modifies the passed array in-place.

gapis/trace/android/adreno/profiling_data.go
johnplate: I considered to mention the same as Matthias, since I am pretty sure that two full passes are still cheaper than the array allocation(s), but then decided not to mention it because I agree that performance is not a concern here.

johnplate (approved):

gapis/trace/android/adreno/profiling_data.go
```diff
@@ -90,6 +90,41 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
...
+			// There are multiple devices
+			// We cannot know which one to fill
+			return
+		}
+	}
+
+	for _, v := range zeroIndices {
+		// If there is only one device in entire trace
+		// We can assume that we possibly have only one device
+		contextIDs[v] = deviceID
```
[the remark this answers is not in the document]
yalcinmelihyasin: Because we save the indices of zero values. So it's an array let's say [1, 3, 7] which are the locations of zeros in contextIDs.

gapis/trace/android/adreno/profiling_data.go
yalcinmelihyasin: I was thinking explicitly marking the zero indices and fixing them later, actually explains what code is doing better than just passing on entire array in the second run but I am open to suggestions.

Because people in the future may ask, why do we run through the entire array and set all elements to the deviceID which is already there. This way, I think it's more clear and explains itself that, we specifically search for zeros, and if there is only one deviceID, we fill them with that handle.

gapis/trace/android/adreno/profiling_data.go
```diff
@@ -90,6 +90,41 @@ func extractTraceHandles(ctx context.Context, replayHandles *[]int64, replayHand
...
 
+func fixContextIds(contextIDs []int64) {
+	// This is a workaround a QC bug(b/192546534)
+	// that causes first deviceID to be zero after a
+	// renderpass change in the same queue submit.
+	// So, we fill the zero devices with the existing
+	// device id, where there is only one device id.
+
+	zeroIndices := make([]int, 0)
+	deviceID := int64(0)
```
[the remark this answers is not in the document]
yalcinmelihyasin: 👍

gapis/trace/android/adreno/profiling_data.go
yalcinmelihyasin: I have added more comment and I think it should be fine with already existing comments on other lines.

gapis/trace/android/adreno/profiling_data.go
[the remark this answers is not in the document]
johnplate: Yep, looks like misunderstanding from Matthias. Had to read the code twice to understand how zeroIndices works.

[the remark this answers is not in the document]
matled: Ah, got it. Confused myself by thinking about the two possible implementations and having to look up how `for` on a `range` works exactly in Go :)

gapis/trace/android/adreno/profiling_data.go
matled: The comment contains all the information. I'm very used to the typical function comment style and think there is value in consistency/making code easier to read that way. I.e. the suggestion is to replace the comment inside the function.

It's a bit of a style question, as I'm new to this and don't really understand if there is consistency or consensus on this, feel free to leave as-is.

matled (approved):
Thanks for fixing this issue!

johnplate (approved):

PR #850, opened 2021-06-30
Workaround the mapping bug due to duplicated Vulkan Handles.

yalcinmelihyasin:
This does not solve the duplication but minimises the risk of
duplication by not mapping the initial command handles for VkFramebuffer.

In Mapping Handler, there is no distinction between initialCmds and
Trace Commands, therefore lots of handles that are being created and
destroyed during the initialCmds also being mapped.

This PR, checks the VkFramebuffer handles and only maps them if they
are used by the application itself.

Bug: b/180532468

### DIFF
--- a/gapis/api/vulkan/transform_mapping_exporter.go
+++ b/gapis/api/vulkan/transform_mapping_exporter.go
@@ -43,20 +43,24 @@ type mappingHandle struct {
 }
 
 type mappingExporter struct {
-	mappings       *map[uint64][]service.VulkanHandleMappingItem
-	thread         uint64
-	path           string
-	traceValues    []mappingHandle
-	notificationID uint64
+	mappings         *map[uint64][]service.VulkanHandleMappingItem
+	thread           uint64
+	path             string
+	traceValues      []mappingHandle
+	notificationID   uint64
+	usedFrameBuffers map[uint64]bool
+	numOfInitialCmds uint64
 }
 
-func newMappingExporter(ctx context.Context, mappings *map[uint64][]service.VulkanHandleMappingItem) *mappingExporter {
+func newMappingExporter(ctx context.Context, numOfInitialCmds uint64, mappings *map[uint64][]service.VulkanHandleMappingItem) *mappingExporter {
 	return &mappingExporter{
-		mappings:       mappings,
-		thread:         0,
-		path:           "",
-		traceValues:    make([]mappingHandle, 0, 0),
-		notificationID: 0,
+		mappings:         mappings,
+		thread:           0,
+		path:             "",
+		traceValues:      make([]mappingHandle, 0, 0),
+		notificationID:   0,
+		usedFrameBuffers: make(map[uint64]bool),
+		numOfInitialCmds: numOfInitialCmds,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


gapis/api/vulkan/transform_mapping_exporter.go
```diff
@@ -105,9 +109,50 @@ func (mappingTransform *mappingExporter) TransformCommand(ctx context.Context, i
...
+			if err := mappingTransform.recordSubmittedRenderPass(ctx, queueSubmit, id.GetID(), inputState); err != nil {
+				return nil, err
+			}
+		}
+	}
+
 	return inputCommands, nil
 }
 
+func (mappingTransform *mappingExporter) recordSubmittedRenderPass(ctx context.Context, cmd *VkQueueSubmit, cmdID api.CmdID, inputState *api.GlobalState) error {
```
matled, line 123: The function is relatively long and boiler-plate-y. A short function comment would help me as a reader to quickly understand what it is supposed to do without having to parse and interpret the whole function body.

If I understand correctly, it just populates the `usedFramebuffers` map?

gapis/api/vulkan/transform_mapping_exporter.go
```diff
@@ -105,9 +109,50 @@ func (mappingTransform *mappingExporter) TransformCommand(ctx context.Context, i
...
+			return err
+		}
+
+		for _, cmdBuffer := range commandBuffers {
+			cmdBufferObj := GetState(inputState).CommandBuffers().Get(cmdBuffer)
+			for cmdIndex := 0; cmdIndex < cmdBufferObj.CommandReferences().Len(); cmdIndex++ {
+				currentCmd := cmdBufferObj.CommandReferences().Get(uint32(cmdIndex))
+				args := GetCommandArgs(ctx, currentCmd, stateObj)
+				if beginRenderPassArgs, ok := args.(VkCmdBeginRenderPassArgsʳ); ok {
+					if uint64(cmdID) > mappingTransform.numOfInitialCmds {
```
matled, line 145: I may be missing something here, but `cmdID` is an argument and `numOfInitialCmds` is not changed, i.e. this check could be done at the start of the function? I guess this should maybe be some other id?

gapis/api/vulkan/transform_mapping_exporter.go
```diff
@@ -178,9 +223,17 @@ func (mappingTransform *mappingExporter) processNotification(ctx context.Context
...
 		}
 
+		if handle.name == "VkFramebuffer" && !mappingTransform.usedFrameBuffers[handle.traceValue] {
+			continue
+		}
+
 		(*mappingTransform.mappings)[replayValue] = append(
 			(*mappingTransform.mappings)[replayValue],
-			service.VulkanHandleMappingItem{HandleType: handle.name, TraceValue: handle.traceValue, ReplayValue: replayValue},
+			service.VulkanHandleMappingItem{
```
matled, line 232: nit: This looks like a pure formatting change based on adding a trailing comma? I'd recommend to avoid such changes mixed with a functional change as it requires a lot of manual comparing to figure out if anything changed or if this is indeed just a formatting change.

gapis/api/vulkan/transform_mapping_exporter.go
yalcinmelihyasin: 👍 good catch

gapis/api/vulkan/transform_mapping_exporter.go
yalcinmelihyasin: This is a leftover, I was trying lots of stuff. I can revert this.

gapis/api/vulkan/transform_mapping_exporter.go
yalcinmelihyasin: Well, it literally records the renderpass that has been actually submitted to GPU. I think one thing to clear out is that it tracks the Framebuffer handle and not the Renderpass handle.

gapis/api/vulkan/transform_mapping_exporter.go
```diff
@@ -105,9 +109,55 @@ func (mappingTransform *mappingExporter) TransformCommand(ctx context.Context, i
...
+					return nil, err
+				}
+			}
+		}
+	}
+
 	return inputCommands, nil
 }
 
+// Record the framebuffers used in the renderpasses that actually submitted to GPU.
```
matled, line 129: Comment style: https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

I'd also add the detail on what it populates, with the boiler plate of the function it takes a bit of effort to figure out what the state change on the object actually is.

// recordFramebuffersInSubmittedRenderPasses marks framebuffers that are submitted to the GPU in `mappingExporter.usedFrameBuffers`.

gapis/api/vulkan/transform_mapping_exporter.go
```diff
@@ -105,9 +109,55 @@ func (mappingTransform *mappingExporter) TransformCommand(ctx context.Context, i
 		mappingTransform.thread = inputCommands[0].Thread()
 	}
 
+	// This is a workaround for b/180532468.
```
matled, line 112: I think it would be easier for the reader to get an idea what's going on here to rephrase this. Particularly, give a short bit of context what's going on and introduce things in order.

For example, from what I think is going on here, this may be easier to read:

// Vulkan allows duplicate handles which we don't track properly (b/180532468).
// As a partial workaround, keep track of the used handles outside of `initialCmds`
// to reduce the chance of mapping to the wrong object.

gapis/api/vulkan/transform_mapping_exporter.go
matled: Thanks. Still a slight change. More for the future :) I try to remember to do a `git diff` to spot these things to keep the diff as easy as possible to review.

matled (approved):

gapis/api/vulkan/transform_mapping_exporter.go
```diff
@@ -43,20 +43,24 @@ type mappingHandle struct {
...
-		path:           "",
-		traceValues:    make([]mappingHandle, 0, 0),
-		notificationID: 0,
+		mappings:         mappings,
+		thread:           0,
+		path:             "",
+		traceValues:      make([]mappingHandle, 0, 0),
+		notificationID:   0,
+		usedFrameBuffers: make(map[uint64]bool),
+		numOfInitialCmds: numOfInitialCmds,
```
pmuetschard, line 63:
```suggestion
		mappings:         mappings,
		usedFrameBuffers: map[uint64]struct{}{},
		numOfInitialCmds: numOfInitialCmds,
```

- don't init zero value fields
- use nil (zero value) for empty slices
- init maps using the `{}` syntax (for some reason that's what we do in our code base)

gapis/api/vulkan/transform_mapping_exporter.go
```diff
@@ -43,20 +43,24 @@ type mappingHandle struct {
...
-	thread         uint64
-	path           string
-	traceValues    []mappingHandle
-	notificationID uint64
+	mappings         *map[uint64][]service.VulkanHandleMappingItem
+	thread           uint64
+	path             string
+	traceValues      []mappingHandle
+	notificationID   uint64
+	usedFrameBuffers map[uint64]bool
```
pmuetschard, line 51: Sets in go are `map[<keytype>]struct[}`. Otherwise it's expected that the map may contain both `true` and `false` values.

PR #851, opened 2021-07-01
In perftab, apply a unified unit to each metric column.

stellama0208:
Before this PR, for some units that have different level of offset,
e.g. Time(ns, us, ms, s), Memory Size(B, KB, MB, GB), perftab will
allow cells in a metric column to have different offset, like 1.172ms
and 698us appearing together. That may make it difficult to compare
values in each column. This PR trys to unify unit's offset level for
each metric and solve the problem, the optimal offset level is picked
based on the average performance number.

Bug: b/190541223.
Test: Changes could be observed in Performance tab, on GPU Time metric
or Memory bandwidth related metrics.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/Unit.java
+++ b/gapic/src/main/com/google/gapid/perfetto/Unit.java
@@ -192,7 +192,8 @@ public class Unit {
   private static interface Formatter {
     public String format(long value);
     public String format(double value);
-    // Conversion interface for units that can convert between different offsets. e.g. Time, Memory.
+    // Conversion interface for units that can convert between different offsets/scales.
+    // e.g. Time can be converted among {ns, us, ms, s, m, h}, Memory among {B, KB, MB, GB}, etc.
     public default boolean convertible() { return false; }
     public default int findOptimalOffset(double value) { return -1; }
     public default String convertTo(double value, int offset) { return "Conversion not supported"; }
--- a/gapic/src/main/com/google/gapid/perfetto/Unit.java
+++ b/gapic/src/main/com/google/gapid/perfetto/Unit.java
@@ -111,15 +111,16 @@ public class Unit {
 
       @Override
       public Formatter withFixedScale(double representativeValue) {
+        Unit fixed = numer.withFixedScale(representativeValue);
         return new Formatter() {
           @Override
           public String format(long value) {
-            return numer.withFixedScale(representativeValue).format(value) + "/" + denom.name;
+            return fixed.format(value) + "/" + denom.name;
           }
 
           @Override
           public String format(double value) {
-            return numer.withFixedScale(representativeValue).format(value) + "/" + denom.name;
+            return fixed.format(value) + "/" + denom.name;
           }
         };
       }
--- a/gapic/src/main/com/google/gapid/perfetto/Unit.java
+++ b/gapic/src/main/com/google/gapid/perfetto/Unit.java
@@ -110,18 +110,18 @@ public class Unit {
       }
 
       @Override
-      public boolean convertible() {
-        return numer.formatter.convertible();
-      }
+      public Formatter withFixedScale(double representativeValue) {
+        return new Formatter() {
+          @Override
+          public String format(long value) {
+            return numer.withFixedScale(representativeValue).format(value) + "/" + denom.name;
+          }
 
-      @Override
-      public int findOptimalOffset(double value) {
-        return numer.formatter.findOptimalOffset(value);
-      }
-
-      @Override
-      public String convertTo(double value, int offset) {
-        return numer.formatter.convertTo(value, offset) + "/" + denom.name;
+          @Override
+          public String format(double value) {
+            return numer.withFixedScale(representativeValue).format(value) + "/" + denom.name;
+          }
+        };
       }
     });
   }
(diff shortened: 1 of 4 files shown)


gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -328,17 +329,25 @@ private void addColumnForMetric(Service.ProfilingData.GpuCounters.Metric metric)
...
+          String maxStr = perf.getMax() < 0 ? "?" : format(perf.getMax(), unit, optimalOffset);
           return minStr + " ~ " + maxStr;
         }
       }
     });
     column.getColumn().setAlignment(SWT.RIGHT);
   }
 
+  private String format(double value, Unit unit, int optimalOffset) {
+    if (unit.convertible() || optimalOffset >= 0) {
```
johnplate, line 344: In my understanding it needs a && operator here.

johnplate (changes requested):
Thanks, nice feature!

First I was confused about the word "offset", because it is defined as a distance to a displacement, until I understood that you meant the unit prefix and value scaling. I am not sure what is the best name to use, but for sure "offset" is not the right choice. I would use "scale", since you choose the best scaling factor until you have the same unit prefix for the column.

gapic/src/main/com/google/gapid/views/PerformanceView.java
stellama0208: Oops, thanks for catching it! Just fixed it with a new commit.

As for the term `offset`, it was chosen because it was already in quite some use in the unit file, for those `Time`, `Base2`, and `Base10` units. I think I will keep the term for now for consistence, but I modified the annotations a bit to make the term meaning more clear, thanks for suggestion!

hevrard (commented):
"Offset" vs "scale": I agree that "scale" would be a better wording.

Best would be to keep changes separate, with a PR to rename offset to scale, and another one to have all value with the same scale. As this PR is already in good shape, it's OK to have both changes in it I guess.

-----------

On another point, you're getting the optimal scale for the average of the values: I wonder if it would make sense to get it for the max value of the column? That would maybe help make the bottleneck stand out.

To take an extreme case, consider a column with these values: 1Ghz, 0Hz, 0Hz, 0Hz.
The average is 1Ghz / 4 == 250Mhz, so I think your current approach will show 1000Mhz, 0Mhz, 0Mhz, 0Mhz. Whereas if you take the optimal of the max, we'll see 1Ghz, 0Ghz, 0Ghz, 0Ghz. @stellama0208 @johnplate what do you think?

johnplate (changes requested):
I agree with Hughes that "scale" is a better wording here. Consistency is not very relevant here since the existing usage is only a few private members, which can be quickly (and should be) changed as well. I also agree that no separate PR is required since the existing PR is in good shape.

---

I was also wondering during my first review if the average value is a good fit to select the scale.  During my tests with several traces and also theoretical cases, I came to the opinion that average is indeed better than min or max, as the average will improve the readability for the majority of the values. Since it is a scale, we could consider to use geometric instead of arithmetic mean for the average, but I think there are not many cases where that actually would make a difference.

gapic/src/main/com/google/gapid/perfetto/Unit.java
```diff
@@ -156,13 +173,30 @@ public String format(double value) {
...
+  }
+
+  public int findOptimalOffset(double value) {
+    return formatter.findOptimalOffset(value);
+  }
+
+  public String convertTo(double value, int offset) {
+    return formatter.convertTo(value, offset);
+  }
+
```
pmuetschard, line 187: I think a better interface for this would be:

```
public Unit withFixedScale(double representativeValue) {
  return new Unit(name, formatter.withFixedScale(representativeValue));
}

private static interface Formatter {
  // ...
  public default Formatter withFixedScale(double representativeValue) { return this; }
}
```

For formatters with a scale, it can compute the offset and then return an anonymous `Formatter` implementation with the simplified formatting. You can then hide the internal implementation details (i.e. the "offset" discussion) inside the `Formatter` implementation.

pmuetschard:
To add to the scale vs offset discussion: it's called `offset`, because its a base index into an array. In other words, there is an array, which should now be accessed as `array[offset + i]` in place of `array[i]`.

`scale` is a bad name, because then it's implied that you should multiply by that amount, which is wrong. The closest could possibly be `exponent`, but that would imply that the base is constant, but that is not the case for the `Time` formatter, where it changes from 1000 to 60.

stellama0208:
Thanks for everyone's discussion and Pascal's suggestion, now with an updated interface, the exposed public interface/signature will use `scale`, and the `offset` will only show up in the implementation detail where `offset` is easy-to-reference, self-explanatory and relevant. Ready for another look!

gapic/src/main/com/google/gapid/perfetto/Unit.java
stellama0208: Thanks for the suggestion! Just updated with this change :)

johnplate (approved):
Thanks, good solution!

gapic/src/main/com/google/gapid/perfetto/Unit.java
```diff
@@ -106,6 +108,21 @@ public String format(long value) {
       public String format(double value) {
         return numer.format(value) + "/" + denom.name;
       }
+
+      @Override
+      public Formatter withFixedScale(double representativeValue) {
+        return new Formatter() {
+          @Override
+          public String format(long value) {
+            return numer.withFixedScale(representativeValue).format(value) + "/" + denom.name;
```
pmuetschard, line 117: Lift `numer.withFixedScale` out of the anonymous class, so you have to compute it only once. i.e. something like
```
Formatter fixed = numer.withFixedScale(representativeValue);
return new Formatter() {
  // ...
 return fixed.format(value) + "/" + denom.name;
 // ...
}
```

pmuetschard (approved):

gapic/src/main/com/google/gapid/perfetto/Unit.java
stellama0208: Sure, just did with a new commit.

PR #852, opened 2021-07-01
Remove irrelevant and flaky test WithoutSpinLockGuardedWrapper

hevrard:
This change removes the WrapperTest.WithoutSpinLockGuardedWrapper test
that was part of the memory tracker tests.

The first issue with this test is that it only use test framework
primitives, and thus doesn't test anything related to the memory
tracker.

The second issue is that the test expectations are incorrect, and this
leads to flaky failures on CI-linux-test. This test starts two
threads, each of them increments a shared counter. The first thread
sleeps on 5000us before incrementing the counter. The test seems to
assume that this sleep (and maybe the lock/unlock operations on m_)
are enough to guarantee that the second thread is the first to
increment the counter and is thus guaranteed to see the counter at
value 1 after that; while the first thread will always see the counter
at value 2 (thus assuming the second thread already had incremented it
before).

This is incorrect, it is possible that the first thread completes
before the second thread is scheduled at all. This is confirmed by
flakes on this test where the expectations are violated:

```
[ RUN      ] WrapperTest.WithoutSpinLockGuardedWrapper
core/memory_tracker/cc/memory_tracker_test.cpp:345: Failure
Expected equality of these values:
  2u
    Which is: 2
  counter
    Which is: 1
core/memory_tracker/cc/memory_tracker_test.cpp:353: Failure
Expected equality of these values:
  1u
    Which is: 1
  counter
    Which is: 2
[  FAILED  ] WrapperTest.WithoutSpinLockGuardedWrapper (9 ms)
```

Bug: b/163547280

### DIFF
--- a/core/memory_tracker/cc/memory_tracker_test.cpp
+++ b/core/memory_tracker/cc/memory_tracker_test.cpp
@@ -331,31 +331,6 @@ TEST_F(SignalBlockerTest, BlockerRecursiveSafe) {
 
 using WrapperTest = TestFixture;
 
-// Call the member function: DoTask() without spin lock guard.
-TEST_F(WrapperTest, WithoutSpinLockGuardedWrapper) {
-  Init();
-  uint32_t counter = 0u;
-  RunInTwoThreads(
-      // Initiated first
-      [this, &counter]() {
-        DoTask([this, &counter]() {
-          m_.unlock();
-          usleep(5000);
-          counter++;
-          EXPECT_EQ(2u, counter);
-        });
-      },
-      // Initiated secondly
-      [this, &counter]() {
-        DoTask([this, &counter]() {
-          m_.lock();
-          counter++;
-          EXPECT_EQ(1u, counter);
-        });
-      });
-  EXPECT_EQ(2u, counter);
-}
-
 // Call the member function: DoTask() with spin lock guarded.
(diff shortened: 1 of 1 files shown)


johnplate (approved):

matled (approved):
Nice detailed commit message! A bit like this example: https://dhwthompson.com/2019/my-favourite-git-commit :)

From a quick look around the code, this test indeed does not appear to test any of the "code under test" (i.e. the memory_tracker API/module). I wonder if this tries to establish some assumption used as a precondition by other tests (e.g. of the test helper or something like that)? Either way, unless we figure out what that assumption is, how to establish this reliably and document it, it is better to remove this.

Any sleep in tests is pretty suspicious and should have a good explanation on why it is necessary. One case I came across where I used sleep in tests was when I wanted to test that "nothing happens", i.e. all events have settled and don't trigger anything further. The code under test did not have a good way to observe this other than waiting. In that case, a flake would always be a true failure and it could only succeed accidentally.

hevrard:
> I wonder if this tries to establish some assumption used as a precondition by other tests (e.g. of the test helper or something like that)?

I also wondered that, but as far as I can tell, it doesn't seem to test anything meaningful. The other test cases using `RunInTwoThreads` are testing uses of `SpinLock`, and I don't see how the test that's being removed here could be relevant as a prerequisite for these others.

PR #853, opened 2021-07-01
Get rid of bind.proto.

pmuetschard:
There is no reason that the status enum is defined in a proto, since it's not actually used from any other proto.


hevrard (approved):
Just need to fix gofmt, see presubmit failure

PR #856, opened 2021-07-01
Update the orientation icon based on the chosen orientation.

pmuetschard:
The main orientation icon in the toolbar now updates to whatever the current orientation is, rather than always showing Y-up.


stellama0208 (approved):
LGTM! Just the PR explanation seems to get mixed with another PR, maybe a bit attention at the merging time.

PR #859, opened 2021-07-02
Get ready for the 2.0.0 release.

pmuetschard:
- Removes the experimental flags and labels of the things to be released.
- Sets the version for the release (i.e. commit 2734cad will be tagged for the release)
- Increments the version for the next release (i.e. commit 9cb0318 will be the first commit after the release)

This PR to be submitted just before cutting the release branch.


version.bzl
```diff
@@ -14,7 +14,7 @@
 
 # True source of AGI versions.
 # Increment these numbers immediately after releasing a new version.
-AGI_VERSION_MAJOR="1"
+AGI_VERSION_MAJOR="2"
 AGI_VERSION_MINOR="1"
```
hevrard, line 18: As Stella raised, shouldn't this be `0`?

johnplate:
> Mostly LGTM! Just to check do we want the first commit after the release to start with 2.1.0 or 2.0.1?

The version 2.0.1 would indicate a patch (bug fixes only) of the release 2.0.0, while 2.1.0 indicates a new development version with new features, similar to our current 1.1.0. So Pascal indeed means 2.1.0.

version.bzl
johnplate: See my reply to Stella.

johnplate (approved):

PR #860, opened 2021-07-03
Swap Block tab and Struct tab inside Memory view.

stellama0208:
With Struct tab giving more straight-forward memory info, and
Block tab providing fall back mechanism, this PR makes the Struct 
tab the default tab when switching to the Memory view. 

Bug: N/A.


hevrard (approved):
Thanks for this!

PR #862, opened 2021-07-05
Add state rebuilder part for VK_EXT_line_rasterization

hevrard:
Add missing part for support of VK_EXT_line_rasterization: make sure
to request this extension's features when creating devices as part of
the state rebuilder.

Also need to declare pNext as `void*` (and not `const void*`) at the
GAPIL level such that the Golang code passes type checking. This is
consistent with other vulkan structs in GAPIL that always use `void*`
as a type for their `pNext` field.

Bug: b/192662962

### DIFF
--- a/gapis/api/vulkan/extensions/ext_line_rasterization.api
+++ b/gapis/api/vulkan/extensions/ext_line_rasterization.api
@@ -64,7 +64,7 @@ enum VkLineRasterizationModeEXT: u32 {
 @extension("VK_EXT_line_rasterization")
 class VkPhysicalDeviceLineRasterizationFeaturesEXT {
   VkStructureType sType
-  const void*     pNext
+  void*           pNext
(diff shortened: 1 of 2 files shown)


johnplate (approved):

gapis/api/vulkan/extensions/ext_line_rasterization.api
```diff
@@ -64,7 +64,7 @@ enum VkLineRasterizationModeEXT: u32 {
 @extension("VK_EXT_line_rasterization")
 class VkPhysicalDeviceLineRasterizationFeaturesEXT {
   VkStructureType sType
-  const void*     pNext
+  void*           pNext
```
pmuetschard, line 67: This change was wrong, but I understand why you did it. Cleaning up the `const void* pNext` vs `void* pNext` usage throughout our API files is on my TODO list.

PR #863, opened 2021-07-05
Fix retired swapchain image deletion

hevrard:
The deletion of retired swapchain images could actually remove a
non-swapchain image from the global Images map that tracks Vulkan
images as part of the Vulkan state.

Consider these commands, and how our code was working before this
change:

```
vkCreateSwapchainKHR(oldSwapchain:0x123) # retires swapchain 0x123
                                         # which has img 0x456
					 # that is not acquired

vkCreateImage() # Returns a fresh new image with handle 0x456

vkDestroySwapchainKHR(0x123) # This removes img handle 0x456 from
                             # Images, and that is incorrect
```

This change fixes how retired swapchain images are handled, and
properly deletes them from Images only when it is relevant. See
code comments for details.

Note: this change was first landed as
https://github.com/google/agi/pull/833, but it was incorrect: in
`gapis/api/vulkan/extensions/khr_swapchain.api:vkDestroySwapchainKHR()`,
it was assuming that the handles of swapchain images would always be
found in `Images`, although that is not the case since some of the
swapchain images may have been deleted from Images if the swapchain
has been retired. This new version of the fix makes sure to only look
up swapchain images whose handle is present in Images. See
b/192265951.

Bug: b/190829522, b/192265951

Test: manual, on an app that systematically triggered an error, and on
the app mentioned in b/192265951 where the first version of this fix
didn't work.

### DIFF
--- a/gapis/api/vulkan/extensions/khr_swapchain.api
+++ b/gapis/api/vulkan/extensions/khr_swapchain.api
@@ -119,6 +119,18 @@ cmd VkResult vkCreateSwapchainKHR(
   if pCreateInfo == null { vkErrorNullPointer("VkSwapchainCreateInfoKHR") }
   create_info := pCreateInfo[0]
 
+  if create_info.oldSwapchain in Swapchains {
+    old := Swapchains[create_info.oldSwapchain]
+    old.Retired = true
+    // Delete images from the oldSwapchain that is being retired, unless these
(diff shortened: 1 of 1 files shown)


johnplate (approved):

hevrard:
(My first push didn't contain the latest fix...)

PR #868, opened 2021-07-07
Fix screen auto-unlock for some devices

hevrard:
Some devices do need a pause between the "wakeup" and the "menu" key
events in order to properly unlock their screens.

Also, remove unreachable code at the end of UnlockScreen(): the
previous switch statement has a default case which always reach a
return, hence the return after this switch statement is effectively
unreachable (the code compiles fine without it).

Bug: b/193011219

Test: manual: take a trace on a device which has a locked screen,
using a device where auto-unlocking was failing before this change.

### DIFF
--- a/core/os/android/adb/screen.go
+++ b/core/os/android/adb/screen.go
@@ -105,18 +105,17 @@ func (b *binding) UnlockScreen(ctx context.Context) (bool, error) {
 		// Devices may do unexpected transitions between screen
 		// states, so this code does not try to be smart about
 		// expected state changes: unless screenOnUnlocked, apply the
-		// wakeup (put screen on) and menu (unlock screen if no
-		// credentials required) key event sequence.
+		// wakeup (put screen on) and dismiss-keyguard (unlock screen
+		// if no credentials required) sequence.
 		if err := b.KeyEvent(ctx, android.KeyCode_Wakeup); err != nil {
 			return false, err
 		}
-		if err := b.KeyEvent(ctx, android.KeyCode_Menu); err != nil {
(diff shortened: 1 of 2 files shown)


yalcinmelihyasin (approved):

pau-baiget (approved):

core/os/android/adb/screen.go
```diff
@@ -105,18 +105,17 @@ func (b *binding) UnlockScreen(ctx context.Context) (bool, error) {
...
 		// expected state changes: unless screenOnUnlocked, apply the
-		// wakeup (put screen on) and menu (unlock screen if no
-		// credentials required) key event sequence.
+		// wakeup (put screen on) and dismiss-keyguard (unlock screen
+		// if no credentials required) sequence.
 		if err := b.KeyEvent(ctx, android.KeyCode_Wakeup); err != nil {
 			return false, err
 		}
-		if err := b.KeyEvent(ctx, android.KeyCode_Menu); err != nil {
+		if err := b.Shell("wm", "dismiss-keyguard").Run(ctx); err != nil {
```
pmuetschard, line 113: What version of Android introduced `wm dismiss-keyguard`? Was it added before Q?

hevrard: It seems to have been around for quite some time. This refactor commit from 2017 moved it to a new file within Android:
https://cs.android.com/android/_/android/platform/frameworks/base/+/5c3296af2d2b9717065b632c1848a45b514540a2

So it predates 2017, I'm confident it was there by Q.

PR #872, opened 2021-07-08
Fix the crash in Memory View when a command without a group selected

yalcinmelihyasin:


yalcinmelihyasin:
Linter moved the imports automatically.

pmuetschard (approved):
I cannot reproduce the crash, but the change is OK. Although, I would prefer we'd find out why command is null, when we expect it to not be null.

yalcinmelihyasin:
I will merge this but not close the bug. 

Although, one note, the other Source classes in Profile, Geometries and DeviceDependentModel all uses `Objects.equal()` so this one was the outlier.

PR #873, opened 2021-07-08
Guard unmap memory to avoid unmapping invalid memory location

yalcinmelihyasin:
b/183102240


pmuetschard (commented):
According to the spec it is invalid to unmap memory that is not mapped, so you should abort here with an error.

yalcinmelihyasin:
The actual bug is why we end up in this state as we should not be. I am closing this PR as this does not fix a problem.

pmuetschard:
Sorry if this wasn't clear from my comment: I think this is a good change. I'm just suggesting that in the false case, instead of doing nothing, we error out. The error would be surfaced to the user in the report view and could thus make it easier for a user to debug and fix their app.

I'm OK not doing anything, but just wanted to make sure you are not closing out the PR due to my comment.

PR #874, opened 2021-07-09
Remove flaky tests WithoutBlocker and WithoutSignalSafeWrapper

hevrard:
Both WithoutBlocker and WithoutSignalSafeWrapper test are flaky: they
relies on a sleep to force a certain order in thread scheduling,
although there is no guarantee that this particular scheduling order
happens in practice.

The body of these two tests are similar (except for a comment), so
we'll focus on WithoutBlocker, the same applies to
WithoutSignalSafeWrapper.

Let's see how WithoutBlocker can fail, but for that we first need a
bit of background on the test framework. FakeLock() simulates
acquiring a mutex: it will always succeed, but if there is two
FakeLock() calls in sequence without a FakeUnlock() in between, then
FakeLock() sets the deadlocked_ shared atomic bool to true, to
indicate that the thread that is trying to acquire FakeLock() two
times in a row would deadlock.

WithoutBlocker starts a first thread T1 that runs
RegisterHandlerAndTriggerSignal(), which itself starts another thread
T2. The test expects T2 to trigger a deadlock by calling FakeLock()
two times in a row, via a signal handler. I won't re-explain all the
code here, but here is a possible timeline that leads to T2 not
hitting the deadlock (and hence failing the test):

1. Init() calls normal_thread_run_before_signal_handler_.lock()

2. T1: register the `handler_func` signal handler

3. T1: start T2

4. T1: block on normal_thread_run_before_signal_handler_.lock()

5. T2: run the following: DoTask(), FakeLock(),
   normal_thread_run_before_signal_handler_.unlock() (this effectively
   unblocks T1, but it doesn't guarantee that T1 gets scheduled),
   usleep(5000) (I guess the idea here is that this sleep will force
   T1 to be scheduled, but that is not guaranteed), FakeUnlock().

6. T1: run the following:
   normal_thread_run_before_signal_handler_.lock(), send signal to T2,
   assuming T2 is still alive this triggers the signal handler on T2.

7. T2: run the signal handler: FakeLock(), FakeUnlock().

8. T1: wait for T2 to terminate, restore the signal handler, return to
   WithoutSignalSafeWrapper.

With this scenario where the usleep() in T2 is not forcing the
scheduling of T1, there is no competing calls to FakeLock(), and thus
the test expectation to see deadlocked_ set to true fails.

We have seen this happen in some CI-linux-test runs:

```
[ RUN      ] WrapperTest.WithoutSignalSafeWrapper
core/memory_tracker/cc/memory_tracker_test.cpp:403: Failure
Value of: deadlocked_.load(std::memory_order_seq_cst)
  Actual: false
Expected: true
[  FAILED  ] WrapperTest.WithoutSignalSafeWrapper (14 ms)
```

So these two tests are flaky. Also, they don't test any primitive from
the memory tracker, they only use the test framework primitives. I
guess both tests were added as counterparts to WithBlocker and
WithSignalSafeWrapper, to check that we *may* have a deadlock when the
memory checker safeguarding primitives are not used, but there is no
strong guarantee that a deadlock happens.

Bug: b/192948700

### DIFF
--- a/core/memory_tracker/cc/memory_tracker_test.cpp
+++ b/core/memory_tracker/cc/memory_tracker_test.cpp
@@ -253,24 +253,6 @@ TEST_F(SpinLockTest, WithSpinLockGuard) {
 
 using SignalBlockerTest = TestFixture;
 
-// A thread acquires the a lock first, then a signal interrupts the thread and
-// acquires the lock again. This results into a deadlock.
-TEST_F(SignalBlockerTest, WithoutBlocker) {
-  Init();
-  auto normal_thread_func = [](void*) -> void* {
-    unique_test_->DoTask([]() { usleep(5000); });
-    return nullptr;
-  };
-
-  std::thread start_normal_thread_and_send_signal(
-      [this, &normal_thread_func]() {
-        RegisterHandlerAndTriggerSignal(normal_thread_func);
-      });
-  start_normal_thread_and_send_signal.join();
-  // Expect dead locked
-  EXPECT_TRUE(deadlocked_.load(std::memory_order_seq_cst));
-}
-
 // A thread acquires the a lock first, then a signal tries to interrupt the
 // thread. But the signal is blocked, signal handler is called after the thread
 // finishes its job. This avoids dead lock.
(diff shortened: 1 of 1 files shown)


johnplate (approved):

matled (approved):
Thanks for investigating and improving the signal we get from our tests!

PR #875, opened 2021-07-09
Fix the crash after opening a second trace in AGI

yalcinmelihyasin:
while Memory Pane is open

b/193250359

### DIFF
--- a/gapic/src/main/com/google/gapid/models/ConstantSets.java
+++ b/gapic/src/main/com/google/gapid/models/ConstantSets.java
@@ -91,7 +91,7 @@ public class ConstantSets {
   }
 
   public static Service.Constant find(Service.ConstantSet constants, Pod.Value value) {
-    if (!Pods.mayBeConstant(value) || constants == null) {
+    if (!Pods.mayBeConstant(value)) {
       return Service.Constant.getDefaultInstance();
     }
     long numValue = Pods.getConstant(value);
--- a/gapic/src/main/com/google/gapid/models/ConstantSets.java
+++ b/gapic/src/main/com/google/gapid/models/ConstantSets.java
@@ -33,7 +33,6 @@ import com.google.gapid.util.MoreFutures;
 import com.google.gapid.util.Pods;
 
 import java.util.List;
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/models/ConstantSets.java
```diff
@@ -94,6 +95,9 @@ public void onReplayDeviceChanged(Instance dev) {
     if (!Pods.mayBeConstant(value)) {
       return Service.Constant.getDefaultInstance();
     }
+    if (Objects.isNull(constants)) {
+      return Service.Constant.getDefaultInstance();
+    }
```
stellama0208, line 100: Didn't reproduce it but it's good to have this null check in. Only a nit that `constants != null` is enough and it can be combined with the check above.

johnplate (approved):

pmuetschard (changes requested):
Can we instead ensure that we don't call this function passing in null? I think it's perfectly fine for a function to throw an NPE if it's passed in a null instance.

johnplate (approved):

gapic/src/main/com/google/gapid/models/Memory.java
```diff
@@ -549,8 +549,12 @@ public String getTypeFormatted() {
 
     public String getValueFormatted() {
       if (type.getTyCase() == TyCase.ENUM && type.getEnum().hasConstants()) {
-        return ConstantSets.find(typesModel.constants.getConstants(type.getEnum().getConstants()),
-            value.getPod()).getName();
+        Service.ConstantSet constantSet = typesModel.constants.getConstants(type.getEnum().getConstants());
+        if (constantSet != null) {
```
pmuetschard, line 553: if `constantSet` is null here, it has not been loaded and you should schedule the load and ensure the UI refreshes once it has been loaded.

pmuetschard (changes requested):

yalcinmelihyasin:
The actual fix requires a different work than this PR and have a lower priority. I am closing this PR and I will open when I have time to fix it.

PR #877, opened 2021-07-13
Update Devdoc

yalcinmelihyasin:
Gofuse requires the generated files to be in place.


hevrard (approved):
I'm not sure this is required? But I guess it doesn't hurt.

PR #880, opened 2021-07-14
Use the Perfetto API even when tracing to a file.

pmuetschard:
If the device supports it, use Perfetto's API to trace, when tracing to a file is requested, via the API's "OutputPath" capability.

### DIFF
--- a/gapis/perfetto/android/trace.go
+++ b/gapis/perfetto/android/trace.go
@@ -51,12 +51,21 @@ const (
 	readInterval               = 100 * time.Millisecond
 )
 
+type traceMode int
+
+const (
+	traceIntoFile traceMode = iota
+	traceDownloadAtEnd
+	traceDownloadWhileTracing
+)
+
 // Process represents a running Perfetto capture.
 type Process struct {
 	device         adb.Device
 	config         *perfetto_pb.TraceConfig
 	deferred       bool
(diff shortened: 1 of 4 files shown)


gapis/perfetto/android/trace.go
```diff
@@ -356,7 +380,26 @@ func (p *Process) captureWithClientApi(ctx context.Context, start task.Signal, s
...
 
 	var numWritten int64
-	if downloadWhileTracing {
+	if p.mode == traceIntoFile {
+		tmp, err := file.Temp()
+		if err != nil {
+			return 0, log.Err(ctx, err, "Failed to create a temp file")
+		}
+		if err := p.device.Pull(ctx, perfettoTraceFile, tmp.System()); err != nil {
+			return 0, err
```
stellama0208, line 389: nit: just wondering whether it's intended here to drop out error message extra wrapping.

stellama0208 (approved):
LGTM!

yalcinmelihyasin (approved):

gapis/perfetto/android/trace.go
pmuetschard: If the pull fails, it's error message is sensible. On the other hand, the file system errors would be kind of surprising, so they are wrapped.

PR #881, opened 2021-07-15
Add a way to set system properties for swarming tests

hevrard:
This is needed by some apps, e.g. for
Filament (https://github.com/google/filament) we need to set
'debug.filament.backend' to '2' to force the Vulkan backend.

Also, fix a typo.

Bug: b/187675797
Test: manual


johnplate (approved):
Nice, LGTM!

PR #882, opened 2021-07-16
Fix size calculation for vertex attrib with stride 0

### LINKED ISSUE #814: GLES1 game getting wrong image
**Environment information:**
 - APK version: versionCode=61603
 - ANGLE version: 89.0.4386.0
 - AGI version: v1.1.0-dev-20210110
 - Host OS: Android 
 If tracing on Android:
 - Device model: Pixel 4 XL (probably doesn't matter)
 - Android Version: 11/RP1A.201005.004/6782484 (probably doesn't matter)

**Bug description**
 - Original description: Replay leads to the overlay layer taking over the whole screen. Maybe an alpha / blending issue?  The drawcall thumbnails show that after the main gameplay rendering is completed, the overlay rendering for the game tutorial instructions is rendered but a huge part of the screen just becomes white, while it should be "transparent" to let the game rendering appear. See screenshots attached.
 - Pau's debugging notes:
   - I attached some screenshot in my previous comment that i thought was correct, but taking a closer look I discovered that replay is not working 100% correctly. The scores above each character are sometimes not rendered correctly.
   - I'm attaching more evidence here. I took a second trace (here) and looked for the drawcalls that are rendering the scores correctly and incorrectly (attached screenshots with expanded command trees). I took a few traces and I can always reproduce the same pattern: first character gets the upper score rendered correctly and then all the rest get it wrong in several ways. See attached dumped screenshot from AGI vs actual screenshot.
   - It feels like AGI is not keeping the internal state consistent after rendering the score panel for the lower-left (first) character of the game.

**Reproduction steps**
More details will be communicated privately ...

**Screenshots**
From Pau
first_character_good
second_character_wrong
screenshot_from_device-summoners-war
screenshot_agi_vkQueuePresent

- Please attach the generated _gapis.log_ and _gapic.log_ files you will find in the temp folder (e.g. /tmp/ on linux).
- If using Android: Please attach a full logcat dump (```adb logcat -d > logcat-full.txt```) that contains logs since AGI was started.

mikes-lunarg:
Previous calculation used stride as vertex size, resulting in memory
read of size 0 when stride was 0. New calculation uses vertex size,
offset, and stride to determine bounds of vertex data:

offset = (start_vertex * vertex_attrib_stride) + vertex_attrib_offset
size   = ((num_vertices - 1) * vertex_attrib_stride) + vertex_attrib_size

Fixes #814

Manual test with GLES1 game on ANGLE, which uses a vertex color
attribute with stride 0 to implement glColor4f.

### DIFF
--- a/gapis/api/vulkan/api/coherent_memory.api
+++ b/gapis/api/vulkan/api/coherent_memory.api
@@ -515,14 +515,18 @@ sub void trackMemoryInCurrentPipelineBoundVertexBuffers(u32 vertexCount, u32 ins
                 }
             }
         }
-        start_offset := bound_vertex_buffer.Offset + as!VkDeviceSize(start_vertex * vertex_binding.stride)
-        num := switch vertexCount == 0xFFFFFFFF {
-          case true:
-            backing_buf.Info.Size - start_offset
-          case false:
-            as!VkDeviceSize(num_vertices * vertex_binding.stride)
+        if num_vertices > 0 {
+          vertex_attrib := ldi.GraphicsPipeline.VertexInputState.AttributeDescriptions[vertex_binding.binding]
+          vertex_attrib_size := getElementAndTexelBlockSize(vertex_attrib.format).ElementSize
+          start_offset := bound_vertex_buffer.Offset + as!VkDeviceSize((start_vertex * vertex_binding.stride) + vertex_attrib.offset)
+          num := switch vertexCount == 0xFFFFFFFF {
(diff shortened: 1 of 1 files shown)


hevrard (approved):
LGTM, it would be nice to have someone else's opinion on this though!

mark-lunarg:
Tested fix locally, works fine, LGTM.

pmuetschard (approved):

PR #886, opened 2021-07-20
Revert "Properly implement VK_KHR_shader_float_controls."

yalcinmelihyasin:
This reverts commit 201f4d7bf879db76f6e665f16b263bb8b5dc6b97.

VkPhysicalDeviceFloatControlsPropertiesKHR does not exists in VkCreateDevice's valid pNext structure list.

This is the case for any Properties(KHR/EXT) structure. I believe this is not true.

https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkDeviceCreateInfo.html


pmuetschard (approved):
Oops. Yes, only the `*Features` structs are valid there.

PR #890, opened 2021-07-23
Don't crash if the shader static analysis fails.

pmuetschard:
SPIRV-cross will throw an exception if it fails to parse/process a shader. Don't bring down the entire server if this happens.

We will need to do proper error reporting/handling here in the future, but this commit fixes the crash.

Bug: http://b/194343914


purvisa-at-google-com (approved):

pmuetschard:
> might want to do a log error here if we can

this code doesn't currently depend on `core/cc` and I didn't want to add this dependency at this time.

PR #892, opened 2021-07-27
Fix a few typos, mostly in DEVDOC

hevrard:
Bug: n/a

### DIFF
--- a/DEVDOC.md
+++ b/DEVDOC.md
@@ -718,7 +718,7 @@ only the (transformed) capture commands needs to be replayed.
 
 If the user-requested replay does require to transform the initial commands,
 then the pre-warm replay is abandonned and a new replay of the transformed
-inital commands and then the transformed capture commands is executed.
+initial and capture commands is executed.
 
 ## Life of a perfetto trace (system profile trace)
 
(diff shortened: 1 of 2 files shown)


DEVDOC.md
```diff
@@ -718,7 +718,7 @@ only the (transformed) capture commands needs to be replayed.
 
 If the user-requested replay does require to transform the initial commands,
 then the pre-warm replay is abandonned and a new replay of the transformed
-inital commands and then the transformed capture commands are executed.
+inital commands and then the transformed capture commands is executed.
```
johnplate, line 721: "commands" is plural, so "are" was correct.
The last part of the sentence sounds weird, maybe we should rephrase it to "then the pre-warm replay is abandonned and a new replay of the transformed inital and capture commands are executed."

johnplate (approved):
LGTM, just one change.

DEVDOC.md
hevrard: I think the verb conjugates with "replay" here, as in: `a new replay (of <whatever>) is executed.`

Thanks for the suggestion, I applied it. But I kept "is". What do you think?

johnplate: Yes, thanks, you are right. I was confused by the initial weird sentence.

PR #893, opened 2021-07-29
Add support for vkGetDeviceQueue2

hevrard:
Add the command in GAPIL, and also use it in state rebuilding.

Bug: b/194480392
Test: capture an ANGLE trace replay doesn't crash anymore

### DIFF
--- a/gapii/cc/spy_base.cpp
+++ b/gapii/cc/spy_base.cpp
@@ -24,7 +24,7 @@
 
 // CurrentCaptureVersion is incremented on breaking changes to the capture
 // format. NB: Also update equally named field in gapis/capture/grahics.go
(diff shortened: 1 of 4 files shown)


hevrard:
Note that the GAPIL implementation of vkGetDeviceQueue2 mirrors vkGetDeviceQueue which is defined just above.

I don't think we can abstract their body into a `sub` because they use the `?` operator.

johnplate (approved):

hevrard:
This change in GAPIL code breaks backward compatibility of the trace used for smoketests, so I had to update this sample trace.

hevrard:
Also updated the trace version.

yalcinmelihyasin (approved):

PR #894, opened 2021-08-02
Clarify gofuse flag 'bazelout' in DEVDOC

johnplate:
Clarify the gofuse flag 'bazelout'. This should avoid that users will use an absolute path instead of only the sub-directory.

Fixes b/194912501

### DIFF
--- a/DEVDOC.md
+++ b/DEVDOC.md
@@ -49,8 +49,11 @@ mkdir <path-outside-agi-source>/agi-gofuse
 # Run gofuse with the previous directory as a target
 bazel run //cmd/gofuse -- -dir <path-to-agi-gofuse>
 
-# If you build with bazel build -c dbg pkg, the <path-to-bazelout> is `k8-dbg` on Linux.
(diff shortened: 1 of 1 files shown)


DEVDOC.md
```diff
@@ -49,8 +49,11 @@ mkdir <path-outside-agi-source>/agi-gofuse
 # Run gofuse with the previous directory as a target
 bazel run //cmd/gofuse -- -dir <path-to-agi-gofuse>
 
-# If you build with bazel build -c dbg pkg, the <path-to-bazelout> is `k8-dbg` on Linux.
-bazel run //cmd/gofuse -- -dir <path-to-agi-gofuse> -bazelout <path-to-bazelout>
+# If the previous command fails to correctly guess the sub-directory under `bazel-out`,
```
purvisa-at-google-com, line 52: Does "previous command" here refer to line 50? I see this isn't just a comment update but a modification of the command. I have to assume previous command does not refer to the line being replaced as the reader won't be able to see that anymore after submission.

johnplate: Correct, "previous command" refers to line 50.

purvisa-at-google-com (approved):

PR #899, opened 2021-08-10
Add a --size flag for `gapit screenshot`.

pmuetschard:
The size flag can be used to provide the requested width and height in the replay request.

### DIFF
--- a/cmd/gapit/flags.go
+++ b/cmd/gapit/flags.go
@@ -380,7 +380,8 @@ type (
 		Max           struct {
 			Overdraw int `help:"the amount of overdraw to map to white in the output"`
 		}
-		DisplayToSurface bool `help:"display the frames rendered in the replay back to the surface"`
+		Size             string `help:"requested framebuffer size (scales output). Either a single integer used as width and height, or <w>x<h>"`
+		DisplayToSurface bool   `help:"display the frames rendered in the replay back to the surface"`
 		CommandFilterFlags
 		CaptureFileFlags
 	}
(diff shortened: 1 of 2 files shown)


yalcinmelihyasin (approved):

cmd/gapit/flags.go
```diff
@@ -380,7 +380,8 @@ type (
 		Max           struct {
 			Overdraw int `help:"the amount of overdraw to map to white in the output"`
 		}
-		DisplayToSurface bool `help:"display the frames rendered in the replay back to the surface"`
+		Size             string `help:"replay request image size. Either a single integer size, or wxh"`
```
matled, line 383: I found the comment not completely clear and still don't understand fully what this does without reading too far into the implementation:
- Does this change the dimensions of the framebuffer and scales things? Or crop the image to this size?
- It looks like the default is the maximum value for uint32. From that it looks like this may be an upper bound, i.e. cropping if this is less than the size used in the replay?
- What does a single integer specify? It wasn't clear to me until I read the code that it would be width and height.
- wxh was a bit hard to parse as it mixes a string literal with two placeholders.

Possible suggestion, depending how the first question is answered:

Request framebuffer size for replay (scales output). Either a single integer used as width and height, or <width>x<height>.

matled (approved):

cmd/gapit/flags.go
pmuetschard: The size is part of the framebuffer request sent to the server. It represents a maximum area that the server will ensure the image will fit in. Providing max int values results in no scaling, since the image would always fit, but it's not a special value, we really could just use another arbitrarily large value.

While all this should probably be documented as part of the gapis API, the flag help text is not the place for it. I've updated the message as you suggested.

matled: I was still slightly confused as scaling would imply that the framebuffer is always the specified size, but from the large value that didn't look like the case. Just tried this PR manually and see that it will only downscale the image and doesn't upscale.

From that, I think this phrasing may clarify what this flag does:

requested maximum framebuffer size (downscales if necessary). Either a single integer used as width and height, or <w>x<h>"

matled:
Thanks!

Added one more clarifying comment to the flag phrasing.

Also, could you add the motivation for this flag to the commit message? It looks like something useful to have in general, I'm not quite sure what the motivation is, though.

PR #906, opened 2021-08-18
Increase timeout waiting for FTrace ready

purvisa-at-google-com:
Content Analysis are seeing some timeouts waiting for FTrace to be ready
when making perfetto system profiles. This increases the timeout to see
if timeout duration is the issue. If the problem persists I will investigate
what else might be going on here and reevaluate rolling the timeout back.

### DIFF
--- a/gapis/perfetto/android/trace.go
+++ b/gapis/perfetto/android/trace.go
@@ -315,8 +315,8 @@ func (p *Process) captureWithClientApi(ctx context.Context, start task.Signal, s
 				}
 				cnt += 1
 
-				// Give up after 1000 tries
-				if cnt == 1000 {
+				// Give up after 5000 tries
+				if cnt == 5000 {
 					timeout = true
 					sessionReadyFunc(ctx)
 					break


pmuetschard (commented):
Did you confirm that when it times out, the ftrace enable eventually does come up? 10s was already way higher than what was expected and was used to work around an issue that is considered fixed in the platform. If we see ftrace taking this long, we need to get the platform to fix it, not make our mess messier.

This wait should only be here for Q support and should not be needed on anything recent.

purvisa-at-google-com:
If ftrace initialises in normal time this change will have no effect. We are not slowing anything down.

If ftrace takes longer than normal to initialise but still completes in the new timeout period, this will result in a longer pause instead of a hard failure. That is an improvement to me, though you're right that if we can talk to the platform people and get ftrace to initalise faster that's an action we should also consider taking.

If ftrace is never going to initialise for some reason, this just causes a longer pause before the hard failure.

On balance this feels like a good change to me. The selection of the original timeout value was arbitrary anyway.

pmuetschard:
My question is whether you have confirmed that waiting longer helps. I.e. did you replicate the issue and can confirm that sometimes waiting more than 10s actually makes things work? My expectation is that on R, but certainly on S, we don't ever need to wait more than a few ms and that we can remove this wait entirely. If we can actually show that we sometimes have to wait this long for ftrace to come up we need to escalate.

pmuetschard:
Superseded by a fix to the underlying problem: https://github.com/google/agi/pull/924

PR #908, opened 2021-08-20
Adb un/root: add sleep when doing retry for unexpected output

purvisa-at-google-com:
We were not waiting long enough before retrying adb root or unroot
in the event of no output or unexpected output.
This adds a sleep and ups the retry count

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -41,7 +41,7 @@ const (
 	ErrRootFailed      = fault.Const("Device failed to switch to root")
 	ErrUnrootFailed    = fault.Const("Device failed to switch out of root")
 
-	maxRootAttempts                         = 5
+	maxRootAttempts                         = 50
 	gpuRenderStagesDataSourceDescriptorName = "gpu.renderstages"
 	gpuMemTotalDataSourceDescriptorName     = "android.gpu.memory"
 
@@ -108,6 +108,7 @@ retry:
 				continue retry
 			default:
 				// Some output we weren't expecting.
+				time.Sleep(time.Millisecond * 100)
 			}
 		}
 	}
@@ -141,6 +142,7 @@ retry:
 				continue retry
 			default:
 				// Some output we weren't expecting.
+				time.Sleep(time.Millisecond * 100)
 			}
 		}
 	}


core/os/android/adb/commands.go
```diff
@@ -108,6 +108,7 @@ retry:
 				continue retry
 			default:
 				// Some output we weren't expecting.
+				time.Sleep(time.Millisecond * 100)
```
pmuetschard, line 111: Shouldn't this be outside of the inner loop?

core/os/android/adb/commands.go
```diff
@@ -41,7 +41,7 @@ const (
 	ErrRootFailed      = fault.Const("Device failed to switch to root")
 	ErrUnrootFailed    = fault.Const("Device failed to switch out of root")
 
-	maxRootAttempts                         = 5
+	maxRootAttempts                         = 50
```
pmuetschard, line 44: I think we should increase the sleep time in response to the "restarting as ..." message, rather than the number of tries. I suspect that what is happening is that each successive time we run `adb [un]root` it is too early and adb aborts whatever work it was doing from the previous call and increasing the number of tries just means you increase your chances of getting lucky that it is fast enough.

In situations where the current version fails, how many times did it need to retry to succeed?

purvisa-at-google-com: purvisa-macbookpro3:~ purvisa$ adb root
adbd is already running as root
purvisa-macbookpro3:~ purvisa$ adb unroot
restarting adbd as non root
purvisa-macbookpro3:~ purvisa$ adb unroot
adbd not running as root
purvisa-macbookpro3:~ purvisa$ adb unroot
adbd not running as root
purvisa-macbookpro3:~ purvisa$ adb root # notice back of output from this command
purvisa-macbookpro3:~ purvisa$ adb root
adbd is already running as root

Somtimes no message is printed at all. In this case our code spins 5 times without a sleep. This is going to cause 5 rapid polls and then a device failure. I've added a sleep to prevent this, so unexpected output or no output at all gets a sleep before trying again with another poll.

I also increased the time from 0.5s (which seems very short given how long manual testing has shown the server to restart on my macbook. The benefit of polling rapidly many times is we get a fast turn around when the server is quick to restart. Polling fewer times with longer waits will result in worse best case performance.

core/os/android/adb/commands.go
purvisa-at-google-com: I was trying to work out why the retry label exists at all. The best guess I can make is that it was the intention of the original author to keep polling adb infinitely as long as it prints the restarting message. For all other messages we want to have a fixed timeout of N tries that are M seconds apart. I tried to keep that intention, though I thought about changing it as it seemed a little weird to me.

pmuetschard: My point is that this is inside the "for each line" loop. So if we get multiple lines of output, this makes it needlessly sleep after each unexpected line. I'm not saying we shouldn't sleep if we don't get any of the other cases, just that this sleep should be after we've processed all the lines. If the inner loop completes, we always took the default case, so just putting the sleep after the inner loop, will make us retry with a sleep in-between, which is what you mentioned above is fixing the underlying issue.

The retry label exists to quickly jump to the next try and ignore the remaining lines. In the current code, yes, you could just break out of the inner loop, but I guess it was written so that it would be easy to add something after the inner loop, which is what I'm proposing we do now.

To combine this with your comment above: should we treat no output with a successful return code as success, rather than try-again?

pau-baiget:
Closing pull request as this workaround doesn't seem to fix any of the aforementioned issues.

PR #911, opened 2021-08-24
Avoid clearing screen when selecting a different draw call

### LINKED ISSUE #900: Frame Profiler: Don't clear the screen when moving to a different drawcall
From user feedback:

"The clearing of the screen each time you step and then the long time for it to draw makes it hard to see what changed. Maybe generate the new frame offscreen and blit without erasing so that it’s possible to tell what changed."

mark-lunarg:
When selecting a new draw call in AGI, the framebuffer view panel clears and the image is replaced by a 'loading' image, and is eventually replaced with the image created by the newly-selected draw call.  This PR will, for framebuffer views only, skip the clear and loading image, so that the new image is rendered directly on top of the old one and making it easer to see what has changed.

The second commit skips the layout changes for the attachment selection control until the image has loaded. This prevents the framebuffer image from jumping around within the panel, again facilitating the ability to discern draw call results.

Fixes #900
Internal bug b/195041011

### DIFF
--- a/gapic/src/main/com/google/gapid/views/FramebufferView.java
+++ b/gapic/src/main/com/google/gapid/views/FramebufferView.java
@@ -354,6 +354,10 @@ public class FramebufferView extends Composite
         image.dispose();
       }
     }
+
+    public boolean isImageReady() {
+      return (image != null && image.hasFinished());
+    }
   }
 
   private static class AttachmentPicker extends Composite implements LoadingIndicator.Repaintable {
@@ -511,8 +515,11 @@ public class FramebufferView extends Composite
 
         image.setImage(attachment.getImage(widgets, this, this));
         label.setText(attachment.label);
-        image.requestLayout();
         label.requestLayout();
+        // Skip 'loading' image layout changes until attachment selection image is available
+        if (attachment.isImageReady()) {
+          image.requestLayout();
+        }
       }
     }
 


gapic/src/main/com/google/gapid/widgets/ImagePanel.java
```diff
@@ -343,7 +343,9 @@ private void setAlphaEnabled(boolean enabled) {
 
   @Override
   public void startLoading() {
-    if (imageComponent.isOpenGL()) {
+    // Avoid switching to 'loading' image for framebuffer views --
+    // the loading image makes it difficult to perceive image changes
+    if (imageComponent.isOpenGL() && (analyticsView != View.Framebuffer)) {
```
pmuetschard, line 348: I don't think we want to do this. `LoadablePanel` already has a delay built in. If it `stopLoading()` is called within 0.5s of `startLoading()`, the loading indicator is never shown.
I think the true solution here is finding out why it takes so much time to load the image if it's already in the cache.

pmuetschard (commented):
Sorry for the long delay.

gapic/src/main/com/google/gapid/widgets/ImagePanel.java
mark-lunarg: Loading attachment images takes multiple seconds in large traces.  Once they have been loaded the internals of `LoadablePanel` cause the image area to be cleared which creates a 'flash' even if the loading indicator is not shown.

gapic/src/main/com/google/gapid/views/FramebufferView.java
```diff
@@ -511,8 +511,11 @@ private void updateUi() {
 
         image.setImage(attachment.getImage(widgets, this, this));
         label.setText(attachment.label);
-        image.requestLayout();
-        label.requestLayout();
+        // Skip 'loading' image layout changes until attachment selection image is available
+        if (attachment.image.hasFinished()) {
+          label.requestLayout();
+          image.requestLayout();
+        }
```
pmuetschard, line 518: I like this part. I wonder if not requesting the layout works reliably on every platform. Can you let me know where you've tested this and I can check the rest?

Code-wise, two changes, please:
1) call `label.requestLayout()` always, since if the text changes we don't want it to get cut off.
2) add a `isImageReady()` to `Attachment`, which will `return image != null && image.hasFinished()` and use it here instead of reaching into the internals of `Attachment`. This will avoid any potential `null` issues, which is why `image` is private in `Attachment` in the first place.

pmuetschard (approved):

PR #914, opened 2021-08-25
ABI by name lookup fixes.

pmuetschard:
Currently the ABIs' names need to be unique, because we lookup ABIs by name. However, this lookup is only done in the Android parsing code, so we can loosen this restriction to only require unique names for the Android ABIs. This change also adds an init check to ensure the Android ABI names are unique. This check would fail at build time, as we use this during the codegen step.

### DIFF
--- a/core/os/android/adb/device.go
+++ b/core/os/android/adb/device.go
@@ -223,7 +223,7 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
 			if seen[abi] {
 				continue
 			}
-			d.To.Configuration.ABIs = append(d.To.Configuration.ABIs, device.ABIByName(abi))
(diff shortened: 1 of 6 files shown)


pmuetschard:
> Is this fixing an existing issue we observed? I haven't figured out from the code or description if this is a proactive improvement or some underlying issue. Do we have a bug for this?

The code currently assumes that, e.g. "arm64-v8a" means Android. This assumption will be invalidated with #848.

> > This check would fail at build time, as we use this during the codegen step.
> 
> Does this mean the check _was_ failing at build time before?

There was no check. I've added a check and I'm saying that even though it's implemented as a runtime check, our build would fail if the check fails as this code gets executed during codegen.

pmuetschard:
Ping for cla.

Issue #901, opened 2021-08-14
AGI missing all debug labels.


alecazam:
**Environment information:**
 - AGI version: Version 2021-2.0.0:aad09f92ea8e608b11001aae1f93af3cc1101096
 - Host OS: macOS 11.4
 If tracing on Android:
 - Device model: Samsung Galaxy 10SE w/Mali-G76 gpu
 - Android Version: 11

**Bug description**
I see that AGI docs state that it supports VK_EXT_debug_marker, but I don't see any of my labels  for assets like Texture.  Otherwise, I just see random pointer values for all commands, textures, geometry and shaders.  This is a Vulkan 1.0 app.

There was already a problem with Android not publishing the VK_EXT_debug_utils extensions properly, so we couldn't move to that for RenderDoc or AGI either.  So our code has to use the old and deprecated VK_EXT_debug_marker/report calls.

We could also use a way to dectect capture is in progress so we could turn on per-draw markers that we don't usually add to the command stream.  We do this in Metal.

That issue is here:
https://github.com/baldurk/renderdoc/issues/2205

**Reproduction steps**
Run a GPU app on AGI.
Perform a single frame capture.
Look at the textures, and geometry, and shaders and nothing has any labels.

**Stacktrace**
Not applicable

**Screenshots**
![image](https://user-images.githubusercontent.com/3750359/129461313-00b101fa-2e90-4181-8d6a-97db7d2d65b5.png)

alecazam:
Just to prove that we're labeling everything, here are the calls in the capture.

![image](https://user-images.githubusercontent.com/3750359/129462425-e5b34627-6e35-4c32-82d1-348502a12230.png)

This is not so beautiful either, opening up that command which should be a human-readable label.

![image](https://user-images.githubusercontent.com/3750359/129462450-9507b89d-8f3f-4120-b067-aa33af41a621.png)

alecazam:
Not sure but this warning is in the Log panel.

W 04:27:00.416 gapis Mismatch of command groups : [2:State Setting Group]%!V(vulkan.markerType=3) != 2

alecazam:
Maybe I'm looking at the wrong turn-down.

![image](https://user-images.githubusercontent.com/3750359/129462718-1ce6ff6b-b2ad-4695-9786-d28a5063f9ad.png)

There's another set of commands where the labels are displayed.  

[...]

PR #917, opened 2021-09-02
Debug Label Support Everywhere

pmuetschard:
Adds debug labels in lots of places. To do so, this change:
 - fixes support for `VK_EXT_debug_markers` where it was incomplete.
 - adds support for `VK_EXT_debug_utils`, which replaces the deprecated `VK_EXT_debug_markers` extension
 - labels everything in the sample app, so we have something to test
 - some minor cleanup of the `api` package (note this wan't enough in the end, but included anyways, since it helps)
 - adds a new interface for handles in the API and makes things implement it
 -  start using this new interface to expose labels to the GAPIS API
 - make the UI show the labels
 - as a side benefit, this also makes handles show up in HEX in the UI

There are more details in the individual commits.
Fixes: #901


pmuetschard:
I know this is a big change. Reviewing it is probably best done by going commit-by-commit, rather than as one big change.

pmuetschard:
Screenshot:
debug_labels

core/data/pod/pod.proto
```diff
@@ -33,6 +33,7 @@ message Value {
     sint32 sint32 = 10;
     uint64 uint64 = 11;
     sint64 sint64 = 12;
+    uint64 handle = 29;
```
matled, line 36: Up until now it appears that this Value has mostly a 1:1 mapping between types (in go?) to fields. The `handle` is different, though, as it adds more semantics without a separate type.

What is the reason for not storing a handle in a uint64? Is this mainly to format them differently?

Have you considered adding a new (wrapper) type to make this type safe instead of using a plain uint64? This may also resolve some of the special cases that are added in the Go code to handle the new `handle` case, which can't rely on the type itself.

Also, it appears this is specifically a Vulkan handle. Would it make sense to be more explicit about the field/type and name it `vulkan_handle`? If we use `handle` in our codebase exclusively to refer to a Vulkan handle it may be fine as-is.

Either way, this probably warrants a comment to explain what this is and why it is somewhat different from the other fields here.

core/data/pod/value.go
```diff
@@ -145,6 +145,11 @@ func NewValue(val interface{}) *Value {
 	return nil
 }
 
+// NewHandle returns a uint64 as a pod.Type_handle Value.
```
matled, line 148: If I understand correctly, this is intended for Vulkan handles. It may be worth highlighting this in the function comment.

gapis/api/data_group.go
```diff
@@ -79,10 +79,16 @@ func CreateEnumDataValue(typeName string, value fmt.Stringer) *DataValue {
 }
 
 func CreatePoDDataValue(ctx context.Context, s *GlobalState, typeName string, val interface{}) *DataValue {
+	var podVal *pod.Value
+	if handle, ok := val.(Handle); ok {
```
matled, line 83: Could we do this check inside `pod.NewValue`? I think this would be a bit cleaner and less error-prone.

matled (commented):
> I know this is a big change. Reviewing it is probably best done by going commit-by-commit, rather than as one big change.

Indeed. Is it possible to break this up into multiple pull requests? I think that would help to keep the review to a more reasonable size and the conversations from the feedback easier.

Just reviewed one commit, should be able to continue tomorrow.

pmuetschard:
> Is it possible to break this up into multiple pull requests?

I intend to submit it keeping the commits intact. I think it's advantageous to maintain the grouping that PRs provide over commits as it can give a better reference for someone looking at the commit history to see how these commits go together. I understand that this is a bit of a different review process, so let me know if reviewing commit-by-commit is infeasible. I intend to keep rebasing, when addressing comments, so the commit history will stay clean in the PR.

pmuetschard:
@matled I think I've addressed your concerns. Basically, `service.DataValue` shouldn't exist. I've already put an item in our list of cleanups for this. I had chosen a different way to make it work in the current system, but I think the way I have it now is better as it avoids those crutches, at the small expense of making `service.DataValue` replicate even more functionality of `box.proto`.

gapis/service/box/box.go
```diff
@@ -33,19 +33,19 @@ func NewValue(v interface{}) *Value {
 }
 
 // NewType returns the Type of value v
-func NewType(t reflect.Type) *Type {
-	return newBoxer().ty(t)
-}
+//func NewType(t reflect.Type) *Type {
```
matled, line 36: I guess this is from trying something? I'd suggest to (1) remove this or (2) make it proper code again.

Same for `Get()` below.

gapis/api/vulkan/links.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/google/gapid/gapis/service/path"
 )
 
+// When updating this file, you probably also want to update labels.go.
```
matled, line 25: The file is pretty large and it is easy to miss this comment. I don't fully understand what either links.go or labels.go does and how they relate exactly. Do you think there is a way to get the code of these closer together to make it obvious that these need to be updated/implemented together?

Also, the comment is relatively vague. Without a good understanding of what labels/links are (and there are no file comments), I cannot really figure out what kind of changes would require an update to the other file.

Maybe a file comment to explain what Links/Labels are, what the expectations/contract is would help, too, in addition to explaining how they are related.

matled (commented):
> @matled I think I've addressed your concerns. Basically, `service.DataValue` shouldn't exist. I've already put an item in our list of cleanups for this. I had chosen a different way to make it work in the current system, but I think the way I have it now is better as it avoids those crutches, at the small expense of making `service.DataValue` replicate even more functionality of `box.proto`.

Thanks.

gapis/service/box/box.go
pmuetschard: Opps, yes indeed. This was me confirming that the box type stuff is not at all used. I think it was intended to be used for the memory stuff, but then `memory_box` was added instead of making this one work for its usecase. Hence, the action item to clean this stuff up...

gapis/api/vulkan/links.go
pmuetschard: They go together, because they follow the same pattern - they add functions to the generated types to make them implement certain interfaces. This is Go's implicit interface stuff, which takes some getting used to.

However, I do like the idea of getting into habits of having file level comments to describe what functionality the file contains. If it becomes too hard to make that documentation specific, it probably means the file should be split. So, I've added a comment to `labels.go` and `links.go`.

matled: Looks good! That makes it a lot clearer what the files are about and why they may need to updated together.

matled (changes requested):
Looks good!

gapis/api/handle.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+// Handle is an interface implemented by every handle type in the API.
```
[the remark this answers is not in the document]
pmuetschard: Done.

matled (approved):

PR #923, opened 2021-09-15
Synchronize status logger's map accesses.

pmuetschard:
Bug: http://b/198709411

### DIFF
--- a/core/app/status/logger.go
+++ b/core/app/status/logger.go
@@ -17,6 +17,7 @@ package status
 import (
 	"context"
 	"runtime"
+	"sync"
 	"time"
 
 	"github.com/google/gapid/core/log"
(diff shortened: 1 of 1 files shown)


core/app/status/logger.go
```diff
@@ -34,19 +35,26 @@ func RegisterLogger(progressUpdateFreq time.Duration) Unregister {
...
 type statusLogger struct {
-	lastProgressUpdate map[*Task]time.Time
-	progressUpdateFreq time.Duration
+	lastProgressUpdate     map[*Task]time.Time
+	lastProgressUpdateLock sync.Mutex
+	progressUpdateFreq     time.Duration
 }
 
 func (l statusLogger) OnTaskStart(ctx context.Context, t *Task) {
 	log.I(ctx, "%v Started", t)
```
purvisa-at-google-com, line 44: Should this move inside the lock? Has a task really started if it's in a queue waiting to acquire a lock?

core/app/status/logger.go
```diff
@@ -34,19 +35,26 @@ func RegisterLogger(progressUpdateFreq time.Duration) Unregister {
...
 	l.lastProgressUpdate[t] = time.Now()
+	l.lastProgressUpdateLock.Unlock()
 }
 
 func (l statusLogger) OnTaskProgress(ctx context.Context, t *Task) {
+	l.lastProgressUpdateLock.Lock()
 	if time.Since(l.lastProgressUpdate[t]) > l.progressUpdateFreq {
-		log.I(ctx, "%v (%v%%) %v", t, t.Completion(), t.TimeSinceStart())
 		l.lastProgressUpdate[t] = time.Now()
+		l.lastProgressUpdateLock.Unlock()
```
purvisa-at-google-com, line 54: Might be safer to unlock once at the end of the function and have a single return point?

purvisa-at-google-com (approved):
Just nits, LGTM

core/app/status/logger.go
pmuetschard: Moved the log message to after the lock. Don't want to log while holding the lock.

core/app/status/logger.go
pmuetschard: Done.

PR #928, opened 2021-09-21
Invert the trace type / device selection order.

pmuetschard:
Changes the trace dialog to ask the user for the trace type first, followed by selecting a device for that trace type. Also contains a small refactor to remove the "API" field from the trace protos, and a small cleanup for the validation UI.

Bug: http://b/195970344


purvisa-at-google-com (approved):
There is a lot of this and I'm unfamiliar with the code, but nothing here stands out to me as wrong.

PR #935, opened 2021-09-27
Add server side command filtering

purvisa-at-google-com:
When listing the commands for the command tree, support filtering by:

Host Commands
Begin/End Marker Commands
Non Renderpass Commands
Sync Commands

### DIFF
--- a/gapis/api/vulkan/api/synchronization.api
+++ b/gapis/api/vulkan/api/synchronization.api
@@ -48,6 +48,7 @@
 }
 
 @indirect("VkDevice")
+@sync_command
 cmd VkResult vkCreateFence(
     VkDevice                     device,
     const VkFenceCreateInfo*     pCreateInfo,
@@ -85,6 +86,7 @@ cmd VkResult vkCreateFence(
 }
 
 @indirect("VkDevice")
+@sync_command
 cmd void vkDestroyFence(
     VkDevice                     device,
     VkFence                      fence,
@@ -94,6 +96,7 @@ cmd void vkDestroyFence(
 }
 
 @indirect("VkDevice")
+@sync_command
 cmd VkResult vkResetFences(
     VkDevice       device,
     u32            fenceCount,
@@ -110,6 +113,7 @@ cmd VkResult vkResetFences(
 }
 
 @indirect("VkDevice")
+@sync_command
 @custom
 cmd VkResult vkGetFenceStatus(
     VkDevice device,
@@ -130,6 +134,7 @@ cmd VkResult vkGetFenceStatus(
 }
 
 @indirect("VkDevice")
+@sync_command
 @threadsafe
 @alive
 @custom
@@ -172,6 +177,7 @@ cmd VkResult vkWaitForFences(
 }
 
 @indirect("VkDevice")
+@sync_command
 cmd VkResult vkCreateSemaphore(
     VkDevice                     device,
     const VkSemaphoreCreateInfo* pCreateInfo,
@@ -208,6 +214,7 @@ cmd VkResult vkCreateSemaphore(
 }
 
 @indirect("VkDevice")
+@sync_command
 cmd void vkDestroySemaphore(
     VkDevice                     device,
     VkSemaphore                  semaphore,
@@ -217,6 +224,7 @@ cmd void vkDestroySemaphore(
 }
 
 // Partial implementation
+@sync_command
 @since("1.2")
 @indirect("VkDevice")
 cmd VkResult vkSignalSemaphore(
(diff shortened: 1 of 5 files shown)


gapis/api/vulkan/api/synchronization.api
```diff
@@ -48,6 +48,7 @@
 }
 
 @indirect("VkDevice")
+@sync_command
```
pmuetschard, line 51: The filtering of sync commands should only filter them inside the submit command (i.e. only filter them if they are sub-commands). While we may provide functionality to filter the host coordination group more narrowly in the future, for now, expanding that group puts the user into the debugging territory and this feature is intended to focus on profiling only. So, I think we should only put the flag on sync commands that are of the form `vkCmd*` and in the filter only filter them if they are sub-commands.

gapis/resolve/filter.go
```diff
@@ -70,5 +70,26 @@ func buildFilter(
 			return (cmd.CmdFlags().IsExecutedDraw() && len(idx) > 1) || cmd.CmdFlags().IsSubmission()
 		})
 	}
+	if f.GetSuppressHostCommands() {
+		filters = append(filters, func(id api.CmdID, cmd api.Cmd, s *api.GlobalState, idx api.SubCmdIdx) bool {
+			return cmd.CmdFlags().IsSubmission()
```
pmuetschard, line 75: This should not filter-out the `vkQueuePresentKHR` commands.

gapis/resolve/filter.go
```diff
@@ -70,5 +70,26 @@ func buildFilter(
...
+		})
+	}
+	if f.GetSuppressBeginEndMarkers() {
+		filters = append(filters, func(id api.CmdID, cmd api.Cmd, s *api.GlobalState, idx api.SubCmdIdx) bool {
+			return !(cmd.CmdFlags().IsPushUserMarker() || cmd.CmdFlags().IsPopUserMarker() || cmd.CmdFlags().IsUserMarker())
```
pmuetschard, line 80: This should also filter out the `vkBeginRenderPass` and `vkEndRenderPass` commands, as well as the ones that start/end the transform feedback grouping. Essentially, all the commands that are implied by a group starting/ending.

gapis/resolve/filter.go
purvisa-at-google-com: What kind of logic condition can I write to exclude this specific command? Or do I need to give it a new special label so I can identify it for the purposes of exception?

gapis/resolve/filter.go
purvisa-at-google-com: Again, I suspect this means adding more custom labels to commands? Is there any way of working this using the existing labels?

gapis/api/vulkan/api/synchronization.api
purvisa-at-google-com: I think the right way to do this is still label all sync commands as sync commands, regardless of whether they are host side or device side, and then change the filter condition to "if(isSync() && isSubcommand())".

gapis/resolve/filter.go
purvisa-at-google-com: @frame_end looks like a suitable way of supporting this.

yalcinmelihyasin: You can cast  cmd to VkQueuePresentKHR and check if it returns ok or not, similar to many operation in transform system.

gapis/api/vulkan/api/synchronization.api
pmuetschard: Sounds good.

gapis/resolve/filter.go
pmuetschard: I would prefer the annotation way for now, but If you'd prefer checking for the Vulkan commands explicitly that's OK, too, since we'll remove the API abstraction some time soon. Also include the frame boundary command `vkFrameBoundaryANDROID`.

gapis/resolve/filter.go
pmuetschard: I don't think there is anything currently.

gapis/resolve/filter.go
purvisa-at-google-com: Implemented this with @frame_end which captures the vkFrameBoundaryANDROID case too

gapis/resolve/filter.go
purvisa-at-google-com: Added a new label for this @begin_end_renderpass

gapis/api/vulkan/api/synchronization.api
purvisa-at-google-com: done

PR #944, opened 2021-10-07
KHR_buffer_device address

yalcinmelihyasin:
Implement KHR_buffer_device address extension

### DIFF
--- a/gapis/api/vulkan/extensions/khr_buffer_device_address.api
+++ b/gapis/api/vulkan/extensions/khr_buffer_device_address.api
@@ -47,56 +47,15 @@
 @extension("VK_KHR_buffer_device_address") define VK_KHR_BUFFER_DEVICE_ADDRESS_SPEC_VERSION 1
 @extension("VK_KHR_buffer_device_address") define VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME "VK_KHR_buffer_device_address"
 
-@extension("VK_KHR_buffer_device_address")
-class VkBufferDeviceAddressInfoKHR {
-    VkStructureType    sType
-    const void*        pNext
-    VkBuffer           buffer
-}
-
-@extension("VK_KHR_buffer_device_address")
-class VkDeviceMemoryOpaqueCaptureAddressInfoKHR {
-    VkStructureType     sType
-    const void*         pNext
-    VkDeviceMemory      memory
-}
-
-@extension("VK_KHR_buffer_device_address")
-class VkBufferOpaqueCaptureAddressCreateInfoKHR {
-    VkStructureType     sType
-    const void*         pNext
-    u64                 opaqueCaptureAddress
-}
-
-@extension("VK_KHR_buffer_device_address")
-class VkMemoryOpaqueCaptureAddressAllocateInfoKHR {
-    VkStructureType     sType
-    const void*         pNext
-    u64                 opaqueCaptureAddress
-}
-
-@extension("VK_KHR_buffer_device_address")
-class VkPhysicalDeviceBufferDeviceAddressFeaturesKHR {
-    VkStructureType     sType
-    void*               pNext
-    VkBool32            bufferDeviceAddress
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/api/physical_device.api
```diff
@@ -53,6 +53,9 @@
   @unused ref!SubgroupProperties         SubgroupProperties
   @unused ref!PhysicalDeviceIDProperties IDProperties
 
+  // Vulkan 1.2 core
+  @unused ref!PhysicalDeviceBufferDeviceAddressFeatures PhysicalDeviceBufferDeviceAddressFeatures
```
pmuetschard, line 57: Why do we have both? This field doesn't appear to be used.
Also, add it to the end, so we don't break backwards compatibility.

gapis/api/vulkan/api/types.api
```diff
@@ -75,4 +76,4 @@ type u32 VkSampleMask
 
 // Vulkan 1.1 core
 @replay_remap @nonDispatchHandle type u64 VkDescriptorUpdateTemplate
-@replay_remap @nonDispatchHandle type u64 VkSamplerYcbcrConversion
+@replay_remap @nonDispatchHandle type u64 VkSamplerYcbcrConversion
```
pmuetschard, line 79: undo

gapis/api/vulkan/api/device_address.api
```diff
@@ -0,0 +1,181 @@
+// Copyright (C) 2020 Google Inc.
```
pmuetschard, line 1: should this be `buffer_device_address.api`?

gapis/api/vulkan/vulkan.api
```diff
@@ -302,6 +304,7 @@ sub ref!ExtensionSet supportedDeviceExtensions() {
   supported.ExtensionNames["VK_KHR_shader_terminate_invocation"] = true
   supported.ExtensionNames["VK_EXT_blend_operation_advanced"] = true
   supported.ExtensionNames["VK_EXT_index_type_uint8"] = true
+  supported.ExtensionNames["VK_KHR_buffer_device_address"] = false
```
pmuetschard, line 307: I thought we weren't going to expose this extension yet, until it's completely implemented? I'm not seeing any of the todos we had talked about? It's possible I am missing something or misremembering.

pmuetschard (commented):
Please be more consistent with whitespace in your API file changes. Close to the C++ style guide would be nice. E.g. there are changes in this PR with 2 space indent and also 4 indent, sometimes both in the same file/function.

yalcinmelihyasin:
> Please be more consistent with whitespace in your API file changes. Close to the C++ style guide would be nice. E.g. there are changes in this PR with 2 space indent and also 4 indent, sometimes both in the same file/function.

I've realized that as well. I think it's due to using different IDEs. After merging this, I can do a round of whitespace fix. so that everything would be consistent.

gapis/api/vulkan/vulkan.api
yalcinmelihyasin: I've assumed that if it's `false`, it would mean that we wouldn't be exposing but I'll remove it from here just in case.

gapis/api/vulkan/api/physical_device.api
yalcinmelihyasin: This is not doable in the long term.

It's either we group extensions by version or we scrap that and put them randomly without no specific order. Because we cannot guarantee that our implementation priority will match the Vulkan version.

If backwards compatibility is important we should ensure it with another way.

yalcinmelihyasin: And with this release due to vertex attribute fix, we will break the backward compatibility anyway, so it doesn't matter. We should I believe do all the breaking changes we think of before the next release and do them at once.

yalcinmelihyasin: I've removed the KHR one.

I think in terms of promoted extensions, we can use the core version without KHR unless we need KHR explicitly.

PR #950, opened 2021-10-12
Fix vertex attribute divisor

yalcinmelihyasin:
By:

- Adding the extension to state reconstruction
- As part of that making vertexInputData optional(as spec states)
- And fixing the code around


pmuetschard (commented):
This is breaking backwards compat. Is it blocking anything? If not, would it be possible to hold off on merging before the release?

yalcinmelihyasin:
> This is breaking backwards compat. Is it blocking anything? If not, would it be possible to hold off on merging before the release?

We will break this anyway, so I am merging this and the other breaking changes. So that we break only once.

PR #967, opened 2021-10-29
VK_KHR_Multiview

yalcinmelihyasin:
Implement VK_KHR_Multiview

### DIFF
--- a/gapis/api/vulkan/api/device.api
+++ b/gapis/api/vulkan/api/device.api
@@ -59,6 +59,7 @@
   @unused ref!HalfPrecisionStorageFeatures HalfPrecisionStorageFeatures
   @unused ref!SamplerYcbcrConversionFeatures SamplerYcbcrConversionFeatures
   @unused ref!PhysicalDeviceProtectedMemoryFeatures PhysicalDeviceProtectedMemoryFeatures
+  @unused ref!PhysicalDeviceMultiviewFeatures PhysicalDeviceMultiviewFeatures
   @unused ref!PhysicalDeviceScalarBlockLayoutFeaturesEXT PhysicalDeviceScalarBlockLayoutFeaturesEXT
 
   @unused ref!Vk8BitStorageFeatures Vk8BitStorageFeatures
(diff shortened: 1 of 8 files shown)


gapis/api/vulkan/api/device.api
```diff
@@ -59,6 +59,7 @@
   @unused ref!HalfPrecisionStorageFeatures HalfPrecisionStorageFeatures
   @unused ref!SamplerYcbcrConversionFeatures SamplerYcbcrConversionFeatures
   @unused ref!PhysicalDeviceProtectedMemoryFeatures PhysicalDeviceProtectedMemoryFeatures
+  @unused ref!PhysicalDeviceMultiviewFeatures PhysicalDeviceMultiviewFeatures
```
pmuetschard, line 62: Add it to the end to not break backwards-compat. I would prefer if we don't have to bump the major version before our next release. I'm OK if there are multiple `// Vulkan 1.1` sections in structs. We can re-order things after the release if we want to. I think we'll also have to decide on how we are going to handle some of that and extensions vs core stuff when I get back.

gapis/api/vulkan/api/physical_device.api
```diff
@@ -52,6 +52,7 @@
   @unused ref!Maintenance3Properties                  Maintenance3Properties
   @unused ref!SubgroupProperties                      SubgroupProperties
   @unused ref!PhysicalDeviceIDProperties              IDProperties
+  @unused ref!PhysicalDeviceMultiviewProperties       MultiviewProperties
```
pmuetschard, line 55: move to end of struct.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -180,6 +181,36 @@ cmd VkResult vkCreateRenderPass(
...
+          ext := as!VkRenderPassMultiviewCreateInfo*(next.Ptr)[0]
+
+          renderPass.MultiviewInfo = new!RenderPassMultiviewCreateInfo(
+            SubpassCount: ext.subpassCount,
+            DependencyCount: ext.dependencyCount,
+            CorrelationMaskCount: ext.correlationMaskCount,
+          )
+
+          if ext.pViewMasks != null {
+            viewMasks := ext.pViewMasks[0:ext.subpassCount]
```
pmuetschard, line 194: can you use the `clone` primitive for all these?

gapis/api/vulkan/extensions/khr_multiview.api
```diff
@@ -0,0 +1,99 @@
...
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+///////////////
+//  Structs  //
+///////////////
+
+@extension("VK_KHR_multiview") define VK_KHR_MULTIVIEW_SPEC_VERSION   1
```
pmuetschard, line 43: We're going to need to think about how we are dealing with KHR stuff that is now core. I feel like we are following different patterns in different places. Something to think about and clean up soon.

pmuetschard (commented):
Please change the title of this PR and give it a description.

gapis/api/vulkan/api/device.api
yalcinmelihyasin: We have decided to break it now(as we merged a fix that already breaks it).

purvisa-at-google-com (approved):

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Done!

PR #976, opened 2021-11-16
Added value representation to memory box

ttanatb:
Added calculation for turning char arrays to string
Screen Shot 2021-11-15 at 5 13 00 PM


gapis/resolve/memory.go
```diff
@@ -181,6 +182,46 @@ func pNextPrecision(typeIndex uint64) (int, error) {
...
 }
 
+// expandCharArrays goes through the range and attaches a human-readable representation to char arrays/slices
+func expandCharArrays(ctx context.Context, ranges []*service.TypedMemoryRange) ([]*service.TypedMemoryRange, error) {
+	if len(ranges) == 0 {
+		return ranges, nil
+	}
+
+	for _, rng := range ranges {
+		ty, err := types.GetType(rng.Type.TypeIndex)
```
purvisa-at-google-com, line 192: I know the go-ism is to use short / single letter names for local variables, but we've found this a real obstacle to reading the gapis code in the past. Melih has made an effort to expand the names of variables to be more descriptive in the past, so it would be great if we could continue that trend here with new code. Perhaps type instead of ty or range instead of rng. This code is neat and much more readable than the gapis average, but it's still good to continue with the descriptive names to reenforce the trend.

gapis/resolve/memory.go
```diff
@@ -227,20 +268,20 @@ func Memory(ctx context.Context, p *path.Memory, rc *path.ResolveConfig) (*servi
 	var reads, writes, observed memory.RangeList
 	typedRanges := []*service.TypedMemoryRange{}
 
-	s.Memory.SetOnCreate(func(id memory.PoolID, pool *memory.Pool) {
-		if id == memory.PoolID(p.Pool) {
-			pool.OnRead = func(rng memory.Range, root uint64, id uint64, apiId coreid.ID) {
+	s.Memory.SetOnCreate(func(poolID memory.PoolID, pool *memory.Pool) {
+		if poolID == memory.PoolID(p.Pool) {
+			pool.OnRead = func(rng memory.Range, root uint64, typeID uint64, apiId coreid.ID) {
```
purvisa-at-google-com, line 273: Glad to see you upgrading the descriptiveness of preexisting names here!

purvisa-at-google-com (approved):
LGTM. Optionally address feedback on variable naming

gapis/resolve/memory.go
yalcinmelihyasin: In this case, `type` is a keyword in Go, therefore we cannot use it. The usual word I see in the codebase is `typ`

yalcinmelihyasin (approved):

gapis/resolve/memory.go
ttanatb: 🎉

gapis/resolve/memory.go
ttanatb: Done, opted for more descriptive names `memType` and `typedRange`

purvisa-at-google-com:
Hey Tanat, this approved 5 days ago. All good to submit! Thanks for the work!

PR #978, opened 2021-11-17
Implement khr create renderpass2

yalcinmelihyasin:
Implements the VK_KHR_create_renderpass2

Also:

Refactor the RenderPass state and the old renderpass functions


pau-baiget:
How are we testing the newly added extension?

gapis/api/vulkan/api/command_buffer_control.api
```diff
@@ -619,13 +637,23 @@ sub void resetBufferCommands(ref!CommandBufferObject obj) {
...
+
   // @extension("VK_EXT_transform_feedback")
   clear(obj.BufferCommands.vkCmdBindTransformFeedbackBuffersEXT)
   clear(obj.BufferCommands.vkCmdBeginTransformFeedbackEXT)
   clear(obj.BufferCommands.vkCmdEndTransformFeedbackEXT)
   clear(obj.BufferCommands.vkCmdBeginQueryIndexedEXT)
   clear(obj.BufferCommands.vkCmdEndQueryIndexedEXT)
   clear(obj.BufferCommands.vkCmdDrawIndirectByteCountEXT)
+
+  // @extension("VK_KHR_create_renderpass2")
```
pau-baiget, line 653: Isn't vkCreateRenderPass2 also supposed to be included here?

gapis/api/vulkan/api/enums.api
```diff
@@ -283,6 +283,15 @@ enum VkStructureType: u32 {
   //@extension("VK_EXT_hdr_metadata")
   VK_STRUCTURE_TYPE_HDR_METADATA_EXT = 1000105000,
 
+  // extension("VK_KHR_create_renderpass2")
+  VK_STRUCTURE_TYPE_ATTACHMENT_DESCRIPTION_2 = 1000109000,
+  VK_STRUCTURE_TYPE_ATTACHMENT_REFERENCE_2 = 1000109001,
+  VK_STRUCTURE_TYPE_SUBPASS_DESCRIPTION_2 = 1000109002,
+  VK_STRUCTURE_TYPE_SUBPASS_DEPENDENCY_2 = 1000109003,
+  VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO_2 = 1000109004,
```
pau-baiget, line 291: Why is this in a different order than the spec? (Also alphabetically)

gapis/api/vulkan/api/structs.api
```diff
@@ -926,8 +926,8 @@ class VkAttachmentDescription {
 }
 
 class VkAttachmentReference {
-  u32           Attachment // We have capitalized these so we can manipulate
-  VkImageLayout Layout     // them in Go
+  u32           attachment
```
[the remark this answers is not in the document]
yalcinmelihyasin: This was needed because we needed to access the fields. (In Go's `Attachment` is public, `attachment` is private). As we don't need to access it outside the package, we do not need to have them public anymore.

gapis/api/vulkan/api/enums.api
yalcinmelihyasin: The order here does not matter. The important bit is the numbers. Lots of enums we have are in different order than the spec. If it's an issue(I don't think it is) we can have a one-off PR to clean them up. The general problem is that, we always add it to last when we add a new extension but that does not necessarily reflect the spec. I have tried to put it to somewhere that more or less where it should be.

gapis/api/vulkan/api/command_buffer_control.api
yalcinmelihyasin: No, because VkCreateRenderPass2(KHR) is not a buffer command.

Buffer command means "a command that is recorded to a command buffer, then executed when that buffer is submitted to a queue". Which in practice means the commands that starts with `VkCmd....` are the buffer commands.

yalcinmelihyasin:
> How are we testing the newly added extension?

Application tested. I have tested it with an old renderpass application with multiview extension and another application that uses vkCreateRenderPass2(KHR). This covers the edge cases in both worlds. I think our swarming tests would also be helpful to show any kind of regression.

Actually, we can add even more tests now because there were quite a bit chunk of sample applications we could not use before due to their dependency on VK_KHR_create_renderpass2

gapis/api/vulkan/extensions/khr_create_renderpass2.api
```diff
@@ -0,0 +1,138 @@
...
+}
+
+@extension("VK_KHR_create_renderpass2")
+@indirect("VkCommandBuffer", "VkDevice")
+cmd void vkCmdNextSubpass2KHR(
+    VkCommandBuffer                             commandBuffer,
+    const VkSubpassBeginInfo*                   pSubpassBeginInfo,
+    const VkSubpassEndInfo*                     pSubpassEndInfo) {
+    // TODO: Implement
+}
```
purvisa-at-google-com, line 200: Nit: newline at end of file

gapis/api/vulkan/transform_overdraw.go
```diff
@@ -484,14 +484,26 @@ func (overdrawTransform *stencilOverdraw) createRenderPass(ctx context.Context,
...
+			ad.InitialLayout(),
+			ad.FinalLayout(),
+		))
+	}
+	newAttachments = append(newAttachments, stencilAttachment)
+	newAttachmentsPtr := allocAndRead(newAttachments).Ptr()
 
 	stencilAttachmentReference := NewVkAttachmentReference(
-		uint32(len(attachments)),
+		uint32(len(newAttachments)-1),
```
purvisa-at-google-com, line 506: Was the lack of a -1 before a bug? Or is adding the -1 now a bug? Not sure if this is meant to be the number of elements in the array or the index of the last element.

gapis/api/vulkan/command_buffer_rebuilder.go
```diff
@@ -141,75 +141,125 @@ func allocateNewCmdBufFromExistingOneAndBegin(
...
+	d VkCmdBeginRenderPassCommonArgsʳ) (func(), api.Cmd, error) {
+	switch d.Version() {
+	case RenderPassVersion_Renderpass:
+		return rebuildVkCmdBeginRenderPass(ctx, cb, commandBuffer, r, s, d)
+	case RenderPassVersion_Renderpass2:
+		return rebuildVkCmdBeginRenderPass2(ctx, cb, commandBuffer, r, s, d)
+	case RenderPassVersion_Renderpass2KHR:
+		return rebuildVkCmdBeginRenderPass2KHR(ctx, cb, commandBuffer, r, s, d)
+	default:
+		panic("There should be a renderpass version")
```
purvisa-at-google-com, line 159: nit. there is a renderpass version, it's just not one we recognise. Better to say unknown renderpass version.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -589,6 +589,11 @@ sub ref!SubpassBeginInfo createSubpassBeginInfoFromInfo(const VkSubpassBeginInfo
   )
 }
 
+sub ref!SubpassEndInfo createSubpassEndInfo(const VkSubpassEndInfo* pSubpassEnd) {
+  _ = pSubpassEnd // this struct is empty and we do not support any extension for this struct
```
purvisa-at-google-com, line 593: This is done so the read/write observer callbacks are triggered for pSubpassEnd?

gapis/api/vulkan/command_buffer_rebuilder.go
```diff
@@ -141,6 +141,14 @@ func allocateNewCmdBufFromExistingOneAndBegin(
 	return newCmdBufID, x, cleanup
 }
 
+func createVkSubpassEndInfo(unused SubpassEndInfoʳ) VkSubpassEndInfo {
+	_ = unused // SubpassEndInfo is currently empty.
```
purvisa-at-google-com: ditto above, this is for the r/w callbacks?

gapis/api/vulkan/command_buffer_rebuilder.go
```diff
@@ -266,17 +274,93 @@ func rebuildVkCmdBeginRenderPass2KHR(
...
+	return cleanup, cmd, nil
+}
+
+func rebuildVkCmdEndRenderPass2KHR(
+	ctx context.Context,
+	cb CommandBuilder,
+	commandBuffer VkCommandBuffer,
+	r *api.GlobalState,
+	s *api.GlobalState,
+	d VkCmdEndRenderPassCommonArgsʳ) (func(), api.Cmd, error) {
```
purvisa-at-google-com, line 333: Might be nice to typedef func() to CleanupFunction or something, so it's more clear what the first return argument is giving you. Or rather, it's clearer what the purpose of the signatureless function is

purvisa-at-google-com: I've seen this in a few places so extend the comment to all of them too

gapis/api/vulkan/api/queued_command_tracking.api
```diff
@@ -267,7 +267,7 @@ sub void callCommand(ref!CommandReference reference) {
     case cmd_vkCmdBeginRenderPass:
       dovkCmdBeginRenderPassCommon(CommandBuffers[reference.Buffer].BufferCommands.vkCmdBeginRenderPass[reference.MapIndex])
     case cmd_vkCmdNextSubpass:
-      dovkCmdNextSubpass(CommandBuffers[reference.Buffer].BufferCommands.vkCmdNextSubpass[reference.MapIndex])
```
purvisa-at-google-com, line 270: Should we be checking these keys are present in the map and reporting errors where appropriate?

gapis/api/vulkan/command_buffer_rebuilder.go
```diff
@@ -361,14 +353,101 @@ func rebuildVkCmdEndRenderPass2KHR(
...
+	d VkCmdNextSubpassCommonArgsʳ) (func(), api.Cmd, error) {
+	switch d.Version() {
+	case RenderPassVersion_Renderpass:
+		return rebuildVkCmdNextSubpass(ctx, cb, commandBuffer, r, s, d)
+	case RenderPassVersion_Renderpass2:
+		return rebuildVkCmdNextSubpass2(ctx, cb, commandBuffer, r, s, d)
+	case RenderPassVersion_Renderpass2KHR:
+		return rebuildVkCmdNextSubpass2KHR(ctx, cb, commandBuffer, r, s, d)
+	default:
+		panic("There should be a renderpass version")
```
purvisa-at-google-com, line 371: Unknown render pass version*

purvisa-at-google-com (approved):
LGTM address nits and comments as you see fit

gapis/api/vulkan/transform_overdraw.go
yalcinmelihyasin: That's index of an element.

Difference is that previously, existing attachments would be copied to a local variable(`attachments`) and currently, we do not do that and copy to the `newAttachments` directly instead while iterating over them. And after the iteration, we add the new attachment to that list as well. that's why the -1 exists.

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: No, it's more for the consistency. As soon as we get a new extension that uses it we need to change the function signature. Rather than that, it's better to pass the argument and not use it.

gapis/api/vulkan/command_buffer_rebuilder.go
yalcinmelihyasin: same with above.

gapis/api/vulkan/command_buffer_rebuilder.go
yalcinmelihyasin: Well, that would require to change the entire file(because there is a pattern). We can still do that but not as a part of this PR.

gapis/api/vulkan/api/queued_command_tracking.api
yalcinmelihyasin: They should exist. If they do not exist, there is nothing we cannot do anything but panic. Because no user application can trigger that case.

And having a panic or a crash does not matter at that point. It might be cleaner to do the panic but again, there is an existing pattern and if we want to do that, we need to do it for the entire file and not part of this PR.

gapis/api/vulkan/command_buffer_rebuilder.go
yalcinmelihyasin: +1

gapis/api/vulkan/command_buffer_rebuilder.go
yalcinmelihyasin: +1

gapis/api/vulkan/transform_overdraw.go
purvisa-at-google-com: Worth a comment on why we're skipping the last element then!

gapis/api/vulkan/api/renderpass_framebuffer.api
purvisa-at-google-com: sounds good

gapis/api/vulkan/api/command_buffer_control.api
```diff
@@ -272,9 +280,9 @@ enum SemaphoreUpdate {
   @untrackedMap dense_map!(u32, ref!vkCmdWriteTimestampArgs)         vkCmdWriteTimestamp
   @untrackedMap dense_map!(u32, ref!vkCmdCopyQueryPoolResultsArgs)   vkCmdCopyQueryPoolResults
   @untrackedMap dense_map!(u32, ref!vkCmdPushConstantsArgs)          vkCmdPushConstants
-  @untrackedMap dense_map!(u32, ref!vkCmdBeginRenderPassArgs)        vkCmdBeginRenderPass
-  @untrackedMap dense_map!(u32, ref!vkCmdNextSubpassArgs)            vkCmdNextSubpass
-  @untrackedMap dense_map!(u32, ref!vkCmdEndRenderPassArgs)          vkCmdEndRenderPass
+  @untrackedMap dense_map!(u32, ref!vkCmdBeginRenderPassCommonArgs)  vkCmdBeginRenderPass
```
pmuetschard, line 283: Personally, I'd prefer something other than `Common`, because "common" implies that there are some non-common args as well. I'd just call it `vkCmdBeginRenderPassArgs`, or `vkCmdBeginRenderPassXArgs`.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -121,39 +121,118 @@ cmd void vkDestroyFramebuffer(
...
-  @unused VkPipelineBindPoint              PipelineBindPoint
-  @unused map!(u32, VkAttachmentReference) InputAttachments
-  @unused map!(u32, VkAttachmentReference) ColorAttachments
-  @unused map!(u32, VkAttachmentReference) ResolveAttachments
-  @unused ref!VkAttachmentReference        DepthStencilAttachment
-  @unused map!(u32, u32)                   PreserveAttachments
+@internal enum RenderPassVersion : u32 {
+  Renderpass = 0,
+  Renderpass2 = 1,
+  Renderpass2KHR = 2,
```
pmuetschard, line 127: Consistent capitalization. `RenderPass`, etc.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -121,39 +121,118 @@ cmd void vkDestroyFramebuffer(
...
 
 @internal class RenderPassObject {
   @unused VkDevice                           Device
   @unused VkRenderPass                       VulkanHandle
-  @unused map!(u32, VkAttachmentDescription) AttachmentDescriptions
-  @unused map!(u32, SubpassDescription)      SubpassDescriptions
-  @unused map!(u32, VkSubpassDependency)     SubpassDependencies
+  @unused map!(u32, AttachmentDescription2)  AttachmentDescriptions
+  @unused map!(u32, SubpassDescription2)     SubpassDescriptions
+  @unused map!(u32, SubpassDependency2)      SubpassDependencies
```
pmuetschard, line 135: `AttachmentDescription2`, `SubpassDescription2`, and `SubpassDependency2` are all "internal" objects, so please drop the `2`. Also keep them here rather than moving them to the extension file. If you want to can annotate the `ViewMask` field to be 1.2 only.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -121,39 +121,118 @@ cmd void vkDestroyFramebuffer(
...
 
-@internal class SubpassDescription {
-  @unused VkSubpassDescriptionFlags        Flags
-  @unused VkPipelineBindPoint              PipelineBindPoint
-  @unused map!(u32, VkAttachmentReference) InputAttachments
-  @unused map!(u32, VkAttachmentReference) ColorAttachments
-  @unused map!(u32, VkAttachmentReference) ResolveAttachments
-  @unused ref!VkAttachmentReference        DepthStencilAttachment
-  @unused map!(u32, u32)                   PreserveAttachments
+@internal enum RenderPassVersion : u32 {
```
pmuetschard, line 124: I would call it `kind`, `type`, or `flavor` rather than version.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -121,39 +121,118 @@ cmd void vkDestroyFramebuffer(
...
+  return AttachmentReference2(
+    Attachment: attachment.attachment,
+    Layout: attachment.layout,
+    AspectMask: 0, // This will be updated by VkRenderPassInputAttachmentAspectCreateInfo
+  )
+}
+
+sub ref!RenderPassObject createRenderPassObjectFromInfo(
+  const VkRenderPassCreateInfo* pCreateInfo,
+  VkDevice device) {
```
pmuetschard, line 154: add an extra indent

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -163,100 +242,210 @@ cmd VkResult vkCreateRenderPass(
...
+
+          for j in (0 .. ext.aspectReferenceCount) {
+            // The input attachment aspect info is only cached but not used for now, so the
+            // dependency can be wrong as a draw call may depends on just one aspect of a
+            // depth stencil image, as specified in the input attachment aspect info, while
+            // GAPID will still treat both aspects as dependees.
+            // TODO: Once the CL that enables us getting used descriptors from shader is in,
+            // we can pull input_attachment_index from the information of the used descriptors
+            // then use the input attachment aspect info to fix the dependency.
+            aspectReference := ext.pAspectReferences[0]
```
pmuetschard, line 254: shouldn't this use `[j]` rather than `[0]`. Looks to me like its processing the same one over and over again.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -163,100 +242,210 @@ cmd VkResult vkCreateRenderPass(
...
-          )
-
           if ext.pViewMasks != null {
-            renderPass.MultiviewInfo.PViewMasks = clone(
-              as!const u32*(ext.pViewMasks)[0:ext.subpassCount])
+            viewMasks := ext.pViewMasks[0:ext.subpassCount]
+            for j in (0 .. ext.subpassCount) {
+              desc := renderPass.SubpassDescriptions[j]
+              desc.ViewMask = viewMasks[j]
+              renderPass.SubpassDescriptions[j] = desc
```
pmuetschard, line 271: why not `renderPass.SubpassDescrptions[j].ViewMask = viewMasks[j]`? does that not compile or fail in some other way?

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -163,100 +242,210 @@ cmd VkResult vkCreateRenderPass(
...
-  renderPass.VulkanHandle = pRenderPass[0]
+
+  if info.pCorrelatedViewMasks != null {
+    pCorrelatedViewMasks := info.pCorrelatedViewMasks[0:info.correlatedViewMaskCount]
+    for i in (0 .. info.correlatedViewMaskCount) {
+      renderPass.CorrelatedViewMasks[i] = pCorrelatedViewMasks[i]
+    }
+  }
+
+  // We do not support any extension for this functionality yet.
```
pmuetschard, line 403: make sure pNext is null then, rather than just ignoring it.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -287,21 +476,32 @@ cmd void vkGetRenderAreaGranularity(
...
+
+  // @extension("VK_KHR_device_group_creation")
   ref!DeviceGroupBeginInfo DeviceGroupBeginInfo
 }
 
-sub void dovkCmdBeginRenderPass(ref!vkCmdBeginRenderPassArgs args) {
+@internal class
+vkCmdBeginRenderPassCommonArgs {
+  ref!RenderPassBeginInfo PRenderPassBeginInfo
+  ref!SubpassBeginInfo    PSubpassBeginInfo
```
pmuetschard, line 492: drop the `P`.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -340,6 +534,73 @@ sub void RecordSubpassBegin(ref!CommandBufferObject cb, u32 subpass) {
...
+
+sub ref!SubpassBeginInfo createSubpassBeginInfoFromContents(VkSubpassContents contents) {
+  return new!SubpassBeginInfo(
+    Contents: contents,
+  )
+}
+
+sub ref!SubpassBeginInfo createSubpassBeginInfoFromInfo(const VkSubpassBeginInfo* pSubpassBeginInfo) {
+  info := pSubpassBeginInfo[0]
+  // We do not support any extension for this functionality.
```
pmuetschard, line 586: Ensure `pNext` is null rather than ignoring it.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -340,6 +534,73 @@ sub void RecordSubpassBegin(ref!CommandBufferObject cb, u32 subpass) {
...
+  return new!SubpassBeginInfo(
+    Contents: info.contents,
+  )
+}
+
+sub ref!SubpassEndInfo createSubpassEndInfo(const VkSubpassEndInfo* pSubpassEnd) {
+  _ = pSubpassEnd // this struct is empty and we do not support any extension for this struct
```
pmuetschard, line 593: Ensure pNext is null rather than ignoring it.

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: 👍

gapis/api/vulkan/api/command_buffer_control.api
yalcinmelihyasin: As longs as we put something that indicates, this is not a usual Vulkan command, that's fine for me.

I am changing the common to x

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Well it's due to the version of the function. I don't know if they have a specific name for it in the spec but `kind` or `type` is actually misleading because it does **not** create a different kind of renderpass, it's just another function to create the same `kind` or `type` of renderpass. And I don't know what does flavor even mean in this context.

I think version is quite fine.

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: 👍

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Unfortunately, it does not compile. It is ugly and unnecessary but I had to do this. Maybe there is a way to fix this from the gapil but I have no idea how.

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Actually the original code does the same but I believe that the original line is incorrect as well. 

Original:

```
if ext.aspectReferenceCount != as!u32(0) {
            if ext.pAspectReferences == null {
              vkErrorNullPointer("VkInputAttachmentAspectReference(KHR)")
            } else {
              for j in (0 .. ext.aspectReferenceCount) {
                // The intput attachment aspect info is only cached but not used for now, so the
                // dependency can be wrong as a draw call may depends on just one aspect of a
                // depth stencil image, as specified in the input attachment aspect info, while
                // GAPID will still treat both aspects as dependees.
                // TODO: Once the CL that enables us getting used descriptors from shader is in,
                // we can pull input_attachment_index from the information of the used descriptors
                // then use the input attachment aspect info to fix the dependency.
                renderPass.InputAttachmentAspectInfo.AspectReferences[j] = ext.pAspectReferences[0]
              }
            }
          }
```

Probably what should have been:

```
renderPass.InputAttachmentAspectInfo.AspectReferences[j] = ext.pAspectReferences[j]
```

And in our case it should be:

```
aspectReference := ext.pAspectReferences[j]
```

Let me know what do you think about this.

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Well we affectively don't do that anywhere in the code.

This is our usual unhandled pNext code which can be found in many places:

```
if info.pNext != null {
    numPNext := numberOfPNext(info.pNext)
    next := MutableVoidPtr(as!void*(info.pNext))
    for i in (0 .. numPNext) {
      sType := as!const VkStructureType*(next.Ptr)[0]
      switch (sType) {
      }
      next.Ptr = as!VulkanStructHeader*(next.Ptr)[0].PNext
    }
```

This does not do anything but iterating over pNext while doing nothing. In reality, we shouldn't let any application with an unknown pNext to be traced and if they do we don't guarantee anything anyway.

Maybe it is actually a good idea to log or return error on an unknown pNext but that is out of the scope of this PR.

If you think that is necessary, we can go over every pNext and ensure that we do the proper handling when we do not support them.  Otherwise it would be inconsistent and confusing to do it for a particular function only.

And between and empty loop and no code  at all, I prefer 'no code'.

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Well, I can't because typename is the same with the field name without P.

yalcinmelihyasin: 👍

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Same comment with previous

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Same comment with previous

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: I believe this is not a very good idea. Because we have exactly a `VkSubpassDescription2` structure that corresponds to the internal structure while the `VkSubpassDescription` does not reflect it.

We will have more of these in the future with Vulkan creating more of these and I would like to keep consistency.

If you want, we can say that there is no reflection between the spec struct and internal Struct and in that case I am ok with something like `VkSubpassDescriptionX` as well.

yalcinmelihyasin: @pmuetschard

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: @pmuetschard

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: @pmuetschard

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: @pmuetschard

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: @pmuetschard

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: @pmuetschard

gapis/api/vulkan/api/renderpass_framebuffer.api
pmuetschard: The difference between `vk<Something>` and `vk<something>2` is usually just the addition of the `pNext` pointer, which we don't put into our internal structs anyways. There's not going to be many more `vk<something>2` as there's not much, if anything, left that doesn't have a `pNext`, there certainly won't be any `vk<something>3` - Khronos have learnt their lesson. I'd rather not follow this crazy `<something>2` pattern, especially since we do not have the `<something>` beside it. It'd be a different story if we had both. Also, when extensions show up that add structs to the `pNext` chain, we will adjust the `<something>` structs and add fields to them that are not part of the `vk<something>` and so they won't match anymore.

gapis/api/vulkan/api/renderpass_framebuffer.api
pmuetschard: I would say they are different types, which is why we have this field, but I'm OK to keep `version` if you feel strongly about it. I was suggesting something different, because to me version made it sound like you could create "another version of this renderpass" and not so much "you used this version of the `vkCreateRenderPass` function to create it".

gapis/api/vulkan/api/renderpass_framebuffer.api
pmuetschard: Yes, please fix it. Just because it was broken, doesn't mean we should keep it broken :).

gapis/api/vulkan/api/renderpass_framebuffer.api
pmuetschard: Thanks, I suspected as much. This is probably to make the memory read/write tracking happy.

gapis/api/vulkan/api/renderpass_framebuffer.api
pmuetschard: Melih, it does something. It goes through the whole chain and will observe at least the `sType` and `pNext` field of every struct during tracing. On mutate it will produce a nice error message containing which struct type we are not supporting, because the switch statement is missing a `default` case. We used to just ignore them, but that was bad. If you want you can add a comment inside the `switch`: `// Nothing supported yet, but ensures the pNext gets observed.`.

yalcinmelihyasin: I am sorry but it does not.

This is the generated code for those line(the function is `VkCreateQuery`) there is no error message here. As far as I know, we don't have nice error messages for trace at all.

```

if ((l_info.mpNext) != (nullptr)) {
                uint32_t l_numPNext = numberOfPNext(observer, l_info.mpNext);
                MutableVoidPtr l_next = MutableVoidPtr((void*)(l_info.mpNext));
                for (uint32_t l_i = static_cast<uint32_t>(0U); l_i < l_numPNext; ++l_i) {
                    uint32_t l_sType = observer->read(slice((uint32_t const*)(l_next.mPtr), static_cast<uint64_t>(0U), static_cast<uint64_t>(1U)), static_cast<uint64_t>(0U));
                    switch (l_sType) {
                    }
                    l_next.mPtr = observer->read(slice((VulkanStructHeader*)(l_next.mPtr), static_cast<uint64_t>(0U), static_cast<uint64_t>(1U)), static_cast<uint64_t>(0U)).mPNext;
                }
            }
```

Only in replay we do have a default case. But the way we do it, does not make any sense. No one reading that code can predict that we generate a default case(which we should not btw, we are generating something that contradicts with what's written) and we will do the error check there. There are much nicer ways to do this. For now, I will just check PNext and if it's not null, I'll return `vkErrUnsupported`

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: I agree that what you say is correct but I think none of the options here makes it any better. Because for example, using "type" may imply that if you create a `RenderPass` with `VkCreateRenderPass2` you cannot use `VkBeginRenderPass`  or vice versa but it's not true. This is a completely valid use case:

`VkCreateRenderPass2`
`VkBeginRenderPass`
`VkNextSubpass2`
`VkEndRenderPass2KHR`

So, the RenderPass we create or use are the exact same RenderPass in all of these functions, it's just the API difference. 

But I agree that it is hard to state this clearly.

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: My question was more on the line that "Are we sure that it's broken" or is this intended. 

I'll try to go over the line of the code again with detail and see if it actually does the correct thing. Probably it does not I you said but I want to be sure and not break it while thinking that we are fixing it.

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Please check the pNext commit

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Please check the pNext commit

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Please check the pNext commit

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: Please see the `rename internal structs` commit. Done!

gapis/api/vulkan/api/renderpass_framebuffer.api
yalcinmelihyasin: I've fixed this and I'll keep it as a separate commit without squashing, just in case.

gapis/api/vulkan/api/renderpass_framebuffer.api
pmuetschard: In our API language `switch` is an expression, not a statement, so it needs a "result value". I think in the C++ code generator the result value is the default value if none is specified, but in the Go generator it's considered a runtime-error to not have a value. This is because we don't really have abort/errors in C++ as we try to do our best to capture everything - trying to recover from error conditions, so you can inspect things when there are issues. On the GAPIS side we can be more strict and abort. Please don't be so quick to dismiss everything as bad or wrong, there is no contradiction here, things are just a bit more complex than you assumed. Of course there's always room to improve things.

pmuetschard (commented):
Finished looking at everything now. If we still abort on the GAPIS side and show an error if there is an unsupported pNext, then it's ready to submit, though I would prefer splitting out the new error stuff into another PR that then uses it everywhere, but am OK to keep it here.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -163,100 +287,216 @@ cmd VkResult vkCreateRenderPass(
...
+      renderPass.CorrelatedViewMasks[i] = pCorrelatedViewMasks[i]
+    }
+  }
+
+  if info.pNext != null {
+    // We do not support any extension for this functionality.
+    next := MutableVoidPtr(as!void*(info.pNext))
+    sType := as!const VkStructureType*(next.Ptr)[0]
+    vkErrorUnrecognizedExtensionPNext(sType)
+  }
```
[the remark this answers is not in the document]
yalcinmelihyasin: if you look at the generated code in the `mutate.go` You can see that the existing code has exactly the same behavior.

I have actually thought about having a loop to check it but with the way the Vulkan release works, it's very unlikely that we will get two new extensions that are unknown and used at the same time. Therefore I wanted to keep it consistently incomplete error handling.

But if you feel strong about it, I can change. In that case these pNext checks will be the only ones that are able to do it and it may confuse us in the future about which missing pNext are reported all and which are repored with only the first one.

For your suggestion, to be fair, I don't know if that will be compiled and run as we expect. My dealing with gapil is I look at how everything else done and follow them. If you want I can try to make it one line but I don't know if there is any benefit of risking that we get wrong behavior.

pmuetschard (approved):

PR #982, opened 2021-11-19
Fix a crash in VkCmdBindPipeline 

yalcinmelihyasin:
when it's called outside of a renderpass

b/200873355

### DIFF
--- a/gapis/api/vulkan/transform_command_splitter.go
+++ b/gapis/api/vulkan/transform_command_splitter.go
@@ -672,16 +672,18 @@ func (splitTransform *commandSplitter) splitCommandBuffer(ctx context.Context, e
 		case VkCmdBindPipelineArgsʳ:
 			if ar.PipelineBindPoint() == VkPipelineBindPoint_VK_PIPELINE_BIND_POINT_GRAPHICS {
 				// Graphics pipeline, must be split (maybe)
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/transform_command_splitter.go
```diff
@@ -672,16 +672,18 @@ func (splitTransform *commandSplitter) splitCommandBuffer(ctx context.Context, e
 		case VkCmdBindPipelineArgsʳ:
 			if ar.PipelineBindPoint() == VkPipelineBindPoint_VK_PIPELINE_BIND_POINT_GRAPHICS {
 				// Graphics pipeline, must be split (maybe)
-				if st.RenderPasses().Get(splitTransform.thisRenderPass.RenderPass()).SubpassDescriptions().Len() > 1 {
-					// If we have more than one renderpass, then we should replace
-					newPipeline, err := splitTransform.rewriteGraphicsPipeline(ctx, ar.Pipeline(), queueSubmit, inputState)
-					if err != nil {
-						log.E(ctx, "Failed during rewriting graphics pipeline : %v", err)
-						return VkCommandBuffer(0), err
+				if splitTransform.thisRenderPass != NilVkCmdBeginRenderPassArgsʳ {
```
chrisforbes, line 675: What happens if a multiple-subpass renderpass is subsequently started? The pipeline will ultimately need to be adjusted to work with the post-split version of that renderpass, but it looks like that doesn't happen?

PR #984, opened 2021-11-24
Convert uint8 to proper char value

ttanatb:
Screen Shot 2021-11-23 at 5 50 25 PM

### DIFF
--- a/core/data/pod/pod.proto
+++ b/core/data/pod/pod.proto
@@ -20,6 +20,7 @@ option java_outer_classname = "Pod";
 option go_package = "github.com/google/gapid/core/data/pod";
 
 message Value {
+  // Next tag: 30
   oneof val {
     float float32 = 1;
     double float64 = 2;
@@ -49,6 +50,9 @@ message Value {
     Sint64Array sint64_array = 26;
     BoolArray bool_array = 27;
     StringArray string_array = 28;
+
+    // C++ Types
+    uint32 char = 29;
   }
 }
 
--- a/core/data/pod/pod.proto
+++ b/core/data/pod/pod.proto
@@ -20,7 +20,6 @@ option java_outer_classname = "Pod";
 option go_package = "github.com/google/gapid/core/data/pod";
 
 message Value {
-  // Next tag: 30
   oneof val {
     float float32 = 1;
     double float64 = 2;
--- a/gapic/src/main/com/google/gapid/util/Pods.java
+++ b/gapic/src/main/com/google/gapid/util/Pods.java
@@ -63,6 +63,11 @@ public class Pods {
   public static StringBuilder append(StringBuilder sb, Pod.Value v) {
     switch (v.getValCase()) {
       case VAL_NOT_SET: return sb.append("[null]");
+      case CHAR: {
+        int charValue = v.getChar();
+        if (charValue == 0) return sb.append("\\0");
+        return sb.append((char)charValue);
+      }
       case STRING: return sb.append(v.getString());
       case BOOL: return sb.append(v.getBool());
       case FLOAT64: return sb.append(v.getFloat64());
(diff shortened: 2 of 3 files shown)


gapis/resolve/memory.go
```diff
@@ -195,38 +196,49 @@ func expandCharArrays(ctx context.Context, typedRanges []*service.TypedMemoryRan
...
 		}
 	}
 	return typedRanges, nil
 }
 
-// valAsString combines the values in the array into one string enclosed by quotation marks (")
-func valAsString(values []*memory_box.Value) string {
+// valAsString converts the int values to char and generates a string that omits the
+// null terminator and is enclosed by quotation marks (")
+func valAsString(values []*memory_box.Value) (*pod.Value, []*memory_box.Value) {
```
yalcinmelihyasin, line 213: I think this function should be split. 

Now it does two different things.

One returns a value as pod.Value(not as string anymore)
The other one, modifies the argument.

yalcinmelihyasin (changes requested):

gapic/src/main/com/google/gapid/util/Pods.java
```diff
@@ -63,6 +63,11 @@ public static Object unpod(Pod.Value o) {
   public static StringBuilder append(StringBuilder sb, Pod.Value v) {
     switch (v.getValCase()) {
       case VAL_NOT_SET: return sb.append("[null]");
+      case CHAR: {
+        int charValue = v.getChar();
+        if (charValue == 0) return sb.append("\\0");
```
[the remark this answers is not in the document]
ttanatb: I'm not sure if there's a cleaner way to do this, but I felt that "\0" is more descriptive than " " (easily mistaken for space)

core/data/pod/pod.proto
```diff
@@ -20,6 +20,7 @@ option java_outer_classname = "Pod";
 option go_package = "github.com/google/gapid/core/data/pod";
 
 message Value {
+  // Next tag: 30
```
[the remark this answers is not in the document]
ttanatb: It's a more common pattern in the rest of the code base at Google (especially helpful when tags are out of order and the presubmit and the IDE helps check against this), but it makes sense to omit this in the repo

gapis/resolve/memory.go
```diff
@@ -195,38 +196,49 @@ func expandCharArrays(ctx context.Context, typedRanges []*service.TypedMemoryRan
...
 	sb.Grow(len(values) + 1)
 	sb.WriteString("\"")
-	for _, v := range values {
+	for i, v := range values {
 		char := v.GetPod().GetUint8()
+		values[i] = &memory_box.Value{
+			Val: &memory_box.Value_Pod{
+				Pod: &pod.Value{
+					Val: &pod.Value_Char{
+						Char: char,
```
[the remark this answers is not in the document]
ttanatb: Done

gapis/resolve/memory.go
ttanatb: Done

gapis/resolve/memory.go
```diff
@@ -238,7 +229,17 @@ func valAsString(values []*memory_box.Value) (*pod.Value, []*memory_box.Value) {
 		Val: &pod.Value_String_{
 			String_: sb.String(),
 		},
-	}, values
+	}
+}
+
+// charsFromValues converts the int values in the array to a char array
+func charsFromValues(values []*memory_box.Value) []*memory_box.Value {
```
yalcinmelihyasin, line 236: I think that this still modifies the argument.

I want to make sure that I understand correctly. 

This function takes a pointer to an array of `value` struct and returns an array of `value` struct (not a `char` array for some reason).

But while doing it it also modifies the argument and returns it. Is that correct? 

I am afraid that it's not a very good naming in this case because, if I would call a `charsFromValues`, I would expect to get a `char` array from values and values to stay **unmodified**.

I don't know why we need the value struct so I cannot say anything about it.

Assuming that the `value` struct is needed, maybe we should rename this function as `convertValueTypeToChar` or something similar that does return void(or it can still return an array of `value` struct, if we are doing some function chaining but I cannot see that it's happening).

yalcinmelihyasin (changes requested):

gapis/resolve/memory.go
ttanatb: Renamed to convertValueTypeToChar

yalcinmelihyasin (approved):

PR #990, opened 2021-12-06
Fix a regression introduced by #950.

pmuetschard:
The map of divisors is not a dense map. The existing map unpacking
functions require dense maps as documented. This change introduces a new
function that will properly unpack a non-dense map, ignoring the keys.


mark-lunarg (commented):
Confirmed that this fixes the panic on the Ludo King application on a Pixel 4.

yalcinmelihyasin (approved):

PR #991, opened 2021-12-07
Fix an issue with an update on a disposed widget.

pmuetschard:
If the framebuffer attachment drop down is closed before the thumbnail finished loading, we end up calling an update function on a disposed widget at 60Hz.

Bug: http://b/204233029


mark-lunarg (approved):
Confirmed this fixes the issue.  Was also looking at [this bug](https://issuetracker.google.com/issues/204233029), and will move onto the next.

PR #992, opened 2021-12-10
Add validation info to gapis

ttanatb:
- Always download perfetto trace from validation
- Include path to trace in result proto
- Separate error from validation & error from downloading trace file


gapis/service/service.go
```diff
@@ -190,7 +190,7 @@ type Service interface {
 
 	// ValidateDevice validates the GPU profiling capabilities of the given device and returns
 	// an error if validation failed or the GPU profiling data is invalid.
-	ValidateDevice(ctx context.Context, d *path.Device) error
+	ValidateDevice(ctx context.Context, d *path.Device) (*ValidateDeviceResponse, error)
```
pmuetschard, line 193: Instead of returning the proto here, could you please create a go struct here that contains the relevant validation results (at this point just the trace file name) and then pack it up into the response proto in `grpc.go`. We haven't been very consistent, but I think it's preferable to not leak the `*Request` and `*Response` protos outside `grcp.go`.

Something like:
```
type DeviceValidation struct {
  ..
}
//..
ValidateDevice(..) (*DeviceValidation, error)
```

gapis/service/service.proto
```diff
@@ -1481,7 +1481,14 @@ message ValidateDeviceRequest {
 }
 
 message ValidateDeviceResponse {
+  // Error in validating device.
   Error error = 1;
+
+  // Error from downloading trace file.
+  Error download_error = 2;
```
pmuetschard, line 1488: What does "download" mean in this context and why is it separate from the regular trace error?

gapis/trace/android/trace.go
```diff
@@ -65,7 +66,7 @@ const (
 	durationMs                              = 7000
 	gpuCountersDataSourceDescriptorName     = "gpu.counters"
 	gpuRenderStagesDataSourceDescriptorName = "gpu.renderstages"
-	minimumSupportedApiLevel                = 29
+	minimumSupportedApiLevel                = 19
```
pmuetschard, line 69: Are you sure about that? I don't think we support Perfetto before Q.

gapis/trace/android/trace.go
```diff
@@ -215,32 +235,65 @@ func (t *androidTracer) Validate(ctx context.Context) error {
 
 		startFunc(ctx)
 		if !doneSignal.Wait(ctx) {
-			err = log.Err(ctx, err, "Fail to wait for done signal from Perfetto.")
+			err = log.Err(ctx, err, "Failed to wait for done signal from Perfetto.")
+			timeout = true
 		}
 		log.I(ctx, "Perfetto trace size %v bytes", written)
 	})
+	if timeout == true {
```
pmuetschard, line 243: Can you just use `if err != nil` here and not need the additional boolean?

gapis/trace/android/trace.go
```diff
@@ -215,32 +235,65 @@ func (t *androidTracer) Validate(ctx context.Context) error {
...
+		return &service.ValidateDeviceResponse{
+			Error: service.NewError(err),
+		}
+	}
+
+	// Attempt to download perfetto trace.
+	res := &service.ValidateDeviceResponse{}
+	temp, err := file.TempWithExt("validation", "perfetto")
```
pmuetschard, line 251: We should only do this and write to the file, if `gapis` was started with the `--enable-local-files` flag. The flag is passed down in the server so you can add it as a parameter to this function.

gapis/trace/android/trace.go
ttanatb: I was using this to test, UI stuff -- didn't mean to check this in

gapis/service/service.go
ttanatb: Done

gapis/service/service.proto
ttanatb: Wanted to separate actual validation error from one that came from temp file creation or copying over the perfetto trace. It felt weird that failing to create temp file or copy perfetto data would cause validation to fail.

gapis/trace/android/trace.go
ttanatb: Done

gapis/trace/android/trace.go
ttanatb: Done

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -446,7 +446,7 @@ public DeviceValidationResult(Service.Error error, boolean passed, boolean skipp
     }
 
     public DeviceValidationResult(Service.ValidateDeviceResponse r) {
-      this(r.getError(), !r.hasError(), false);
+      this(r.getError(), !r.hasError() && r.getResult().getValidationErrorMsg().length() == 0, false);
```
yalcinmelihyasin, line 449: Same comment with above.

cmd/gapit/validate_gpu_profiling.go
```diff
@@ -51,11 +51,15 @@ func (verb *validateGpuProfilingVerb) Run(ctx context.Context, flags flag.FlagSe
 	someDeviceFailed := false
 	for i, p := range devices {
 		fmt.Fprintf(stdout, "-- Device %v: %v --\n", i, p.ID.ID())
-		err = client.ValidateDevice(ctx, p)
+		res, err := client.ValidateDevice(ctx, p)
 		if err != nil {
 			fmt.Fprintf(stdout, "%v\n", log.Err(ctx, err, "Failed to validate device"))
 			someDeviceFailed = true
 			continue
+		} else if len(res.ValidationErrorMsg) > 0 {
```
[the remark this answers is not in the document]
pmuetschard: There are two kinds of "errors" here: 1) if we fail to run the validation for some reason (RPC, device communication issues, etc.) - e.g. we failed to run the validation - and 2) we successfully ran validation, but the device did not meet the checks and thus failed validation. Maybe instead of `ValidationErrorMsg` a better name for the field could be `ValidationFailureMsg`.

Now, "validation failure" could be made into an error type, yes, but I argued for it to be separate, because we have an established pattern, where RPCs return either a result or an error with the error indicating that something went wrong or the request was invalid. For validation, I see a "validation didn't pass" as a successful result to the "validate this device" request. Furthermore, reporting validation failures as errors would expand the error namespace too much.

I think the future plan is to also make the validation result more actionable in the UI. For example, we should probably give different advice if the device failed validation because it's too old (can't do anything about that) vs. when some counter failed to report the right data (try again, ensure the device is not otherwise busy, etc).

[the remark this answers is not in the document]
pmuetschard: To add: currently the validator returns a go `error` if validation doesn't pass. I think that will need to be changed in the future and it will return a `ValidationResult` - whatever that will look like - or an `error`.

cmd/gapit/validate_gpu_profiling.go
```diff
@@ -51,11 +51,15 @@ func (verb *validateGpuProfilingVerb) Run(ctx context.Context, flags flag.FlagSe
...
+			fmt.Fprintf(stdout, "%v\n", log.Errf(ctx, nil, "Failed to validate device: %s", res.ValidationErrorMsg))
```
pmuetschard, line 60: We should show different messages here for the two error cases. Maybe `Failed to validate device: <error>` and `Device did not pass validation: <msg>, trace file: <trace file>`.

Also, no need to turn the string into an error to then turn it back into a string.

gapic/src/main/com/google/gapid/models/Devices.java
pmuetschard, line 449: Just a note for the future: when you update the UI to show more detail, I think we should change `Client.java` to not leak the `*Response` message and turn the error into an exception like we do for other requests. But for this PR, this is OK.

gapis/service/service.proto
```diff
@@ -1481,7 +1481,19 @@ message ValidateDeviceRequest {
...
-  Error error = 1;
+  reserved 1;
+  oneof res {
+    DeviceValidationResult result = 2;
+    Error error = 3;
+  }
+}
+
+message DeviceValidationResult {
+  // Error from validiting the device.
```
pmuetschard, line 1492: Make this comment be more explicit that we are talking about a device that didn't pass validation rather than we somehow failed to run validation. Maybe "Message with details of how the device didn't pass validation." or something like it.

gapis/trace/android/trace.go
```diff
@@ -150,45 +151,52 @@ func (t *androidTracer) ProcessProfilingData(ctx context.Context, buffer *bytes.
...
 	if t.v == nil {
-		return log.Errf(ctx, nil, "No validator found for device %d", t.b.Instance().ID.ID())
+		return nil, log.Errf(ctx, nil, "No validator found for device %d", t.b.Instance().ID.ID())
 	}
 	d := t.b.(adb.Device)
 	osConfiguration := d.Instance().GetConfiguration()
 	if osConfiguration.GetOS().GetAPIVersion() < minimumSupportedApiLevel {
-		return log.Errf(ctx, nil, "Unsupported OS version on device %d", d.Instance().ID.ID())
+		return &service.DeviceValidationResult{
+			ValidationErrorMsg: fmt.Sprintf("Unsupported OS version on device %d", d.Instance().ID.ID()),
```
pmuetschard, line 165: We should include the device's version and the minimum version in the message.

gapis/trace/android/trace.go
```diff
@@ -150,45 +151,52 @@ func (t *androidTracer) ProcessProfilingData(ctx context.Context, buffer *bytes.
...
 	if osConfiguration.GetPerfettoCapability() == nil {
-		return log.Errf(ctx, nil, "No Perfetto Capability found on device %d", d.Instance().ID.ID())
+		return &service.DeviceValidationResult{
+			ValidationErrorMsg: fmt.Sprintf("No Perfetto Capability found on device %d", d.Instance().ID.ID()),
+		}, nil
 	}
 	if gpuProfiling := osConfiguration.GetPerfettoCapability().GetGpuProfiling(); gpuProfiling == nil || gpuProfiling.GetGpuCounterDescriptor() == nil {
-		return log.Errf(ctx, nil, "No GPU profiling capabilities found on device %d", d.Instance().ID.ID())
+		return &service.DeviceValidationResult{
+			ValidationErrorMsg: fmt.Sprintf("No GPU profiling capabilities found on device %d", d.Instance().ID.ID()),
```
pmuetschard, line 175: "The device does not support GPU profiling or profiling support was not detected".

gapis/trace/android/trace.go
```diff
@@ -215,32 +224,62 @@ func (t *androidTracer) Validate(ctx context.Context) error {
...
+	if enableLocalFiles {
+		temp, err = file.TempWithExt("validation", "perfetto")
+		if err != nil {
+			return nil, log.Err(ctx, err, "Failed to create a temp file for trace")
+		}
 	}
 
 	var processor *perfetto.Processor
+	var traceLoadingErr error
+	res := &service.DeviceValidationResult{}
```
pmuetschard, line 246: This is not actually needed inside the closure below, so I think it would be clearer to move the construction of it below the `status.Do` call. You can then initialize the path field, when creating it:
```
res := &service.DeviceValidationResult{
  TracePath: temp.System(),
}
```

pmuetschard (commented):
Looks mostly good to me, just minor comments.

cmd/gapit/validate_gpu_profiling.go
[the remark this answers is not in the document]
yalcinmelihyasin: in that case, we should at least ensure that we do this consistently in our RPCs and it's somehow explicit about the need to check the result.

Maybe it already is. I haven't checked what the actual rpc is carrying. But if we have a clear error case in RPCs and this is consistent amongst the code, then I am ok with it.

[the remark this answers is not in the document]
ttanatb: Yes, the other RPC result protos in the [service.proto](https://github.com/google/agi/blob/master/gapis/service/service.proto#L1336) include a similar structure
```
oneof res {
  Message result;
  Error err;
}
```

gapis/trace/android/trace.go
```diff
@@ -215,32 +224,62 @@ func (t *androidTracer) Validate(ctx context.Context) error {
...
+	if traceLoadingErr != nil {
+		return nil, traceLoadingErr
 	}
 	defer processor.Close()
-	ctx = status.Start(ctx, "Validation")
 	defer status.Finish(ctx)
-	return t.v.Validate(ctx, processor)
+
+	err = t.v.Validate(ctx, processor)
+	if err != nil {
```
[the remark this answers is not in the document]
ttanatb: Resolving this since it's answered by the thread in the first comment

gapis/trace/android/trace.go
```diff
@@ -215,32 +224,62 @@ func (t *androidTracer) Validate(ctx context.Context) error {
...
-		return err
+	res := &service.DeviceValidationResult{
+		TracePath: temp.System(),
 	}
-	defer processor.Close()
 	ctx = status.Start(ctx, "Validation")
+	if traceLoadingErr != nil {
+		return nil, traceLoadingErr
+	}
+	defer processor.Close()
```
pmuetschard, line 274: These two defers are now in the wrong place. the `defer processor.Close()` should be right after the first error check after the processor is created, line 252. `defer status.Finish(ct)` should be right after the `status.Start` line, otherwise we'd not call finish if `traceLodingErr` is not nil. Maybe move that check to before the `status` stuff?

pmuetschard (approved):

PR #997, opened 2021-12-15
Improve Command Tree Filtering

pmuetschard:
This contains some some of the followups from #935:
- Don't show the "Host Coordination" group if it's empty.
- Allow filtering out the submit info nodes.
- Don't group by frame if there's only one.

### DIFF
--- a/gapis/api/cmd_id_group.go
+++ b/gapis/api/cmd_id_group.go
@@ -602,6 +602,22 @@ func (g *CmdIDGroup) Cluster(maxChildren, maxNeighbours uint64) {
 	}
 }
 
+// Flatten replaces this node's children with its grandchildren.
+func (g *CmdIDGroup) Flatten() {
+	var newSpans Spans
+	for _, s := range g.Spans {
+		switch v := s.(type) {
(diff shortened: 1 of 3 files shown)


gapis/api/cmd_id_group.go
```diff
@@ -602,6 +602,22 @@ func (g *CmdIDGroup) Cluster(maxChildren, maxNeighbours uint64) {
 	}
 }
 
+// Flatten replaces this node's children with its grandchildren.
+func (g *CmdIDGroup) Flatten() {
+	var newSpans Spans
+	for _, s := range g.Spans {
+		switch v := s.(type) {
+		case *SubCmdRoot:
+			newSpans = append(newSpans, v.SubGroup.Spans...)
```
pau-baiget, line 611: Is this correct? It seems v is the type, not a Span (maybe i'm out of depth here). Shouldn't this be ```s.SubGroup.Spans``` ?

pau-baiget: Got explanation - Go shenanigans! :D

pau-baiget (approved):

PR #999, opened 2021-12-15
Automatically expand some nodes in the command tree.

pmuetschard:
This will automatically expand command tree nodes down to the renderpass level.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/CommandTree.java
+++ b/gapic/src/main/com/google/gapid/views/CommandTree.java
@@ -313,6 +313,12 @@ public class CommandTree extends Composite
         }
 
         @Override
+        protected boolean isDefaultExpanded(CommandStream.Node element) {
+          Service.CommandTreeNode data = element.getData();
+          return data != null && data.getExpandByDefault();
+        }
+
+        @Override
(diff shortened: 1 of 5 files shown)


gapic/src/main/com/google/gapid/views/CommandTree.java
```diff
@@ -312,6 +312,12 @@ protected boolean isLoaded(CommandStream.Node element) {
           return element.getData() != null;
         }
 
+        @Override
+        protected boolean isDefaultExpanded(CommandStream.Node element) {
+          Service.CommandTreeNode data = element.getData();
+          return data != null && data.getExpandByDefault();
```
pau-baiget, line 318: We always check if the node is expanded by default even though we're hardcoding true for every node in the command tree in the Go file. I guess we want to have the freedom to control this from the Go side in the future?

pau-baiget (approved):

gapic/src/main/com/google/gapid/views/CommandTree.java
pmuetschard: We only hard code it for `SubCmdRoot` nodes, not any of the other grouping nodes. So some groups will indeed have false for the UI (Command buffers and below).

PR #1000, opened 2021-12-15
Fix command -> tree node lookup for filtered trees.

pmuetschard:
The lookup made assumptions about the tree layout that are no longer correct when the tree is filtered.

### DIFF
--- a/gapis/resolve/command_tree.go
+++ b/gapis/resolve/command_tree.go
@@ -72,30 +72,28 @@ func (t *commandTree) index(indices []uint64) (api.SpanItem, api.SubCmdIdx) {
 	return group, subCmdRootID
 }
 
-func (t *commandTree) indices(idx []uint64, preferGroup bool) []uint64 {
+func (t *commandTree) indices(cmdIdx []uint64, preferGroup bool) []uint64 {
 	out := []uint64{}
 	group := t.root
 
-	for _, id := range idx {
-		brk := false
-		for {
-			if brk {
-				break
+	for i := 0; i < len(cmdIdx); i++ {
+		cmdId := api.CmdID(cmdIdx[i])
+		idx := group.IndexOf(cmdId)
+		out = append(out, idx)
+
+		switch item := group.Index(idx).(type) {
+		case api.CmdIDGroup:
+			group = item
+			if group.Bounds().Last() > cmdId {
+				// We are looking for something inside this group, not the group itself, so go around again.
+				i--
 			}
-			i := group.IndexOf(api.CmdID(id))
-			out = append(out, i)
-			switch item := group.Index(i).(type) {
-			case api.CmdIDGroup:
-				group = item
-				if preferGroup {
-					brk = true
-				}
-			case api.SubCmdRoot:
-				group = item.SubGroup
-				brk = true
-			default:
-				return out
-			}
+		case api.SubCmdRoot:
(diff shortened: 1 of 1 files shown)


gapis/resolve/command_tree.go
```diff
@@ -72,30 +72,28 @@ func (t *commandTree) index(indices []uint64) (api.SpanItem, api.SubCmdIdx) {
 	return group, subCmdRootID
 }
 
-func (t *commandTree) indices(idx []uint64, preferGroup bool) []uint64 {
+func (t *commandTree) indices(cmdIdx []uint64, preferGroup bool) []uint64 {
```
pau-baiget, line 75: This is me trying to understand the code below.

Do I understand correctly that ```cmdIdx``` is a list of indices that indicate the path to a given command and what this function is doing is actually finding the list of positions within the corresponding group (either CmdIdGroup or SubCmdRoot) for each of the indices? How do we ensure that ```cmdIdx``` always finishes in a leaf node? Also, what happens if we end the loop and we're still in some intermediate group?

gapis/resolve/command_tree.go
```diff
@@ -72,30 +72,28 @@ func (t *commandTree) index(indices []uint64) (api.SpanItem, api.SubCmdIdx) {
...
+		cmdId := api.CmdID(cmdIdx[i])
+		idx := group.IndexOf(cmdId)
+		out = append(out, idx)
+
+		switch item := group.Index(idx).(type) {
+		case api.CmdIDGroup:
+			group = item
+			if group.Bounds().Last() > cmdId {
+				// We are looking for something inside this group, not the group itself, so go around again.
+				i--
```
pau-baiget, line 89: OK so this is confusing. So the cmdIds within the tree are ordered? How is decreasing ```i``` going to get you inside the group?

pau-baiget (approved):

gapis/resolve/command_tree.go
pmuetschard: `cmdIdx` is "the command ID", but the server is in this odd state, where we treat command IDs and sub-command IDs  differently. In most places, it's not a problem, but here is where some of the brokenness surfaces. This is one of the planned re-factors.

This function's job is to figure out which node (which is also represented by a list of numbers) in the tree is best one to use for the given command/sub-command. So, yes, it's trying to find the best `CmdIdGroup`, `SubCmdRoot`, or `CmdIDRange`.

It doesn't have to finish at a leaf, e.g. command `[10 0 1]` refers to the second command buffer in the submit, this will return the node created for that command buffer. Or if the command given is the last one in a group, the group is chosen, rather than the command itself.

We actually never end the loop early, because of the built in assumptions that `SubCmdRoot` nodes only contain other `SubCmdRoot` or `CmdIdGroup` nodes and `CmdIdGroup` nodes only contain other `CmdIDGroup` or `CmdIDRange` nodes.

gapis/resolve/command_tree.go
pmuetschard: Decreasing `i` makes the loop go around again in the same "place", but we've updated `group` to point to the next group, so the it will be processed next. This loop is a bit confusing, because it's mixing an iterative pattern with a recursive pattern. It was the same before with two nested loops. This part didn't change and since we'll refactor it later, I didn't want to change this too much.

PR #1003, opened 2021-12-20
Better Command Tree Filtering UI

pmuetschard:
Use a more "drop-down" style UI and bring back the search.


pau-baiget (approved):
LGTM, however it would be *great* if we can add screenshots with the before/after when we do UI changes.

gapic/src/main/com/google/gapid/util/Paths.java
```diff
@@ -571,6 +565,64 @@ public static String toString(Object path) {
...
+          showSubmitInfoNodes != from.showSubmitInfoNodes ||
+          showSyncCommands != from.showSyncCommands ||
+          showBeginEndCommands != from.showBeginEndCommands;
+      if (changed) {
+        showHostCommands = from.showHostCommands;
+        showSubmitInfoNodes = from.showSubmitInfoNodes;
+        showSyncCommands = from.showSyncCommands;
+        showBeginEndCommands = from.showBeginEndCommands;
+      }
+      return changed;
```
yalcinmelihyasin, line 605: Is this changed value used somewhere else? Or would it matter if we return it true every time we get a command filter update and just assign the values?

Calculating changed is pretty ugly and it will only get uglier with more filters added. 

I can see that this has been called from `SetFilter()` and my assumption is that `setFilter()` is called when filter is actually needed to be changed for example changing a filter in UI. (Please let me know if that's not the case) So based on that assumption(if it is true), every time we update the filter returning `true` might be much cleaner.

gapic/src/main/com/google/gapid/views/CommandTree.java
```diff
@@ -87,18 +96,41 @@
...
 
   public CommandTree(Composite parent, Models models, Widgets widgets) {
     super(parent, SWT.NONE);
     this.models = models;
+    this.filter = models.commands.getFilter();
 
     setLayout(new GridLayout(1, false));
 
-    EventsFilter eventsFilter = new EventsFilter(this);
+    Composite top = createComposite(this, withMarginOnly(new GridLayout(2, false), 0, 0));
```
yalcinmelihyasin, line 113: can we rename this to what this component actually responsible for?

gapic/src/main/com/google/gapid/models/CommandStream.java
```diff
@@ -181,6 +190,9 @@ public void selectCommands(CommandIndex index, boolean force) {
     RootNode root = (RootNode)getData();
     if (index.getNode() == null) {
       resolve(index.getCommand(), node -> selectCommands(index.withNode(node), force));
+    } else if (!index.getNode().getTree().equals(root.tree)) {
+      // TODO
```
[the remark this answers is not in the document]
pmuetschard: The extra info is in the exception message. This is adding back a check that was erroneously removed by #983.

yalcinmelihyasin (commented):
I have two minor comments, otherwise it seems fine.

gapic/src/main/com/google/gapid/util/Paths.java
pmuetschard: Yes it is used, if the filter doesn't change, the UI is not updated. There is no point in reloading the tree if the filtering options didn't change (i.e. the user opens the filter pop-up, but doesn't end up changing things, or changes them and then changes them back before closing it).

pmuetschard (commented):
Screenshot:
image

PR #1004, opened 2021-12-22
Profiling Data Fixes and Refactorings

pmuetschard:
This makes the Profiling data mapping work again by correctly associating slices to commands. It also ensures that the profiling data tree matches the command tree in preparation for merging the command and profiling tabs and adding in the new draw call counters.


gapis/trace/android/profile/slices.go
```diff
@@ -224,3 +208,75 @@ func (e *extras) get(ctx context.Context, argSet int64) []*service.ProfilingData
...
+	groupTreeNode
+}
+
+// TODO: this function makes some assumptions about command/sub command IDs:
+// 1. we only get groups for command buffers, renderpasses and draw calls.
+// 2. no overlaps.
+// 3. the sub command ids are [cmdId, submission, cmdbuff, cmd].
+// All these assumptions currently hold and are also made in other parts of the
+// code in some way. The assumptions will need to be codified as part of the
+// command/sub-command refactor that is already planned.
```
pau-baiget, line 231: Do we have an internal task to track this?

gapic/src/main/com/google/gapid/views/PerformanceView.java
```diff
@@ -706,7 +704,7 @@ public void updateSelectedCell(Service.ProfilingData.GpuCounters.Entry entry, Se
       this.selectedEntry = entry;
       this.selectedMetric = metric;
       String coordinateStr = ENTRY_LABEL_PREFIX
-          + (entry != null ? entry.getGroup().getName() : "")
+          + (entry != null ? entry.getGroup() : "")
```
pau-baiget, line 707: Is this converted to String automatically?

gapis/service/service.proto
```diff
@@ -1078,8 +1078,8 @@ message ProfilingData {
     message Group {
       int32 id = 1;
       string name = 2;
-      Group parent = 3;
-      path.Command link = 4;
+      int32 parent = 3; // references Group.id
```
yalcinmelihyasin, line 1081: Instead of commenting should we name it to parentID?

gapis/service/service.proto
```diff
@@ -1132,7 +1132,7 @@ message ProfilingData {
 
     // Entry contains performance data for a specific command.
     message Entry {
-      GpuSlices.Group group = 1;
+      int32 group = 1; // references slices.groups
```
yalcinmelihyasin, line 1135: again should the name reflect what it refers instead of only commenting?

yalcinmelihyasin (changes requested):

yalcinmelihyasin:
There is at least a crash that we need to fix and some other comments

gapic/src/main/com/google/gapid/models/Profile.java
```diff
@@ -131,7 +133,7 @@ public void linkCommandToGpuGroup(List<Long> commandIndex) {
       return;
     }
     for (Service.ProfilingData.GpuSlices.Group group : getData().getSlices().getGroupsList()) {
-      if (group.getLink().getIndicesList().toString().equals(commandIndex.toString())) {
+      if (group.getLink().getToList().toString().equals(commandIndex.toString())) {
```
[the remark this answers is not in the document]
pmuetschard: `getToList`

gapis/service/service.proto
pmuetschard: Done.

gapis/service/service.proto
pmuetschard: Done.

gapis/api/sync/lookup.go
```diff
@@ -0,0 +1,249 @@
...
+		byCommandBuffer: map[uint64][]RenderPassKey{},
+		byFramebuffer:   map[uint64][]RenderPassKey{},
+	}
+}
+
+func (l *renderPassLookup) add(key RenderPassKey, idx SubCmdRange) {
+	key.RenderPass = 0 // clear it out, since we don't care about it anymore.
+
+	if current, ok := l.mappings[key]; !ok || !current.contains(idx) {
+		l.mappings[key] = current.expand(idx)
```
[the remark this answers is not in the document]
yalcinmelihyasin: My refactor suggestion is based on that if a key cannot be found in a map, it would return `nil` and if that assumption is wrong, therefore it is expected that the refactor will not do the right thing. It does not have anything about the code or the code style. 

I would not expect a value returned by a non-existing element but if that's what Go does the let's make sure that the returned element, whatever it is, suits to the logic.

And for your comment, I can't see how it prevents easily adding something at the end of the function.

yalcinmelihyasin (approved):

gapis/api/sync/lookup.go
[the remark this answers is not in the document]
pmuetschard: Even with that assumption, the re-writes are broken as they are not updating `l.mappings` in the `!ok` case, making the lookup redundant as there never will be any mappings. My point is that early return doesn't inherently make code more readable or avoid control-flow bugs any more than non-early return style.

re: adding stuff to the end of a function. If I have a block that may contain a bunch of `if`s to skip parts of it, control flow will still continue after the block. While, if I go ahead and change it to do early returns instead of skips, control flow will not always reach the end of the block. If the early return is a bit buried, someone might miss it and wonder why new code added after the block is not run in some circumstances. Don't get me wrong, this is a minor issue in most cases and I'm fine with early returns.

PR #1008, opened 2022-01-07
Don't crash the UI if the tree is empty.

pmuetschard:


gapic/src/main/com/google/gapid/widgets/LinkifiedTree.java
```diff
@@ -293,7 +293,7 @@ protected T getElement(Widget item) {
 
     @Override
     public Object[] getChildren(Object parent) {
-      return !hasChildren(parent) ? null : getChildNodes(cast(parent));
+      return !hasChildren(parent) ? new Object[0] : getChildNodes(cast(parent));
```
ttanatb, line 296: Shouldn't calls to getChildren check for null instead of returning an empty array?

pmuetschard: `getChildren` is called from the JFace library, which is where the exception originates.

ttanatb (approved):

PR #1011, opened 2022-01-14
Don't show the filter and metadata button in the ProfileView.

pmuetschard:
When showing the system profiler view in the frame profiler, hide some of the UI elements that don't make sense in this view.

Bug: http://b/211637885

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/TraceView.java
+++ b/gapic/src/main/com/google/gapid/perfetto/TraceView.java
@@ -72,7 +72,8 @@ public class TraceView extends Composite
 
   private static TraceComposite<State.ForSystemTrace> createTraceUi(
       Composite parent, Models models, Theme theme) {
-    return new TraceComposite<State.ForSystemTrace>(parent, models.analytics, models.perfetto, theme) {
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/perfetto/TraceView.java
```diff
@@ -72,7 +72,8 @@ public TraceView(Composite parent, Models models, Widgets widgets) {
 
   private static TraceComposite<State.ForSystemTrace> createTraceUi(
       Composite parent, Models models, Theme theme) {
-    return new TraceComposite<State.ForSystemTrace>(parent, models.analytics, models.perfetto, theme) {
+    return new TraceComposite<State.ForSystemTrace>(
+        parent, models.analytics, models.perfetto, theme, true) {
```
ttanatb, line 76: Generally a fan of leaving a parameter comment like `/* fullView= */ true` for parameters that aren't as obvious

ttanatb (approved):

gapic/src/main/com/google/gapid/perfetto/TraceView.java
pmuetschard: Done.

PR #1014, opened 2022-01-18
De-select source tab before destroying it

mikes-lunarg:
Leaving it selected after destruction was causing a UI exception on macOS.

Bug: b/191637172
Test: Manual

@pmuetschard I wonder if you have another way to test this fix. While I did see an exception popup while reproducing the issue, I was not able to reproduce "AGI freezes and needs to be restarted" behavior from the bug report.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/ShaderView.java
+++ b/gapic/src/main/com/google/gapid/views/ShaderView.java
@@ -73,7 +73,6 @@ import java.io.BufferedWriter;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.Writer;
-import java.util.Arrays;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/views/ShaderView.java
```diff
@@ -305,6 +307,9 @@ public void setShader(Service.Resource resource, API.Shader shader) {
           sourceTab.setText(shaderMessage.getSourceLanguage());
         }
       } else if (sourceTab != null) {
+        if (Arrays.asList(tabFolder.getSelection()).contains(sourceTab)) {
+          tabFolder.setSelection(spirvTab);
+        }
```
pmuetschard, line 312: Since we only have two tabs, I would just do an unconditional `tabFolder.setSelection(spirvTab)` here, but this is fine, too.

pmuetschard (approved):
If this fixes an exception we should do it, and it is likely that if you cannot reproduce the hang otherwise that the hang was due to this, but now is just an exception due to changes in the OS or the UI library.

gapic/src/main/com/google/gapid/views/ShaderView.java
mikes-lunarg: I like simpler, I'll go ahead and push the change.

PR #1019, opened 2022-01-24
Implement khr create renderpass2

pmuetschard:
This is a re-submit of #978 along with a fix to the replay handling mapping code, which got broken by said PR.


yalcinmelihyasin (approved):

yalcinmelihyasin:
See #1021

PR #1027, opened 2022-02-03
Change the location of the dmgbuild 

yalcinmelihyasin:
to update the build machine to Big-sur


purvisa-at-google-com:
LGTM

purvisa-at-google-com (approved):

pau-baiget (approved):

yalcinmelihyasin:
Change actually works, the timeout error is a known error which is unrelated. The log shows that the line works fine.

PR #1035, opened 2022-02-10
Add third set of Mali counters for validation

m-atanassov:
* Add driver version to GPU protobuf
 * Introduce a hardware-agnostic mapping between `id` and Mali counter
 * Use the new driverVersion field in the GPU protobuf to determine
   which set of counters to use for validation
 * Note: this is a breaking change for users who have a text-based
   perfetto config, since counter IDs have changed

### DIFF
--- a/gapis/trace/android/mali/validate.go
+++ b/gapis/trace/android/mali/validate.go
@@ -42,6 +42,17 @@ var (
 		{65536, "GPU utilization", counterChecker()},
 		{65581, "Execution core utilization", counterChecker()},
 	}
+
+	// csfCounters2 are the hardware counters found on CSF based GPUs, with
+	// hardware-independent numerical IDs.
+	csfCounters2 = []validate.GpuCounter{
+		{6, "GPU active cycles", counterChecker()},
+		{1, "Any iterator active cycles", counterChecker()},
+		{45, "Fragment jobs", counterChecker()},
+		{196, "Fragment active cycles", counterChecker()},
+		{65536, "GPU utilization", counterChecker()},
+		{65579, "Execution core utilization", counterChecker()},
+	}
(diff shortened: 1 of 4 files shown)


gapis/trace/android/mali/validate.go
```diff
@@ -85,6 +97,18 @@ func (v *MaliValidator) GetCounters() []validate.GpuCounter {
 		strings.HasSuffix(gpuName, "G78AE") {
 		return jmCounters
 	} else {
-		return csfCounters
+		/*
+		 * Driver versions reported via Vulkan have the following format:
+		 * | 31 .. 29 | 28 .. 22 | 21 .. 12 | 11 .. 0 |
+		 * | variant  |  major   |  minor   |  patch  |
+		 */
```
pmuetschard, line 104: go style prefers `//` comments, even for multi-line, for some reason.

pmuetschard (commented):
Looks good, just a tiny nit.

gapis/trace/android/mali/validate.go
m-atanassov: Thanks, I'll get it changed. It's not something `go fmt` flagged so thought it'd be fine.

pmuetschard: Thank you! gofmt doesn't lint/check style. Here's the style guide: https://go.dev/doc/effective_go#commentary, :man_shrugging:

pmuetschard (approved):

m-atanassov:
A quick  scan of the macos run suggests the failure isn't related to the patch itself:
```
fail("Please install Xcode and setup the path using xcode-select. You need Xcode, the CLI tools are not enough.")
```

PR #1047, opened 2022-03-02
Add more UI for device validation

ttanatb:
Add UI elements to display more info around validation failure

Here's what it looks like for failures before tracing (no trace file link). We don't actually require Android API level 99 lol
Screen Shot 2022-03-02 at 7 02 24 PM

Here's what it looks like when the trace validation fails (also shows what it looks like in a different dialog)
Screen Shot 2022-03-02 at 7 05 20 PM

And for good measure here's what it looks like when it passes:
Screen Shot 2022-03-02 at 7 07 43 PM

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Devices.java
+++ b/gapic/src/main/com/google/gapid/models/Devices.java
@@ -426,22 +426,30 @@ public class Devices {
   }
 
   public static class DeviceValidationResult {
-    public static final DeviceValidationResult PASSED = new DeviceValidationResult(null, true, false);
-    public static final DeviceValidationResult FAILED = new DeviceValidationResult(null, false, false);
-    public static final DeviceValidationResult SKIPPED = new DeviceValidationResult(null, true, true);
+    public static final DeviceValidationResult PASSED = new DeviceValidationResult(null, null, "", true, false);
+    public static final DeviceValidationResult FAILED = new DeviceValidationResult(null, null, "", false, false);
+    public static final DeviceValidationResult SKIPPED = new DeviceValidationResult(null, null, "", true, true);
 
-    public final Service.Error error;
+    public final Service.Error internalErr;
+    public final String validationFailureMsg;
+    public final String tracePath;
     public final boolean passed;
     public final boolean skipped;
 
-    public DeviceValidationResult(Service.Error error, boolean passed, boolean skipped) {
-      this.error = error;
+    public DeviceValidationResult(Service.Error internalErr, String validationFailureMsg, String tracePath, boolean passed, boolean skipped) {
+      this.internalErr = internalErr;
+      this.validationFailureMsg = validationFailureMsg;
+      this.tracePath = tracePath;
       this.passed = passed;
       this.skipped = skipped;
     }
 
     public DeviceValidationResult(Service.ValidateDeviceResponse r) {
-      this(r.getError(), !r.hasError() && r.getResult().getValidationFailureMsg().length() == 0, false);
+      this(r.hasError() ? r.getError() : null,
+           r.getResult() == null ? "" : r.getResult().getValidationFailureMsg(),
+           r.getResult() == null ? "" : r.getResult().getTracePath(),
+           !r.hasError() && r.getResult().getValidationFailureMsg().length() == 0,
+           false);
     }
 
     @Override
@@ -455,6 +463,13 @@ public class Devices {
       }
     }
 
+    public String errorMessage() {
+      if (internalErr != null) {
+        return "Internal Error: " + internalErr.getErrInternal().getMessage();
+      } else {
+        return "Validation Failure: " + validationFailureMsg;
+      }
+    }
   }
 
   private static class DeviceValidationCache {
--- a/gapic/src/main/com/google/gapid/views/DeviceValidationView.java
+++ b/gapic/src/main/com/google/gapid/views/DeviceValidationView.java
@@ -17,7 +17,13 @@ package com.google.gapid.views;
 
 import static com.google.gapid.util.Logging.throttleLogRpcError;
 import static com.google.gapid.widgets.Widgets.createLink;
+import static com.google.gapid.widgets.Widgets.withLayoutData;
+import static com.google.gapid.widgets.Widgets.withSpans;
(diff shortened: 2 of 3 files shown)


gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -426,22 +426,30 @@ public boolean isFuchsia() {
...
+      this.tracePath = tracePath;
       this.passed = passed;
       this.skipped = skipped;
     }
 
     public DeviceValidationResult(Service.ValidateDeviceResponse r) {
-      this(r.getError(), !r.hasError() && r.getResult().getValidationFailureMsg().length() == 0, false);
+      this(r.hasError() ? r.getError() : null,
+           r.getResult() == null ? "" : r.getResult().getValidationFailureMsg(),
+           r.getResult() == null ? "" : r.getResult().getTracePath(),
```
pmuetschard, line 450: `r.getResult()` will never return `null` and since the default string proto value is `""`, you can simply get rid of the check.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -455,6 +463,13 @@ public String toString() {
       }
     }
 
+    public String errorMessage() {
+      if (internalErr != null) {
+        return "Internal Error: " + internalErr.getErrInternal().getMessage();
```
pmuetschard, line 468: This assumes we don't ever get any other type of error. As I had pointed out in the previous PR, this should be handled by the `Client` class, which already knows how to translate `Service.Error` into messages (via exceptions).

gapic/src/main/com/google/gapid/views/DeviceValidationView.java
```diff
@@ -63,26 +78,29 @@ public DeviceValidationView(Composite parent, Models models, Widgets widgets) {
...
-    statusText = createLink(this, "", e -> {
-      Program.launch(URLs.DEVICE_COMPATIBILITY_URL);
-    });
-    statusText.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
+    statusLoader.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, false, false));
+    statusText = withLayoutData(createLabel(this, ""),
+      new GridData(SWT.FILL, SWT.CENTER, true, false));
+    retryButton = withLayoutData(createButton(this, "Retry", e -> {
+      // Intentionally empty.
+    }), new GridData(SWT.FILL, SWT.CENTER, false, false));
```
pmuetschard, line 92: `CENTER` doesn't make sense, without grabbing the extra space (e.g. if `verticalAlignment` is `CENTER`, `grabExcessVerticalSpace` should be `true`).

pmuetschard (commented):
This makes the UI showing the validation status resize. This is a problem as it is used in dialogs, which do not resize if their content is resized.

I think a better approach would be to show a "View Details" clickable, when validation fails, that will show the validation failures in a new dialog.

ttanatb (commented):
I don't think that a dialog that opens another dialog is a healthy practice and would rather resize the shell.

yalcinmelihyasin:
> I don't think that a dialog that opens another dialog is a healthy practice and would rather resize the shell.

+1

lpy:
IMHO, the reason of device validation failure is only useful or actionable to certain types of users, e.g. GPU vendors. This piece of information is useful in bug report as well, but from a user's perspective, perhaps we should think about whether the majority of users would find this useful.

Hence IMHO having a clickable to show details is better than showing a textbox with the shell resized.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -193,9 +194,37 @@ public void selectReplayDevice(Device.Instance dev) {
...
+        } catch (ExecutionException e) {
+          throttleLogRpcError(LOG, "LoadData error", e);
+          return error(null);
+        }
+      }
+
+      @Override
+      protected void onUiThreadSuccess(Service.DeviceValidationResult r) {
+        DeviceValidationResult result = new DeviceValidationResult(r);
+        submitIfNotDisposed(shell, () -> validationCache.add(device, result));
```
pmuetschard, line 217: No need to schedule this on the UI thread, since you already are on the UI thread here.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -193,9 +194,37 @@ public void selectReplayDevice(Device.Instance dev) {
...
+        onValidationResultCb.accept(result);
+      }
+
+      @Override
+      protected void onUiThreadError(RpcException error) {
+        DeviceValidationResult result = new DeviceValidationResult(error);
+        submitIfNotDisposed(shell, () -> validationCache.add(device, result));
```
pmuetschard, line 224: No need to schedule this on the UI thread, since you already are on the UI thread here.

gapic/src/main/com/google/gapid/views/DeviceValidationView.java
```diff
@@ -103,46 +136,62 @@ public void ValidateDevice(Device.Instance dev) {
...
+          OS.openFileInSystemExplorer(new File(result.tracePath));
+        } catch (IOException exception) {
+          LOG.log(SEVERE, "Failed to open log directory in system explorer", exception);
+        }
+      }), new GridData(SWT.LEFT, SWT.TOP, false, false));
+    }
+
+    // Resize the rest of the window if needed.
+    Point curr = getShell().getSize();
+    Point want = getShell().computeSize(SWT.DEFAULT, SWT.DEFAULT);
```
pmuetschard, line 189: Since you will not resize the X dimension, pass it to `computeSize`: `getShell().computeSize(curr.x, SWT.DEFAULT);`

pmuetschard (approved):

ttanatb:
> Hence IMHO having a clickable to show details is better than showing a textbox with the shell resized.

That's the idea I first started with (initially show just actionable info that's useful to most users with an ExpandBar hiding more info), but I had a lot of trouble getting an expand widget to work correctly with SWT (a "learn more" type button that shows more info about the error). I tried using a similar approach to the error dialog but I had a lot of trouble getting the expand bar to work correctly while also making sure the formatting inside the expanded section is correct.

PR #1048, opened 2022-03-03
Improves the GPU counter selection.

pmuetschard:
- Use the select-by-default field of the counter info to determine whether a counter should be enabled by default.
- Remember selected counters by GPU name, as counter IDs vary by GPU model and vendor, making the remembered counter IDs messy for users tracing on multiple devices.

Bug: http://b/172579026

### DIFF
--- a/gapic/src/main/com/google/gapid/settings.proto
+++ b/gapic/src/main/com/google/gapid/settings.proto
@@ -142,13 +142,18 @@ message Perfetto {
   }
   CPU cpu = 1;
 
+  message GPUCounters {
+    repeated int32 counter_ids = 1;
+  }
+
   message GPU {
     bool enabled = 1;
     bool slices = 2;
     bool counters = 3;
     int32 counter_rate = 4;
-    repeated int32 counter_ids = 5;
+    reserved 5;
     bool surface_flinger = 6;
+    map<string, GPUCounters> counters_by_gpu = 7;
   }
   GPU gpu = 2;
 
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/settings.proto
```diff
@@ -142,13 +142,18 @@ message Perfetto {
...
+    repeated int32 counter_ids = 1;
+  }
+
   message GPU {
     bool enabled = 1;
     bool slices = 2;
     bool counters = 3;
     int32 counter_rate = 4;
-    repeated int32 counter_ids = 5;
+    reserved 5;
```
yalcinmelihyasin, line 154: We may not need this as we distribute the server and client at the same time. You can instead just switch it. You can ask Pascal for details.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -511,10 +511,12 @@ protected String createAndShowDialog(String current) {
         systemTracingConfigLabel = createLabel(systemTracingConfig, PERFETTO_LABEL);
         Widgets.createButton(systemTracingConfig, "Configure", e -> {
           DeviceCaptureInfo dev = getSelectedDevice();
-          if (dev != null && dev.isFuchsia()) {
+          if (dev == null) {
```
yalcinmelihyasin, line 514: This might be a better approach.

```
if (dev != null) {
    if (dev.isFuchsia()) {
            showFuchsiaConfigDialog(getShell(), models, widgets);
     } else {
            showPerfettoConfigDialog(getShell(), models, widgets, dev);
     }
}
```

gapic/src/main/com/google/gapid/settings.proto
pmuetschard: Actually, the settings are serialized and stored, so it's the only place where it would matter. Although, we currently serialize it to text-format, so we'd be OK re-using tag IDs, it's better practice not to.

gapic/src/main/com/google/gapid/views/TracerDialog.java
pmuetschard: I don't think that's better. It's extra indentation, it doesn't explain why `dev == null` is not being handled, and it doesn't match the same pattern in `updateSystemTracingConfigLabel`.

ttanatb (approved):

PR #1049, opened 2022-03-03
Better defaulting to ANGLE/Vulkan for frame captures. 

pmuetschard:
In the old logic, if a System trace is taken, any subsequent frame trace would default to Vulkan. Now, we default to ANGLE, until a Vulkan trace is taken and switch back to defaulting to ANGLE if an ANGLE trace is taken.

Bug: http://b/208505553

### DIFF
--- a/gapic/src/main/com/google/gapid/settings.proto
+++ b/gapic/src/main/com/google/gapid/settings.proto
@@ -114,6 +114,7 @@ message Trace {
   string device_serial = 1;
   string device_name = 2;
   string type = 3;
+  bool last_frame_trace_was_vulkan = 21;
   string api = 4 [deprecated = true];  // Kept for de-serialization.
   string uri = 5;
   string arguments = 6;
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -365,7 +365,8 @@ public class TracerDialog {
 
         setLayout(new GridLayout(1, false));
 
-        Service.TraceType lastType = Service.TraceType.Graphics;
+        Service.TraceType lastType = trace.getLastFrameTraceWasVulkan() ?
+            Service.TraceType.Graphics : Service.TraceType.ANGLE;
         try {
           lastType = Service.TraceType.valueOf(trace.getType());
         } catch (IllegalArgumentException e) {
@@ -375,7 +376,7 @@ public class TracerDialog {
         if (type == null) {
           // Use the last used type if no specific type was requested.
           type = TraceType.from(lastType);
-        } else if (type == TraceType.Vulkan && lastType == Service.TraceType.ANGLE) {
+        } else if (type == TraceType.Vulkan && !trace.getLastFrameTraceWasVulkan()) {
           // If frame profiler was requested, use ANGLE if the previous trace was an ANGLE trace.
           type = TraceType.ANGLE;
         }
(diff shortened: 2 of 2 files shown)


gapic/src/main/com/google/gapid/settings.proto
```diff
@@ -114,6 +114,7 @@ message Trace {
   string device_serial = 1;
   string device_name = 2;
   string type = 3;
+  bool default_no_angle = 21;
```
yalcinmelihyasin, line 117: Can we add this to the end instead of middle.

yalcinmelihyasin (changes requested):

gapic/src/main/com/google/gapid/settings.proto
pmuetschard: It makes more logical sense to be here as it goes along with the type. Why add it to the end?

ttanatb, line 117: Does it make more sense to have this as something like:

```
TraceType trace_type = 21; // or default_trace_type
```
using the `TraceType` enum defined in `gapis/service/service.proto` ?

I think it takes less mental reasoning/gymnastics to understand compared to default_no_angle

ttanatb (approved):

gapic/src/main/com/google/gapid/settings.proto
pmuetschard: `type` is already that - it's the last treace type used and becomes the default for future traces. However, we need some extra information to keep track of whether the last *frame* profiling trace (which may not have been the previous trace, if a System trace is taken) was ANGLE or not. The name is a bit odd, I agree, because for the settings the default proto value should be the default for new fields, so that people upgrading don't get odd behavior. Since `false` is the default value, we have to invert the field from `default_angle`.

`type` is a string rather than `Service.TraceType`, because this is serialized and needs to remain backwards compatible and this field is actually older than `Service.TraceType`.

ttanatb: Instead of `default_no_angle` maybe something like `default_to_vulkan` instead? Just to get rid of the double negatives.

When calling `setDefaultToVulkan(false)` it makes some amount of sense, but the getter isn't the most clear `getDefaultToVulkan()` (not clear that we're getting a bool here)

Would prefer something like `default_frame_profiling_type` which is an enum to either vulkan or ANGLE, so that the setters & getters both make sense. If we use the TraceType enum in service.proto this would be the vulkan trace and should be okay? Thoughts?

pmuetschard: Changed it to `last_frame_trace_was_vulkan`.

Originally, I had chosen `not_angle` instead of `vulkan`, because `not angle != vulkan` if we get a third alternative, which arguably may not happen.

yalcinmelihyasin: Because there is a number `21` next to it. When someone needs to add another type in that proto, they shouldn't be looking at all the proto to find highest number. if that really needs to be next to it, then we should order the numbers again.

Protobuffers fundamentally are a backward compatible serialisation API. That is the first and foremost purpose of  protobuffers(or any such APIs). So the highest priority in using them and keeping things in order should be based on that. 

Luckily we don't have backward compatibility issues so that we can avoid most of the compatibility handling but we can at least keep the numbers in order. Any other ordering, alphabetical, logical etc. comes after that.

pmuetschard: Again, this proto is serialized and stored on disk, so we actually do have to keep backwards compatibility.

Having the tag IDs get out-of-order is, well, just a part of protobufs. Take a look at https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto.

In my opinion, before you add/delete/modify a protobuf message, you better read the entire message, so you understand how your change fits in with what is already there. Some people like to add a comment at the beginning of the message with the next tag ID to use. That is a strategy we could adopt, if we really do get into trouble figuring out what the next ID should be. I don't know, IDEs help with that, so I'm not worried about this.

PR #1052, opened 2022-03-09
Fix counter stats for duration highlights.

pmuetschard:
Bug: http://b/187988971

Supersedes #1020


yalcinmelihyasin (approved):

pau-baiget (approved):
Looks good, we need to test it thoroughly if we're merging this for GA.

PR #1065, opened 2022-03-18
Fix the data attributes of gapit & gapis.

pmuetschard:


mikes-lunarg (approved):
Yep, this fixes `bazel run cmd/agi` on my development machine

PR #1068, opened 2022-03-23
Fix VK_KHR_separate_depth_stencil_layouts

mikes-lunarg:
Even though the test app captured and replayed without issue, the state view
showed that `VkAttachmentDescriptionStencilLayout` was not saved for the
render passes that included it. This appears to be caused by the api
file silently failing to modify `renderPass.AttachmentDescriptions[i]`
after initially setting it, so set it after processing the pNext chain
instead.

### DIFF
--- a/gapis/api/vulkan/api/renderpass_framebuffer.api
+++ b/gapis/api/vulkan/api/renderpass_framebuffer.api
@@ -366,7 +366,7 @@ sub ref!RenderPassObject createRenderPassObjectFromInfo2(
   attachments := info.pAttachments[0:info.attachmentCount]
   for i in (0 .. info.attachmentCount) {
     attachment := attachments[i]
-    renderPass.AttachmentDescriptions[i] = AttachmentDescription(
+    attachmentDescription := AttachmentDescription(
       Flags: attachment.flags,
       Format: attachment.format,
       Samples: attachment.samples,
@@ -386,7 +386,7 @@ sub ref!RenderPassObject createRenderPassObjectFromInfo2(
         switch sType {
           case VK_STRUCTURE_TYPE_ATTACHMENT_DESCRIPTION_STENCIL_LAYOUT: {
             ext := as!VkAttachmentDescriptionStencilLayout*(next.Ptr)[0]
-            renderPass.AttachmentDescriptions[j].StencilLayout = new!AttachmentDescriptionStencilLayout(
+            attachmentDescription.StencilLayout = new!AttachmentDescriptionStencilLayout(
               StencilInitialLayout: ext.stencilInitialLayout,
               StencilFinalLayout: ext.stencilFinalLayout,
             )
@@ -394,6 +394,7 @@ sub ref!RenderPassObject createRenderPassObjectFromInfo2(
         }
       }
     }
+    renderPass.AttachmentDescriptions[i] = attachmentDescription
   }
 
   subpasses := info.pSubpasses[0:info.subpassCount]


mikes-lunarg:
@pmuetschard I'm still unsure why the original code was failing to update the render pass object, any thoughts?

mikes-lunarg:
@mark-lunarg That's exactly how I found it. As an extra sanity check I went looking for the new extension data in the state tab of AGI and it wasn't there.

AWoloszyn:
Map access operations return a value not a reference.
See https://github.com/google/agi/tree/master/gapil#map (although that could probably be a bit clearer)
So given that:

`renderPass.AttachmentDescriptions[j]` returns a value
`renderPass.AttachmentDescriptions[j].StencilLayout` gives you a temporary, then sets the value on that temporary, then discards it.

mikes-lunarg:
> Map access operations return a value not a reference. See https://github.com/google/agi/tree/master/gapil#map (although that could probably be a bit clearer) So given that:
> 
> `renderPass.AttachmentDescriptions[j]` returns a value `renderPass.AttachmentDescriptions[j].StencilLayout` gives you a temporary, then sets the value on that temporary, then discards it.

Makes sense, thanks for the info!

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -376,7 +376,6 @@ sub ref!RenderPassObject createRenderPassObjectFromInfo2(
       StencilStoreOp: attachment.stencilStoreOp,
       InitialLayout: attachment.initialLayout,
       FinalLayout: attachment.finalLayout,
-      StencilLayout: null,
```
yalcinmelihyasin, line 379: I would keep the null if possible to explicitly show that it's not set.

While looking at the constructor, it's not obvious to see there is an unset field.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -366,7 +366,7 @@ sub ref!RenderPassObject createRenderPassObjectFromInfo2(
   attachments := info.pAttachments[0:info.attachmentCount]
   for i in (0 .. info.attachmentCount) {
     attachment := attachments[i]
-    renderPass.AttachmentDescriptions[i] = AttachmentDescription(
```
yalcinmelihyasin, line 369: I am having hard time to see how changing this solves the problem

yalcinmelihyasin:
Based on the change are these two examples are semanticly different?

```
renderPass.AttachmentDescriptions[j] = AttachmentDescripton(...)
```

```
attachmentDescription = AttachmentDescripton(...)
renderPass.AttachmentDescriptions[j] = attachmentDescripton
```

AWoloszyn:
> > Map access operations return a value not a reference. See https://github.com/google/agi/tree/master/gapil#map (although that could probably be a bit clearer) So given that:
> > `renderPass.AttachmentDescriptions[j]` returns a value `renderPass.AttachmentDescriptions[j].StencilLayout` gives you a temporary, then sets the value on that temporary, then discards it.
> 
> Are the a difference in handling lvalue and rvalue?

AWoloszyn:
> Based on the change are these two examples are semantically different? @AWoloszyn
> 
> ```
> renderPass.AttachmentDescriptions[j] = AttachmentDescripton(...)
> ```
> 
> ```
> attachmentDescription := AttachmentDescripton(...)
> renderPass.AttachmentDescriptions[j] = attachmentDescripton
> ```

Those are the same.  What would be different is 
```
a := A(...)
a.foo = Foo()
blah.a[j] = a
```

```
a := A(...)
blah.a[j] = a
blah.a[j].foo = Foo()
```

The difference is that
```
blah.a[j].foo = Foo()
```
is a no-op.

AWoloszyn:
> > Map access operations return a value not a reference. See https://github.com/google/agi/tree/master/gapil#map (although that could probably be a bit clearer) So given that:
> > `renderPass.AttachmentDescriptions[j]` returns a value `renderPass.AttachmentDescriptions[j].StencilLayout` gives you a temporary, then sets the value on that temporary, then discards it.
> 
> Are the a difference in handling lvalue and rvalue?

This is more about with what value `map[]` returns. In psuedo-C++ it would look something like:

```
template<typename K, typename V>
class map {
    V operator[](K key);
};
```

This differs from a STD unordered_map which looks like
```
template<typename K, typename V>
class map {
    V& operator[](K key);
};
```

I.e. the [] operator in GAPIL returns a COPY of the element in the map, not a REFERENCE to the element in the map.

gapis/api/vulkan/api/renderpass_framebuffer.api
mikes-lunarg: I think @AWoloszyn gave a pretty thorough answer, but the short version is:

I need to might need to modify this struct when processing the pNext loop so I assign it to a local variable. Then, once I have the final version of the struct, I can put it in the map. Modifying it in-place in the map (the original code) won't work because the map access or will give me a copy.

gapis/api/vulkan/api/renderpass_framebuffer.api
mikes-lunarg: Omitting null members seems to be a pretty common pattern in the api files, but I'm happy to put it back in

mikes-lunarg:
@yalcinmelihyasin What is our next step? Can we merge this fix or would you like to see additional changes?

yalcinmelihyasin (approved):

PR #1076, opened 2022-04-07
Fixes b/228443485 by preventing resource fetch when resource is null

yalcinmelihyasin:


yalcinmelihyasin:
This fix includes two things:

- Preventing asking for get resource when resource is null.  
- Not setting shader when shader is null.

@pmuetschard please review and let me know if one is necessary or there are unknown side effects.

This bug seems to be introduced by combination of #925 and #1041. I don't know about both PRs so this fix may introduce unwanted behaviour. Please check for it.

When I open VS or FS in pipeline view, the only thing it says is "Select Shader" and when I select a shader from shader view tab nothing happens. I don't know if that's intended.

ttanatb (approved):

pmuetschard:
Root cause fixed in #1083.

PR #1079, opened 2022-04-11
[fuchsia] Add namespace augmentation to gapii-pkg.

rosasco-wk:


gapii/fuchsia/gapii_server.cc
```diff
@@ -0,0 +1,55 @@
...
+
+#include <lib/async-loop/cpp/loop.h>
+#include <lib/async-loop/default.h>
+#include <lib/fdio/directory.h>
+#include <lib/fdio/io.h>
+#include <lib/vfs/cpp/pseudo_dir.h>
+#include <lib/vfs/cpp/remote_dir.h>
+#include <zircon/processargs.h>
+
+// Serve /pkg as /pkg in the outgoing directory.
```
pmuetschard, line 23: Is this "pkg" arbitrary? if so maybe name it something "AGI-y". Note that in our distribution, there is no "pkg" directory, instead `bazel-bin/pkg` is the root of our distributable and its name is user-controlled.

pmuetschard (approved):

gapii/fuchsia/gapii_server.cc
rosasco-wk: pkg/ is a fuchsia filesystem convention and not related to AGI host packaging.

PR #1088, opened 2022-04-20
Support copy constructors in the API language.

pmuetschard:
With this change, the following syntax is supported:

```
class Foo {
  u32 foo
  u32 bar
}

f := Foo(42, 21)
g := Foo(f)        // new
h := new!Foo(f)    // new
```

This is syntactic sugar that translates into:

```
g := Foo(f.foo, f.bar)
h := new!Foo(f.foo, f.bar)
```

### DIFF
--- a/gapil/analysis/value.go
+++ b/gapil/analysis/value.go
@@ -464,9 +464,20 @@ func (s *scope) valueOf(n semantic.Expression) (out Value, setter func(Value)) {
 
 	case *semantic.ClassInitializer:
 		c := s.defaultOf(n.Class).(*ClassValue)
-		for _, f := range n.Fields {
-			v, _ := s.valueOf(f.Value)
-			c.Fields[f.Field.Name()] = v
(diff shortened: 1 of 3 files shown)


yalcinmelihyasin (approved):
Presubmit is failing, otherwise LGTM.

You can merge after fixing the presubmit

yalcinmelihyasin (changes requested):
The code looks fine but I actually tried it on a local branch and it does not entirely work. 

I have create a branch in my repo so that you can see the failing cases.

From ref->value still does not compile(and if it's too hard, I think one way conversion is enough for now)

and value->ref compiles in APIC but it causes another compilation error with templates.

Link to the PR:  https://github.com/yalcinmelihyasin/agi/tree/copy_experiment

pmuetschard:
Yes, ref->value is not supported, but that should also not be required.
I've fixed the analyzer error, PTAL.

PR #1094, opened 2022-04-25
BUILD rules for our python code

pmuetschard:
- Creates a bazel target for vk.xml
- Adds python setup to our workspace
- Changes imports to be fully-qualified off of our workspace root
- Adds build files with library targets and a main entry point target
- Adds build files for the tests, and linter rule for each python package


vulkan_generator/main.py
```diff
@@ -17,7 +17,7 @@
 from pathlib import Path
 import sys
 
-import vulkan_generator
+from gapid.vulkan_generator import vulkan_generator
```
purvisa-at-google-com, line 20: I was telling Melih that vulkan_generator shouldn't be in the base namespace and we should have some enclosing namespace for all the new project Sunrise code but I'm not sure we should use the name gapid. Since the tool is now called AGI maybe we could call it agi or agir (to copy the gapir idiom but with the new product name?)

vulkan_generator/BUILD.bazel
```diff
@@ -0,0 +1,42 @@
...
+    deps = [
+        "//vulkan_generator/vulkan_parser",
+    ],
+)
+
+py_binary(
+    name = "main",
+    srcs = ["main.py"],
+    args = [
+        "$(location @vulkan-headers//:vk.xml)",
```
purvisa-at-google-com, line 34: Have we added vk.xml to thirdparty? Is this OK from a license perspective?

yalcinmelihyasin: One more thing is the version of this repo. As far as I can see Vulkan 1.2 repo being used which will not work with the current parser, as the XML has changed. We need to update it to version is compatible with our parser.

I believe we have not intention to support more than one version of Vulkan XML. Therefore, if we are going to use a unified Vulkan repo for all our Vulkan related work. It has to be the one that parser works with.

yalcinmelihyasin (changes requested):
Thanks for adding the Bazel part.

But I believe thar we need to figure out the compatibility between bazel and python tools.

This change works with bazel but it breaks everything else.

IDE's cannot find the packages because path is wrong(for them). you cannot run the main or the tests from IDE without bazel because again path is wrong for the plain python and pytest.

It also changes the testing framework from pytest to unittest and I don't know if we have reason for that.

We have to figure out a non-hacky way to make is compatible tools with bazel.

vulkan_generator/main.py
pmuetschard: moved it to generator.py, so that it can be properly imported.

vulkan_generator/BUILD.bazel
pmuetschard: We can update our vulkan_headers dep and get the latests.

vulkan_generator/main.py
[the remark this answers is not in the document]
pmuetschard: fixed.

vulkan_generator/tests/test_types.py
```diff
@@ -20,35 +20,40 @@
...
 
+import unittest
 import xml.etree.ElementTree as ET
 
-from vulkan_parser import handle_parser
-from vulkan_parser.types import VulkanHandle, VulkanHandleAlias
+from gapid.vulkan_generator.vulkan_parser import handle_parser
+from gapid.vulkan_generator.vulkan_parser.types import VulkanHandle, VulkanHandleAlias
 
+class TypesTest(unittest.TestCase):
```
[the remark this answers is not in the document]
pmuetschard: Done.

pmuetschard:
PTAL

yalcinmelihyasin (approved):
The process is different than I would imagine but I think this is the "bazel way" to do it.

PR #1098, opened 2022-04-28
Vulkan generator 3

yalcinmelihyasin:
Implements Vulkan Function pointer parsing from Vulkan XML

### DIFF
--- a/vulkan_generator/tests/test_funcptr_parsing.py
+++ b/vulkan_generator/tests/test_funcptr_parsing.py
@@ -27,10 +27,41 @@ from vulkan_parser import types
 
 
 def test_vulkan_func_pointer() -> None:
-    """""Test the parsing of a function pointer with a pointer argument and pointer return type"""
-
+    """""Test the parsing of a function pointer"""
     xml = """<?xml version="1.0" encoding="UTF-8"?>
-    <type category="funcpointer">typedef void* (VKAPI_PTR *<name>PFN_vkReallocationFunction</name>)(
+    <type category="funcpointer">typedef void (VKAPI_PTR *
+    <name>PFN_vkInternalAllocationNotification</name>)(
+    <type>void</type>*                                       pUserData,
+    <type>size_t</type>                                      size,
+    <type>VkInternalAllocationType</type>                    allocationType,
+    <type>VkSystemAllocationScope</type>                     allocationScope);</type>
+    """
+    funcptr = funcptr_parser.parse(ET.fromstring(xml))
+
+    assert isinstance(funcptr, types.VulkanFunctionPtr)
+
+    assert funcptr.typename == "PFN_vkInternalAllocationNotification"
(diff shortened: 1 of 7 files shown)


vulkan_generator/tests/test_funcptr_parsing.py
```diff
@@ -0,0 +1,62 @@
...
+
+    funcptr = funcptr_parser.parse(ET.fromstring(xml))
+
+    assert isinstance(funcptr, types.VulkanFunctionPtr)
+    assert funcptr.typename == "PFN_vkReallocationFunction"
+    assert funcptr.return_type == "void*"
+
+    assert len(funcptr.arguments) == 5
+
+    assert funcptr.arguments[0].typename == "void*"
```
purvisa-at-google-com, line 49: Just wondering. Do we have anything to standardise these strings so we don't end up with "void*" and "void *" not being equal?

purvisa-at-google-com (approved):

vulkan_generator/tests/test_funcptr_parsing.py
yalcinmelihyasin: So simply, when parsing, I remove every whitespace between pointers. 

If the type system of C++ wouldn't be this complicated we probably would store the pointer and const values as fields of the type but as we discussed it's really hard to represent "const char* const*" without reimplementing C++ type parser.

vulkan_generator/vulkan_parser/funcptr_parser.py
```diff
@@ -0,0 +1,79 @@
...
+            parsing_utils.get_tail_from_tag(elem, "type"))
+
+        # Pointers of the type is actually in the argument name
+        if "*" in argument_name:
+            typename = typename + "*"
+            argument_name = argument_name[1:]
+
+            # Currently no function pointer has double pointers
+            # We should add support if they ever have in the future
+            if "*" in argument_name:
```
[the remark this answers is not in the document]
yalcinmelihyasin: We actually don't.

No function pointer has it but we can add something that, if there is a const in variable name(because for some reason `*` is in variable name) we can raise an exception. So if they ever add such thing to the XML, we would be aware of it and parse accordingly.

vulkan_generator/tests/test_funcptr_parsing.py
purvisa-at-google-com: There are probably some simple rules we can use here to clean up most cases without having a full type parser. Replacing any run of whitespace with a single space character and inserting a space before a * if there isn't already one there will go a long way.

yalcinmelihyasin: For now my rule is `const char* const*` a whitespace between `const` and `type` and `*` directly after `type` or `const`.

If you don't have a strong reason to add a space, let's keep it this way.

PR #1100, opened 2022-05-03
This is a small maintaince PR that changes some of the decision

yalcinmelihyasin:
I made earlier in implementation
  - Moves AllVulkanTypes dictionary under the types
  - VulkanStructMembers and VulkanFunctionArguments are no more a VulkanType
    as they don't define a type
  - Change how we store the arguments and members
  - Remove some printfs
  - Linting issues

### DIFF
--- a/vulkan_generator/vulkan_utils/parsing_utils.py
+++ b/vulkan_generator/vulkan_utils/parsing_utils.py
@@ -19,9 +19,17 @@ from typing import Optional
 import os
 
 
-def get_text_from_tag(elem: ET.Element, tag: str) -> str:
-    """Gets the text of the element with the given tag"""
-    value = try_get_text_from_tag(elem, tag)
+def get_text_from_tag_in_children(elem: ET.Element, tag: str, recursive: bool = False) -> str:
+    """Gets the text of the element with the given tag in element
+
+    Args:
+        recursive(bool):
+            If True:
+                The function will search the entire subtree recursively including the root node
+            If False:
+                The function will search only the first level children.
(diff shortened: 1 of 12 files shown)


vulkan_generator/vulkan_utils/parsing_utils.py
```diff
@@ -51,10 +78,20 @@ def get_tail_from_tag(elem: ET.Element, tag: str) -> str:
...
 
 
-def try_get_tail_from_tag(elem: ET.Element, tag: str) -> Optional[str]:
+def try_get_tail_from_tag_in_children(elem: ET.Element, tag: str, recursive: bool = False) -> Optional[str]:
     """Tries to Gets the text of the element with the given tag
-    and returns None if the tag does not exits"""
-    for child in elem.iter():
+    and returns None if the tag does not exits
+
+    Args:
```
purvisa-at-google-com, line 85: Do we need to put this same large comment in so many places? In this case I think the name explains what the variable does without the need for any special documentation

purvisa-at-google-com (approved):
LGTM not much to say about this one

vulkan_generator/vulkan_utils/parsing_utils.py
yalcinmelihyasin: This one I put specifically because behaviour changes unexpectedly(due to how XML parser works)

yalcinmelihyasin: Let's not bloat the comments that much

PR #1105, opened 2022-05-12
spvc memory leak and error code handling fixes

kocdemir:
This fixes a memory leak with the shader compilation requests made on the server, which can lead to high memory consumption due to redundant requests when the compilations are failed.
Also changed the return value check on spvc_context_create_compiler to cover other error codes.

### DIFF
--- a/gapis/shadertools/shadertools.go
+++ b/gapis/shadertools/shadertools.go
@@ -254,37 +254,48 @@ func ExtractDebugSource(shader []uint32) (string, string, bool, error) {
 	var options C.spvc_compiler_options = nil
 	var result *C.char = nil
 
+	var retSource string = ""
+	var retError error = nil
+
 	C.spvc_context_create(&context)
 	C.spvc_context_parse_spirv(context, (*C.uint)(shaderPtr), (C.size_t)(len(shader)), &ir)
 	switch sourceLanguage {
 	case "GLSL":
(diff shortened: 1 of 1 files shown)


gapis/shadertools/shadertools.go
```diff
@@ -254,37 +254,48 @@ func ExtractDebugSource(shader []uint32) (string, string, bool, error) {
...
-		C.spvc_compiler_options_set_uint(options, C.SPVC_COMPILER_OPTION_HLSL_SHADER_MODEL, module.source_language_version)
-		C.spvc_compiler_install_compiler_options(compiler, options)
-
-		C.spvc_compiler_compile(compiler, &result)
-		return C.GoString(result), sourceLanguage, true, nil
+		break
 
 	default:
-		return "", sourceLanguage, false, errors.New("Source language not supported")
+		retError = errors.New("Source language \"" + sourceLanguage + "\" is not supported: ")
```
pmuetschard, line 291: is there supposed to be something after the `:`?

pmuetschard (approved):

gapis/shadertools/shadertools.go
kocdemir: Good catch Pascal! I was initially going to add sourceLanguage to the end. Will fix that and merge. 
Thanks

yalcinmelihyasin (approved):

PR #1111, opened 2022-05-23
Remove the LLVM-based API file compilation. 

pmuetschard:
This removes the code generator that uses LLVM to generate the API files mutate code. It keeps the current LLVM code generation, but only for the encoder functions, which is the only thing currently in use.


pmuetschard:
>  I guess the concern here is that we've missed deleting something

There is some more stuff to remove, but it would require larger changes to what remains, so I'm planning on removing that when removing the LLVM dep all-toghether.

gapil/compiler/plugins/encoder/test/test.cpp
```diff
@@ -72,11 +70,11 @@ inner_class* create_inner_class_ref(arena* a, ref** p) {
...
-  return gapil_encoder_module.create_context(arena);
+  context* ctx = (context_t*)gapil_alloc(arena, sizeof(context_t), 8);
+  ctx->arena = arena;
+  return ctx;
 }
 
-void destroy_context(context* ctx) {
-  return gapil_encoder_module.destroy_context(ctx);
-}
+void destroy_context(context* ctx) { gapil_free(ctx->arena, ctx); }
```
kocdemir, line 78: Might be good to add null checks?

kocdemir (approved):
Looks like a great cleanup, thanks Pascal!

gapil/compiler/plugins/encoder/test/test.cpp
pmuetschard: No, if `ctx` is null it would be a coding error and the test should fail and `gapil_free` already handles `null`.

PR #1116, opened 2022-05-24
Fix validation issue

ttanatb:
Between `err` and `result`, only one of them will ever be nil (when `err` is `nil`, then `result` is not, and vice versa). Fix the error in the if check

Also fix case where validation error code is being replaced by an OK result instead of trace validation failure.


pau-baiget (approved):
LGTM!

PR #1117, opened 2022-05-24
Implement parsing Vulkan defines and constants

yalcinmelihyasin:


yalcinmelihyasin:
Some minor comment fixing is in this PR by mistake.

vulkan_generator/vulkan_parser/defines_parser.py
```diff
@@ -0,0 +1,110 @@
...
+    return types.VulkanDefine(
+        variable_name=name,
+        value=value
+    )
+
+
+def parse_define_by_tag(define_elem: ET.Element) -> types.VulkanDefine:
+    """Parses a define that has a name tag
+
+    If a define has a name tag, it's "usually" a macro
```
purvisa-at-google-com, line 42: Can we give better definition of "usually" here?

purvisa-at-google-com (approved):

vulkan_generator/vulkan_parser/defines_parser.py
yalcinmelihyasin: I wish I could... but I can't it's the XML's irregularity. It can be a macro or sometime a constant value. There is no rule. But in our actual code we should not need this. This is only required for Vulkan.h.

PR #1118, opened 2022-05-25
Fix validation link in gh pages

ttanatb:
https://gpuinspector.dev/validation somehow has a link that links to itself instead of the correct page. This hopefully will fix it.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -19,6 +19,7 @@ url: "https://gpuinspector.dev" # the base hostname & protocol for your site, e.
 project-url: "https://github.com/google/agi"
 releases-url: "https://github.com/google/agi/releases"
 docurl: "https://developer.android.com/agi"
+validation-url: "https://developer.android.com/agi/supported-devices"
 feedback-url: "https://github.com/google/agi/issues"
 markdown: kramdown
 sass:
--- a/_config.yml
+++ b/_config.yml
@@ -19,7 +19,7 @@ url: "https://gpuinspector.dev" # the base hostname & protocol for your site, e.
 project-url: "https://github.com/google/agi"
 releases-url: "https://github.com/google/agi/releases"
 docurl: "https://developer.android.com/agi"
-validation-url: "https://developer.android.com/agi/supported-devices"
+validation-url: "https://developer.android.com/agi/start#device-validation"
 feedback-url: "https://github.com/google/agi/issues"
 markdown: kramdown
 sass:
--- a/validation.md
+++ b/validation.md
@@ -5,4 +5,4 @@ title: Device Support Validation
 permalink: /validation
 ---
 
-The device support validation documentation has moved to [{{site.docurl}}]({{site.validation-url}}).
+The device support validation documentation has moved to [{{site.validation-url}}]({{site.validation-url}}).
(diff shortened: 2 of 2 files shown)


_config.yml
```diff
@@ -19,6 +19,7 @@ url: "https://gpuinspector.dev" # the base hostname & protocol for your site, e.
 project-url: "https://github.com/google/agi"
 releases-url: "https://github.com/google/agi/releases"
 docurl: "https://developer.android.com/agi"
+validation-url: "https://developer.android.com/agi/supported-devices"
```
kocdemir, line 22: I think "https://developer.android.com/agi/start#device-validation" would be a better link to forward people for validation related information?

validation.md
```diff
@@ -5,4 +5,4 @@ title: Device Support Validation
 permalink: /validation
 ---
 
-The device support validation documentation has moved to [{{site.docurl}}]({{https://developer.android.com/agi/start#device-validation}}).
```
kocdemir, line 8: The removed link here looks correct for the validation page, but it's not used in the webpage output(https://gpuinspector.dev/validation). So, not sure if we should actually change the 'site.docurl' part here.

kocdemir (changes requested):

_config.yml
pau-baiget: I wonder if we should move the whole Device Validation subsection into the Supported Devices section?

kocdemir:
Also, doesn't have to be within this CL but I think it would be good to change the link to "https://gpuinspector.dev/validation" in URLs.java file as well.

validation.md
ttanatb: site.docurl resolves to `https://developer.android.com/agi` which would be the text the user actually sees. The hyperlink is wrapped in `{{ }}` and my guess is that it's trying to resolve that but couldn't find a variable named  `https://developer.android.com/agi/start#device-validation` making it link to itself.

_config.yml
ttanatb: I think validation warrants its own page to make it easy to skim and find (including validation process description, table, and troubleshooting info in one page) -- I'll leave it at https://developer.android.com/agi/start#device-validation for now (since there are links to and from each page) until we adjust the docs

kocdemir: I'd prefer having Supported Devices in a separate page but 'Device Validation' subsection can be moved out of 'Quickstart' into its own page. That way we can add more explanation related to the validation error messages and any known issues.

validation.md
kocdemir: So, I think we should change the text to show 'site.validation-url' as well. Currently it will still show docurl but point to somewhere else, right? 
Interestingly all other pages on https://gpuinspector.dev/docs/ just says 'The documentation has moved' without giving direct links (e.g. https://gpuinspector.dev/docs/devices is not forwarding to https://developer.android.com/agi/supported-devices).

ttanatb:
https://github.com/google/agi/pull/1118#issuecomment-1142261330

Can't do that with this PR because this is for the gh-pages branch -- planning on doing that with upcoming UI changes

validation.md
ttanatb: done

kocdemir (approved):
lgtm, thanks

PR #1119, opened 2022-05-25
Implement command parsing from the XML

yalcinmelihyasin:


vulkan_generator/vulkan_parser/types.py
```diff
@@ -169,11 +169,59 @@ class VulkanDefine:
...
+
+    # Is this field has to be set and/or not-null
+    optional: Optional[bool]
+
+    # Is this parameter must be externally synced
+    externally_synced: Optional[bool]
+
+    # If the parameter is an array, it's size is defined by another
+    # parameter of the command. This is the name of the referring parameter.
+    array_reference: Optional[str]
```
purvisa-at-google-com, line 186: Maybe array size reference or array length reference?

purvisa-at-google-com (approved):

PR #1122, opened 2022-05-31
Don't use LLVM to generate the API encoders.

pmuetschard:
Replaces the system that uses LLVM to generate the API to proto encoders with a system that generates C++ code.


gapii/cc/runtime.cpp
```diff
@@ -54,13 +54,13 @@ void* gapil_encode_object(context* ctx, uint8_t is_group, uint32_t type,
   return nullptr;
 }
 
-void gapil_slice_encoded(context* ctx, slice_t* slice) {
+void gapil_slice_encoded(context* ctx, const void* slice) {
   DEBUG_PRINT("gapil_on_encode_slice(%p, %p)", ctx, slice);
```
kocdemir, line 58: Just noticed that the DEBUG_PRINT here prints a different function name, might be good to fix with this CL or another one later.

kocdemir (approved):
lgtm.
as a reference, my pkg compilation on debug time went down from ~42mins to ~37mins.

gapii/cc/runtime.cpp
pmuetschard: I'm removing the entire C API in following changes, so this will go away.

PR #1127, opened 2022-06-16
Disable Go modules explicitly

yalcinmelihyasin:


yalcinmelihyasin:
The bug is fixed by itself

PR #1134, opened 2022-06-23
Implement parsing vullkan image formats

yalcinmelihyasin:


vulkan_generator/vulkan_parser/formats_parser.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    Parser for the image format planes if format is multi-planar
+
+    Sample image format plane:
+    <plane index="0" widthDivisor="1" heightDivisor="1" compatible="VK_FORMAT_R8_UNORM"/>
+    """
+    planes: List[types.ImageFormatPlane] = []
+
+    for plane in format_element:
+        if plane.tag == "spirvimageformat":
```
purvisa-at-google-com, line 37: This code does nothing except communicate to the user that parsing this data is handled elsewhere. I can imagine in the future someone adds more parsing code to the "elsewhere" and doesn't add similar no-op code here leaving the two out of sync. Because the no-op code does nothing, this may never be noticed if it happens. Maybe remove the no-op code here and just add a comment saying something like "Not all tags are handled in this switch statement. Tags not handled here are parsed in the image format component parser"?

purvisa-at-google-com: Alternatively maybe we should call the other parsers from here?

purvisa-at-google-com (approved):

vulkan_generator/vulkan_parser/formats_parser.py
yalcinmelihyasin: So there is a problem there: You can either handle or cases(sometimes do nothing for it) or only handle the cases are relevant.


I picked the first one because the only time that we will change the code is when a new XML is released. Therefore I want to pick up if a new tag(a different kind of component) is added. So removing these checks would mean we cannot raise exception anymore. That's a choice as well. It's more about which one we pick. We can talk about it if you want to.

PR #1137, opened 2022-06-24
Update UI to display different instructions for different error codes

ttanatb:
Failed to validate trace file:
Screen Shot 2022-06-24 at 5 25 32 PM

**Consistently** fails to validate trace file (after 3 tries):
Screen Shot 2022-06-24 at 5 16 58 PM

Precondition error:
Screen Shot 2022-06-24 at 5 55 47 PM

Other kinds of error
Screen Shot 2022-06-24 at 5 57 32 PM

Validation Passes:
Screen Shot 2022-06-24 at 5 21 55 PM


ttanatb:
Also ran `gapic/src/main/com/google/gapid/views/DeviceValidationView.java` through `google-java-format`

gapic/src/main/com/google/gapid/views/DeviceValidationView.java
```diff
@@ -15,175 +15,304 @@
...
-    });
 
-    models.devices.validateDevice(dev, this::setValidationStatus);
-  }
+    // Separate error text from help text as a workaround to enable proper text wrapping.
+    errorMessageGroup =
+        withLayoutData(
+            createGroup(this, ""), withSpans(new GridData(SWT.FILL, SWT.TOP, true, false), 3, 1));
+    Text errText =
+        withLayoutData(
```
yalcinmelihyasin, line 225: Nit: Is this line break after assignment a java style?

yalcinmelihyasin (approved):

gapic/src/main/com/google/gapid/views/DeviceValidationView.java
ttanatb: According to the Google Java style guide:
`When a line is broken at an assignment operator the break typically comes after the symbol, but either way is acceptable.`

I think there's nothing prohibiting `withLayoutData` to be on the same line as `Text errText =` but I ran the java formatter and this is what it produced

PR #1138, opened 2022-06-25
Enable Adreno GPU counters

ttanatb:
Would like input/feedback on where this code should be called


core/os/android/adb/device.go
```diff
@@ -412,6 +420,33 @@ func (b *binding) IsLocal(ctx context.Context) (bool, error) {
...
+	}
+
+	// Early exit if the file has already been edited.
+	if fileContent != "0" {
+		return nil
+	}
+
+	_, err = b.Shell(fmt.Sprintf("echo 1 > %s", counterFilePath)).Call(ctx)
+	return err
+}
```
pmuetschard, line 448: I'd just do this to keep it short, it's a work-around after all:
```
b.Shell(fmt.Sprintf("[ -w %s ] && echo 1 > %s", counterFilePath, counterFilePath))
```

pmuetschard (approved):
I think this is the right place.

PR #1139, opened 2022-06-27
Use OrderedDict instead of List, Dict pair.

yalcinmelihyasin:
Remove OrderedDict for basetypes which does not require ordering


purvisa-at-google-com (approved):

yalcinmelihyasin:
MacOS fail is irrelevant

PR #1143, opened 2022-06-29
Linter integration

yalcinmelihyasin:
Integrate Autopep8, Flake8 and mypy to our build system.


BUILD.bazel
```diff
@@ -347,15 +347,24 @@ test_suite(
...
+        # __END_LINT
+    ],
+)
 
-        # Tests
+test_suite(
+    name = "tests-vulkan-generator",
+    tests = [
+        # Run the linting as part of Vulkan generator tests
+        "//:lint",  # DO NOT REMOVE!
```
purvisa-at-google-com, line 367: Do we need a comment saying do not remove?

kokoro/presubmit/presubmit.sh
```diff
@@ -85,12 +86,41 @@ function run_buildozer() {
...
 
+function run_enumerate_lints() {
+  # Get all the linting test targets in Vulkan generator
+  TARGETS=$($BAZEL query 'attr(generator_function, py_lint, //vulkan_generator/...)')
+
+  # Exclude the sub targets to prevent repetation
+  TARGETS=$(echo "$TARGETS" | grep -v "_flake8" | grep -v "_mypy" | grep -v "_pylint")
+
+  # Sort the linting targets alphabetically
+  TARGETS=$(echo "$TARGETS" | sort -t: -k1,1)
```
purvisa-at-google-com, line 114: Can we factor this common stuff out somehow?

tools/build/python/lint_flake8.py
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+from flake8.main import cli as flake8_cli
+
+
+if __name__ == "__main__":
+    flake8_cli.main(sys.argv[1:])
```
purvisa-at-google-com, line 20: new line

purvisa-at-google-com (approved):

purvisa-at-google-com (approved):

purvisa-at-google-com (approved):

BUILD.bazel
yalcinmelihyasin: Well, it is very easy to remove a test from bazel and not catch its missing. I just want to ensure that no one comes and says "why this is here" and remove it without thinking the implications. Having it there has no harm.

kokoro/presubmit/presubmit.sh
yalcinmelihyasin: I was thought of but almost every function there is a singular presubmit case. So the benefit of splitting it have not justified break the pattern for me. Instead I put them in a way to make the process clearer for both case.

PR #1145, opened 2022-07-04
Implement parsing vulkan core features

yalcinmelihyasin:


yalcinmelihyasin:
Updated to https://github.com/google/agi/pull/1147

PR #1148, opened 2022-07-06
Create DEVDOC.md

yalcinmelihyasin:
Document the steps(Python) to be able to develop the Vulkan generator.


ttanatb (approved):
Would note that a lot of our users are windows users, and i'm not sure how much of this translates to windows

ttanatb:
Also not sure if devs will make it to this page, might want to add something to the presubmit section here https://github.com/google/agi/blob/master/DEVDOC.md#setup-to-run-presubmit-tests-locally

yalcinmelihyasin:
I added a reference to the DEVDOC.md in the common folder. Please see if it's ok

PR #1149, opened 2022-07-07
remove the unnecessary vulkan_utils package

yalcinmelihyasin:


vulkan_generator/vulkan_parser/BUILD.bazel
```diff
@@ -15,11 +15,27 @@
 load("@rules_python//python:defs.bzl", "py_library")
 load("//tools/build:rules.bzl", "py_lint")
 
+py_library(
+    name="vulkan_parser_internal",
+    srcs = glob(
+        ["*.py"],
+        exclude = ["types.py", "parser.py",],
```
purvisa-at-google-com, line 22: should we put these files into directories called internal and external or something?

This works, but if you add a new file it will go into include by default. You may not realise you need to add it to the exclude list here

purvisa-at-google-com (approved):

vulkan_generator/vulkan_parser/BUILD.bazel
yalcinmelihyasin: I am doing it in a next RP so this was a prework for that PR. But you are right they should be in the internal folder. You'll see the end result in the next PR.

Issue #1153, opened 2022-07-20
Cannot capture SaschaWillems/Vulkan demos anymore


mikes-lunarg:
**Environment information:**
 - AGI version: master branch
 - Host OS: Linux
 - Device model: Pixel 6
 - Android Version: 12

**Bug description**
Tracing any demo from https://github.com/SaschaWillems/Vulkan crashes at replay. These traced fine until recently. I was able to bisect the crash back to commit c429fc939456b54f79d893914d7b5e51b56ed5b6

**Reproduction steps**
1. Build and install triangle demo from https://github.com/SaschaWillems/Vulkan
2. Manual capture single frame
3. Open capture

**Stacktrace**
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: Panic at command 1:vkWaitForFences(device: 0xd901f880, fenceCount: 1, pFences: 0xfffffffff2526320, waitAll: 1, timeout: 18446744073709551615) → VK_SUCCESS:
runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: 
--- *dependencygraph2.DependencyGraph2Resolvable ---
 capture:{ID:{data:"쮦H\x7f\xecX\x98\xbc\xe7\xe94u\xd9k\xfb8\x91Σ"}}  mergeSubCmdNodes:true
 Store():
  gapis/database/database.go:56  github.com/google/gapid/gapis/database.Build
  gapis/resolve/dependencygraph2/resolvables.go:29  github.com/google/gapid/gapis/resolve/dependencygraph2.GetDependencyGraph
  gapis/server/server.go:232  github.com/google/gapid/gapis/server.(*server).LoadCapture.func1
  core/app/crash/crash.go:65  github.com/google/gapid/core/app/crash.Go.func1

 Build() #0:
  gapis/database/database.go:60  github.com/google/gapid/gapis/database.Build
  gapis/resolve/dependencygraph2/resolvables.go:29  github.com/google/gapid/gapis/resolve/dependencygraph2.GetDependencyGraph
  gapis/server/server.go:232  github.com/google/gapid/gapis/server.(*server).LoadCapture.func1
  core/app/crash/crash.go:65  github.com/google/gapid/core/app/crash.Go.func1
 [recovered]
	panic: 
--- *dependencygraph2.DependencyGraph2Resolvable ---
[...]

PR #1154, opened 2022-07-20
Add VK_EXT_descriptor_indexing

mikes-lunarg:
- Add extension types and enums
- Add props/features queries
- Trace and rebuild VkDescriptorSetVariableDescriptorCountAllocateInfo
  in vkAllocateDescriptorSets
- Trace and rebuild Flags and VkDescriptorSetLayoutBindingFlagsCreateInfo
  in vkCreateDescriptorSetLayout

Tested manually by capturing the following demos:
- https://github.com/google/vulkan_test_applications/tree/master/application_sandbox/descriptor_indexing
- https://github.com/KhronosGroup/Vulkan-Samples/tree/master/samples/extensions/descriptor_indexing
- https://github.com/SaschaWillems/Vulkan/tree/master/examples/descriptorindexing (before c429fc939456b54f79d893914d7b5e51b56ed5b6, see issue #1153)


yalcinmelihyasin (approved):
I have some minor comments otherwise it's fine

gapis/api/vulkan/api/descriptor.api
```diff
@@ -279,6 +297,9 @@ cmd VkResult vkAllocateDescriptorSets(
     if (handle != as!VkDescriptorSet(0)) {
       object := new!DescriptorSetObject(device, handle,
         info.descriptorPool)
+      if as!u32(len(variableDescriptorCounts.Val)) > i {
+        object.VariableDescriptorCount = variableDescriptorCounts.Val[i]
```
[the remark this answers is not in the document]
mikes-lunarg: Yes, the spec is pretty clear that this is invalid

> If descriptorSetCount is not zero, descriptorSetCount must equal VkDescriptorSetAllocateInfo::descriptorSetCount

I'll add a check before the fence

gapis/api/vulkan/api/bitfields.api
```diff
@@ -368,11 +368,17 @@ type VkFlags VkPipelineLayoutCreateFlags
...
+  // Vulkan 1.2 core
+  VK_DESCRIPTOR_SET_LAYOUT_CREATE_UPDATE_AFTER_BIND_POOL_BIT = 0x00000002,
+}
 type VkFlags VkDescriptorSetLayoutCreateFlags
 
 @unused
 bitfield VkDescriptorPoolCreateFlagBits {
   VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT = 0x00000001,
+  // Vulkan 1.2 core
+  VK_DESCRIPTOR_POOL_CREATE_UPDATE_AFTER_BIND_BIT_EXT = 0x00000002,
```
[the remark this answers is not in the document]
mikes-lunarg: Will do

gapis/api/vulkan/api/vk12structs.api
```diff
@@ -260,3 +260,132 @@ class VkSubpassDescriptionDepthStencilResolve {
...
+
+class VkDescriptorSetLayoutBindingFlagsCreateInfo {
+    VkStructureType                 sType
+    const void*                     pNext
+    u32                             bindingCount
+    const VkDescriptorBindingFlags* pBindingFlags
+}
+
+@internal
+class DescriptorSetLayoutBindingFlagsCreateInfo {
```
pmuetschard, line 279: Drop the `CreateInfo` part of the name.

gapis/api/vulkan/errors.api
```diff
@@ -57,6 +57,7 @@ extern void vkErrInvalidImageSubresource(VkImage img, string subresourceType, u3
 extern void vkErrImageMemoryNotBound(VkImage img)
 extern void vkErrSemaphoreNotSubmitted(VkSemaphore semaphore)
 extern void vkErrInvalidDescriptorCopy(VkDescriptorSet srcSet, u32 srcBinding, VkDescriptorSet dstSet, u32 dstBinding)
+extern void vkErrInvalidStructMember(string structName, string structMember)
```
pmuetschard, line 60: I don't understand this error message. From where this error is used, it sound like the value is bad, but we are only showing the name of the field? I think it makes sense to add a specific error message for this case and show both the expected value and the passed in value. I think we've been trying to use more, but specific errors in favor of less, but more general errors.

gapis/api/vulkan/api/vk12structs.api
yalcinmelihyasin: Wouldn't it be confusing with bitmask `DescriptorSetLayoutBindingFlags`

gapis/api/vulkan/errors.api
mikes-lunarg: I don't like it either. It's awkward because the string formatting is separated from the error location so each unique error message has to be plumbed through a handful of files. I was trying to make something that could be reused but I agree it is almost too vague to be useful. What about something like this:

```suggestion
extern void vkErrInvalidValueU32(string name, u32 expected, u32 actual)
```

I'm second-guessing if we really want this error at all. It isn't fatal to the capture or replay operations and putting API validation in the .api files feels out of place.

gapis/api/vulkan/api/vk12structs.api
pmuetschard: you mean `VkDescriptorSetLayoutBindingFlags`, I presume? I mean, the confusion is no more between `VkDescriptorSetLayoutBindingFlags` and `DescriptorSetLayoutBindingFlags` as is between `VkDescriptorSetLayoutBindingFlagsCreateInfo` and `DescriptorSetLayoutBindingFlags`, no? Looking at the rest of the api files, I suppose there isn't really that consistent of a pattern, but a somewhat broadly used pattern would be `DescriptorSetLayoutBindingFlagsInfo`, let's go with that one then.

gapis/api/vulkan/errors.api
pmuetschard: Yeah, I think you're right, drop the error all-together. We've only really been minimal validation, such as null checks, so let's not start with this one. In general, it's a requirement that the app has no validation errors for tracing.

gapis/api/vulkan/api/descriptor.api
```diff
@@ -256,12 +268,18 @@ cmd VkResult vkAllocateDescriptorSets(
...
   if info.pNext != null {
     numPNext := numberOfPNext(info.pNext)
     next := MutableVoidPtr(as!void*(info.pNext))
     for i in (0 .. numPNext) {
       sType := as!const VkStructureType*(next.Ptr)[0]
       switch (sType) {
+        case VK_STRUCTURE_TYPE_DESCRIPTOR_SET_VARIABLE_DESCRIPTOR_COUNT_ALLOCATE_INFO: {
+          ext := as!VkDescriptorSetVariableDescriptorCountAllocateInfo*(next.Ptr)[0]
+          variableDescriptorCounts.Val = ext.pDescriptorCounts[0:ext.descriptorSetCount]
+          read(ext.pDescriptorCounts[0:ext.descriptorSetCount])
```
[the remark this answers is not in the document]
pmuetschard: I don't know what you mean with first or second? do you mean whether the read should be before the assignment to `variableDescriptorCounts.Val`? if so, it doesn't matter, since the first one is just a pointer assignment. In reality, the read is going to be unnecessary, because we'll dereference `variableDescriptorCounts.Val` below, but it makes sense here, since relying on the derefs below may leave holes in the array.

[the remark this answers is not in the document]
mikes-lunarg: In my testing I was missing memory observations without the explicit read, I suspect because the `variableDescriptorCounts.Val` slice isn't accessed until after `fence`

gapis/api/vulkan/api/vk12structs.api
mikes-lunarg: `DescriptorSetLayoutBindingFlagsCreateInfo` is not a great name, but it is consistent with the analogous vulkan struct
 `VkDescriptorSetLayoutBindingFlagsCreateInfo`. I can drop the `Create` part to make it less wordy if there is a strong preference.

gapis/api/vulkan/api/descriptor.api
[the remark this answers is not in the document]
pmuetschard: hmm... I guess the automatic fence placement misses this read, probably due to the indirection.

gapis/api/vulkan/api/vk12structs.api
pmuetschard: I see the pattern of `FooCreateInfo` becomming `FooInfo` in other places, so let's follow that.

PR #1164, opened 2022-08-02
Corrected the Power Rail Tracks Graph

sonakshisaxena1:
The power rail tracks initially were unaffected by the values as a wrong `min` variable was used to calculate the next Y coordinate required in plotting. 

Before:

Screenshot from 2022-08-02 12-35-21

After: 

Screenshot from 2022-08-02 12-44-48

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
@@ -95,7 +95,10 @@ public class CounterPanel extends TrackPanel<CounterPanel> implements Selectable
       }
 
       CounterInfo counter = track.getCounter();
-      double min = counter.range.min, range = counter.range.range();
+
+      //TODO: Move all the calculations below in the backend
+      double min = counter.min;
+      double range = counter.range.range() - min;
 
       Selection<?> selected = state.getSelection(Selection.Kind.Counter);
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
```diff
@@ -95,7 +95,10 @@ protected void renderTrack(RenderContext ctx, Repainter repainter, double w, dou
       }
 
       CounterInfo counter = track.getCounter();
-      double min = counter.range.min, range = counter.range.range();
+
+      //TODO: Move all the calculations below in the backend
+      double min = counter.min;
+      double range = counter.range.range() - min;
```
purvisa-at-google-com, line 101: Does this imply range is actually "max - 0" instead of "max - min"?

purvisa-at-google-com (approved):

yalcinmelihyasin (approved):

gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
pmuetschard: `Range.range()` does return `max - min`. `range.min`, which was used before instead of `counter.min` was `max(counter.min, 0)` so that the bottom of the y-axis is never > zero. This was done on purpose, so I think this change is not the right change, at least not for all counters.

PR #1160, opened 2022-07-25
Fixed Power Rail Track Names

sonakshisaxena1:
This commit beautifies the way the names of power rail tracks get displayed. 

Before 

Screenshot 2022-07-25 at 4 16 41 pm

Now 

Screenshot 2022-07-25 at 4 18 01 pm

PR #1165, opened 2022-08-04
Modify Power Rail Names in the UI instead of backend

sonakshisaxena1:
This change moves the modification of Power Rail Track names done in #1160  from backend to the UI to avoid unnecessary creation of `CounterInfo` object.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
@@ -63,7 +63,9 @@ public class CounterPanel extends TrackPanel<CounterPanel> implements Selectable
   @Override
   public String getTitle() {
     CounterInfo info = track.getCounter();
-    if (info.type == CounterInfo.Type.Gpu && "gpufreq".equals(info.name)) {
+    if (info.name.startsWith("power.rails")) {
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
```diff
@@ -63,7 +63,10 @@ public CounterPanel copy() {
   @Override
   public String getTitle() {
     CounterInfo info = track.getCounter();
-    if (info.type == CounterInfo.Type.Gpu && "gpufreq".equals(info.name)) {
+    if(info.name.startsWith("power.rails")) {
+      return info.name.replace("power.rails.","").replace(".", "-");
+    }
+    else if (info.type == CounterInfo.Type.Gpu && "gpufreq".equals(info.name)) {
```
pmuetschard, line 69: Please follow Google Java style: space after `if` and no newline between closing brace and `else`/`else if`.

pmuetschard (approved):
Just a small nit.

gapic/src/main/com/google/gapid/perfetto/views/CounterPanel.java
sonakshisaxena1: My bad. Fixed it

PR #1169, opened 2022-08-09
Fix some Adreno device not showing up in the list

ttanatb:
This occurs in Adreno devices that restrict permission for the gpu
counter sysfs node. Makes sure that the device shows up in the
list and that error is printed instead.


kocdemir (approved):
LGTM

PR #1171, opened 2022-08-09
Fix device validation of desktop devices.

pmuetschard:


ttanatb (commented):
Should we return that desktop devices pass validation or should it instead be that they don't? What about fuchsia in `gapis/trace/fuchsia/trace.go` ?

kocdemir:
Hi @pmuetschard , @ttanatb , is this something we should merge for the next release?

kocdemir:
Hi @rosasco-wk , do you have any interest in this PR?

rosasco-wk:
Yes.  Right now, I see an error in the UI in the trace dialog that I believe is related to this fix.

What all does device validation verify if you know offhand ?  Depending on the answer to this, I can respond to the Fuchsia case more easily.

kocdemir:
Device validation normally checks if the device supports the required profiling data, such as GPU counters, to make sure the tool will work as expected and the profiling session will produce valid results. Are you using skip-device-validation flag? Looks like a similar update is required on the Fuchsia side, with ideally some real validation checks later on. 
@ttanatb can we put this one in for the next release?

rosasco-wk:
On Tue, Jan 10, 2023 at 6:41 AM kocdemir ***@***.***> wrote:

> Device validation normally checks if the device supports the required
> profiling data, such as GPU counters, to make sure the tool will work as
> expected and the profiling session will produce valid results. Are you
> using skip-device-validation flag? Looks like a similar update is required
> on the Fuchsia side, with ideally some real validation checks later on.
> @ttanatb <https://github.com/ttanatb> can we put this one in for the next
> release?
>

Ah, thanks for that detail.  In this phase of development I won't be doing
any GPU counter work because Fuchsia doesn't currently support GPU counters
via perfetto without considerable changes.  So the right thing for Fuchsia
for the time being is to consider Fuchsia devices as not supporting GPU
counters - just Vulkan tracing for now.

As an aside, it would be nice if Validate() was renamed
SupportsGPUCounters() or something similar unless I'm oversimplifying.

kocdemir:
> it would be nice if Validate() was renamed SupportsGPUCounters() or something similar

We're doing some actual validation over the counter data as well, so I think the naming is alright here for the time being. Inside Android validator we have separated functions for ValidateGpuSlices, ValidateVulkanEvents..etc but in the future, it can be separated into different parts in a higher level (e.g. SupportsGPUCounters, SupportsPerfetto, SupportsPowerProfiling...etc) for actual consumers of the data. Currently it's not very clear which parts won't work properly without the 'full' validation.

ttanatb (approved):

PR #1172, opened 2022-08-10
Added Power Usage Summary Track and corresponding graph

sonakshisaxena1:
Currently, System Profiler displays the raw power utilized by different power rail tracks individually. This change introduces a new power summary track, which depicts the combined power used by all the tracks in that trace. It also displays a graph based on the aggregated values.

Power Summary Graph

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (C) 2022 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.google.gapid.perfetto.models;
 
 import static com.google.gapid.perfetto.models.QueryEngine.createSpan;
@@ -28,8 +44,8 @@ public class PowerSummaryTrack extends Track.WithQueryEngine<PowerSummaryTrack.D
   private static String viewSql;
   private static int numPowerRailTracks;
   public static Unit unit;
-  public static Double minValue;
-  public static Double maxValue;
+  public double minValue;
+  public double maxValue;
 
   private static final String VIEW_SQL_MONOTONIC =
       "select ts + 1 ts, lead(ts) over win - ts dur, lead(value) over win - value value, lead(id)"
@@ -62,6 +78,8 @@ public class PowerSummaryTrack extends Track.WithQueryEngine<PowerSummaryTrack.D
 
     // Power Rails tracks.
     for (CounterInfo powerRail : powerRails) {
+      powerSummaryTrack.minValue = Math.min(powerSummaryTrack.minValue, powerRail.min);
+      powerSummaryTrack.maxValue = Math.max(powerSummaryTrack.maxValue, powerRail.max);
       unit = powerRail.unit;
       CounterTrack powerRailTrack = new CounterTrack(data.qe, powerRail);
       data.tracks.addTrack(
@@ -74,37 +92,30 @@ public class PowerSummaryTrack extends Track.WithQueryEngine<PowerSummaryTrack.D
               true));
     }
 
-    Group.UiFactory ui =
-        group(
-            state ->
-                new PowerSummaryPanel(state, powerSummaryTrack, POWER_RAIL_COUNTER_TRACK_HEIGHT),
-            false);
+    powerSummaryTrack.minValue = Math.min(0, powerSummaryTrack.minValue);
+    Group.UiFactory ui = group(state -> new PowerSummaryPanel(state, powerSummaryTrack), false);
     data.tracks.addLabelGroup(null, powerSummaryTrack.getId(), "Power Usage", ui);
     return data;
   }
 
   private static String buildViewSql(List<CounterInfo> powerRails) {
-    minValue = powerRails.get(0).min;
-    maxValue = powerRails.get(0).max;
     StringBuilder sb =
         new StringBuilder()
-            .append("select aggregateTable.ts ts, sum(aggregateTable.value) totalValue from")
-            .append(" (")
-            .append(VIEW_SQL_MONOTONIC)
-            .append(powerRails.get(0).id)
-            .append(" window win as (order by ts) ")
-            .append(") as aggregateTable");
+            .append(
+                String.format(
+                    "select aggregateTable.ts ts, sum(aggregateTable.value) totalValue,"
+                        + " aggregateTable.dur dur from (%s%d window win as (order by ts) ) as"
+                        + " aggregateTable",
+                    VIEW_SQL_MONOTONIC, powerRails.get(0).id));
+
     for (int i = 1; i < numPowerRailTracks; i++) {
-      sb.append("left join (")
-          .append(VIEW_SQL_MONOTONIC)
-          .append(powerRails.get(i).id)
-          .append(" window win as (order by ts) ")
-          .append(") ");
-      minValue = Math.min(minValue, powerRails.get(i).min);
-      maxValue = Math.max(maxValue, powerRails.get(i).max);
+      sb.append(
+          String.format(
+              "left join (%s%d window win as (order by ts) )",
+              VIEW_SQL_MONOTONIC, powerRails.get(i).id));
     }
+
     sb.append(" group by ts");
-    minValue = Math.min(0, minValue);
     return sb.toString();
   }
 
--- a/gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
@@ -1,0 +1,170 @@
+package com.google.gapid.perfetto.models;
+
+import static com.google.gapid.perfetto.models.QueryEngine.createSpan;
+import static com.google.gapid.perfetto.models.QueryEngine.createView;
+import static com.google.gapid.perfetto.models.QueryEngine.createWindow;
+import static com.google.gapid.perfetto.models.QueryEngine.dropTable;
+import static com.google.gapid.perfetto.models.QueryEngine.dropView;
+import static com.google.gapid.perfetto.views.StyleConstants.POWER_RAIL_COUNTER_TRACK_HEIGHT;
+import static com.google.gapid.perfetto.views.TrackContainer.group;
+import static com.google.gapid.perfetto.views.TrackContainer.single;
+import static com.google.gapid.util.MoreFutures.transform;
+import static com.google.gapid.util.MoreFutures.transformAsync;
+import static java.lang.String.format;
+
+import com.google.common.collect.ImmutableListMultimap;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.gapid.models.Perfetto;
+import com.google.gapid.perfetto.Unit;
+import com.google.gapid.perfetto.models.TrackConfig.Group;
+import com.google.gapid.perfetto.views.CounterPanel;
+import com.google.gapid.perfetto.views.PowerSummaryPanel;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/** {@link Track} summarizing the total Power usage. */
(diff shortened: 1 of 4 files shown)


gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
```diff
@@ -0,0 +1,184 @@
...
+    this.numPowerRailTracks = numTracks;
+  }
+
+  public int getNumPowerRailTracks() {
+    return numPowerRailTracks;
+  }
+
+  public static Perfetto.Data.Builder enumerate(Perfetto.Data.Builder data) {
+    ImmutableListMultimap<String, CounterInfo> counters = data.getCounters(CounterInfo.Type.Global);
+    List<CounterInfo> powerRails =
```
ttanatb, line 65: Is the count of this list always expected to be equal to numPowerRailTracks

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
```diff
@@ -0,0 +1,184 @@
...
+              true));
+    }
+
+    Group.UiFactory ui = group(state -> new PowerSummaryPanel(state, powerSummaryTrack), false);
+    data.tracks.addLabelGroup(null, powerSummaryTrack.getId(), "Power Usage", ui);
+    return data;
+  }
+
+  private static String buildViewSql(List<CounterInfo> powerRails) {
+    minValue = powerRails.get(0).min;
```
ttanatb, line 99: Maybe, instead get the `minValue` and `maxValue` in the loop in line 80 instead. It seems to be iterating over the same data, but it's a little confusing to me for this function named `buildViewSql` to both build a sql query and also get the proper min/max value

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
```diff
@@ -0,0 +1,184 @@
...
+  private ListenableFuture<Data> computeData(DataRequest req, Window win) {
+    return transform(
+        qe.query(sql()),
+        res -> {
+          int rows = res.getNumRows();
+          if (rows == 0) {
+            return Data.empty(req);
+          }
+
+          Data data = new Data(req, new long[rows + 1], new double[rows + 1]);
```
ttanatb, line 154: Why is there a `rows + 1` here? Would prefer comment clarifying magic numbers (and the `rows -1` below

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
```diff
@@ -0,0 +1,184 @@
...
+  protected ListenableFuture<Data> computeData(DataRequest req) {
+    Window win = Window.quantized(req, 5);
+    return transformAsync(win.update(qe, tableName("window")), $ -> computeData(req, win));
+  }
+
+  private ListenableFuture<Data> computeData(DataRequest req, Window win) {
+    return transform(
+        qe.query(sql()),
+        res -> {
+          int rows = res.getNumRows();
```
ttanatb, line 149: nit: prefer rowCount to clarify that it's count of rows

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
```diff
@@ -0,0 +1,279 @@
...
+import com.google.gapid.perfetto.models.Selection;
+
+public class PowerSummaryPanel extends TrackPanel<PowerSummaryPanel> {
+  private static final double HOVER_MARGIN = 10;
+  private static final double HOVER_PADDING = 4;
+  private static final double CURSOR_SIZE = 5;
+  private static Unit unit;
+
+  protected final PowerSummaryTrack track;
+  protected static final double trackHeight = 80;
```
ttanatb, line 38: Why does this variable use a different format to the other `final` vars?

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
```diff
@@ -0,0 +1,279 @@
...
+          double range =
+              (track.minValue == track.maxValue && track.minValue == 0)
+                  ? 1
+                  : (track.maxValue - track.minValue);
+
+          Selection<?> selected = state.getSelection(Selection.Kind.Counter);
+          mainGradient().applyBaseAndBorder(ctx);
+          ctx.path(
+              path -> {
+                double lastX = state.timeToPx(data.ts[0]), lastY = h;
```
ttanatb, line 93: Would prefer each variable initialization to be in separate lines (especially since there's a level of complexity here)
```
double lastX = state.timeToPx(data.ts[0]);
double lastY = h;
```

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
```diff
@@ -0,0 +1,279 @@
...
+                  lastY = nextY;
+                }
+
+                path.lineTo(lastX, h);
+                ctx.fillPath(path);
+                ctx.drawPath(path);
+              });
+
+          if (hovered != null) {
+            double y = (trackHeight - 1) * (1 - (hovered.value - track.minValue) / range);
```
ttanatb, line 111: nit: The formula used to derive `y` here and in line 98 (`(trackHeight - 1) * (1 - (x - track.minValue) / range)`) looks like it could be extracted to its own private function just to make it easier to parse (since it's also used in line 98)

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
```diff
@@ -0,0 +1,279 @@
...
+      return Hover.NONE;
+    }
+
+    long time = state.pxToTime(x);
+    if (time < data.ts[0]) {
+      return Hover.NONE;
+    }
+
+    int idx = 0;
+    for (; idx < data.ts.length - 1; idx++) {
```
ttanatb, line 165: Why is everything offset by one here?

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
```diff
@@ -0,0 +1,279 @@
...
+        break;
+      }
+    }
+
+    if (idx >= data.ts.length) {
+      return Hover.NONE;
+    }
+
+    double startX = state.timeToPx(data.ts[idx]);
+    double endX = (idx >= data.ts.length - 1) ? startX : state.timeToPx(data.ts[idx + 1]);
```
ttanatb, line 176: This could use a comment as to why `endX` is set to `startX`

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
```diff
@@ -0,0 +1,279 @@
...
+        // Determine the x-axis coordinate of the right boundary of the redrawn area by comparing
+        // the right end of the sample with the right boundary of the default redrawn area.
+        double redrawRx = Math.max(defaultX + defaultW, endX);
+
+        // Calculate the real redrawn width.
+        double redrawW = redrawRx - redrawLx;
+
+        if (defaultX >= state.getWidth() - defaultW) {
+
+          // re-calculate the left boundary of the redrawn area by comparing the start end of the
```
ttanatb, line 208: Capitalize first word, same in line 213

ttanatb, line 208: nit: `comparing the start of the` (remove end)

ttanatb (approved):

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/** {@link Track} summarizing the total Power usage. */
+public class PowerSummaryTrack extends Track.WithQueryEngine<PowerSummaryTrack.Data> {
+  private static String viewSql;
+  private static int numPowerRailTracks;
+  public static Unit unit;
+  public static Double minValue;
+  public static Double maxValue;
```
pmuetschard, line 48: Avoid mutable static state. This would break if multiple traces are loaded. These fields are only used from an instance, so just make them instance fields.

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
```diff
@@ -0,0 +1,184 @@
...
+            .append(" (")
+            .append(VIEW_SQL_MONOTONIC)
+            .append(powerRails.get(0).id)
+            .append(" window win as (order by ts) ")
+            .append(") as aggregateTable");
+    for (int i = 1; i < numPowerRailTracks; i++) {
+      sb.append("left join (")
+          .append(VIEW_SQL_MONOTONIC)
+          .append(powerRails.get(i).id)
+          .append(" window win as (order by ts) ")
```
pmuetschard, line 115: If you use the `String.format` pattern as used elsewhere in the Perfetto UI, you can have the entire query in the constant.

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
sonakshisaxena1: Yes, we are assigning the length of the list to `numPowerRailTracks`

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/** {@link Track} summarizing the total Power usage. */
+public class PowerSummaryTrack extends Track.WithQueryEngine<PowerSummaryTrack.Data> {
+  private static String viewSql;
+  private static int numPowerRailTracks;
+  public static Unit unit;
+  public static Double minValue;
```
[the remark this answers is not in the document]
sonakshisaxena1: I agree that we don't need to use the `Double` class here as we don't expect to have a NULL value. I'll update that.

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
sonakshisaxena1: Yep. I should have moved it to Line 80. Thanks for pointing that out.

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
sonakshisaxena1: That's how it has been referred to in other tracks - [CounterTrack](https://github.com/google/agi/blob/90c1fc8732d10b7878703f31fcc1d695f9d3bd77/gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java#L152) .  I think I would let it this way be for uniformity.

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
sonakshisaxena1: Could be. Again, that's how the calculation has been done in other Views, but personally I agree that it would be neater to have a function. I can made that change.

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
sonakshisaxena1: Done, thanks for the catch!

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
sonakshisaxena1: Makes sense, updated.

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
sonakshisaxena1: Cool

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
sonakshisaxena1: Will add.

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
sonakshisaxena1: Will add this as a comment

gapic/src/main/com/google/gapid/perfetto/models/PowerSummaryTrack.java
sonakshisaxena1: Makes sense.

gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
sonakshisaxena1: Nothing specific. Missed to resolve this, will update now.

PR #1173, opened 2022-08-10
Enable per process power data hardware support detection.

sonakshisaxena1:
We want to help the users see the per process power usage by the trace. This change enables AGI to detect if the energy breakdown per process is supported by the hardware.


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -1019,6 +1026,9 @@ private void updateBat() {
       if (batPowerRail != null) {
         batPowerRail.setEnabled(enabled);
       }
+      if(batEnergyBreakDown != null) {
```
ttanatb, line 1029: space between `if` and `(`

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -776,6 +778,7 @@ public BasicInputArea(Composite parent, Settings settings, Theme theme,
       batLabels[1] = createLabel(batGroup, "ms");
       if (caps.getHasPowerRail()) {
         batPowerRail = createCheckbox(batGroup, "Collect Power Rails", sBatt.getCollectPowerRail());
+        //TODO: @nexa to add the checkbox (b/241746026)
```
ttanatb, line 781: `// TODO(b/241746026): ... `is the normal format for adding TODOs (there's no auto syntax for GitHub but useful for other stuff

ttanatb (approved):

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
sonakshisaxena1: Got it, thanks

PR #1174, opened 2022-08-11
Added Build Config and build script for Angle

sonakshisaxena1:


kokoro/angle/build.sh
```diff
@@ -0,0 +1,143 @@
...
+"is_official_build = true"
+"proprietary_codecs = true"
+"symbol_level = 2"
+"target_cpu = \"arm64\""
+"target_os = \"android\""
+)
+
+
+function setup_python_virtual_env {
+  pyenv uninstall -f "$1"
```
yalcinmelihyasin, line 50: Can we not use the pyenv in the system instead of reinstalling?

kokoro/angle/build.sh
```diff
@@ -0,0 +1,143 @@
...
+### Check if depot_tools exists othrewise Download Chromium's depot_tools
+cd $KOKORO_ARTIFACTS_DIR
+ls depot_tools
+if [[ $? -ne 0 ]]
+then
+  git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+fi
+
+if [[ $? -ne 0 ]]
+then
```
yalcinmelihyasin, line 78: As this repeated a few times, you may consider to move this to a function.

yalcinmelihyasin (approved):
I have two minor comment, otherwise LGTM.

kokoro/angle/build.sh
sonakshisaxena1: If we move this to a function, we'll have to report back a generic failure error instead of a specific one. I think it's helpful to have specific error statements to spot failures in the build script if any.

kokoro/angle/build.sh
sonakshisaxena1: I tried but Kokoro was throwing errors. The only workaround I was able to find was this.

PR #1189, opened 2022-08-25
Update Devdoc

yalcinmelihyasin:
Clarify how to setup Python and its tools.

### DIFF
--- a/vulkan_generator/DEVDOC.md
+++ b/vulkan_generator/DEVDOC.md
@@ -5,7 +5,7 @@ This document is to be able to build, run and test the Vulkan Generator in AGI.
 
 ## Pre-requirements:
 
-Vulkan generator requires Python 3.9.2 Install Python 3.9.2 and pyenv.
+Vulkan generator requires Python 3.9.2 Install Python 3.9.2 with an environment(e.g pyenv, virtualenv)
 
 Then add these to `.bashrc` or `.zshrc`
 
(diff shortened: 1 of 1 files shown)


vulkan_generator/DEVDOC.md
```diff
@@ -59,6 +59,12 @@ Presubmit can be run from `./kokoro/presubmit/presubmit.sh`. This will require t
 autopep8==1.6.0
 ```
 
+Note: If you do not have a python environment, ensure that autopep8 is in your path. For default Pyhon installation:
+
+```
+export AUTOPEP8=~/.local/bin/autopep8
```
kocdemir, line 65: We need a better error message when autopep8 cannot be found on the system, and ideally default to python installation automatically. 
Also this document should focus on vulkan generator development related items only, currently this is a core requirement for the presubmit tests and such information should be under compiling.md or agi/devdoc.md

kocdemir (changes requested):

vulkan_generator/DEVDOC.md
yalcinmelihyasin: We cannot change the error message as it's given by the bash. 

We can restructure the doc later if we want. But I think we should at least land this document so that other people in the team can run presubmit.

kocdemir: I've made some changes for presubmit tests in PR #1199, this one can be closed now.

PR #1201, opened 2022-09-01
Clang update reorder

yalcinmelihyasin:
This PR updates a bunch of dependency, adds support for Apple M1 and switches Linux compiler from GCC-9 to Clang-12


tools/build/third_party/breakpad.bzl
```diff
@@ -23,16 +23,16 @@ def _breakpad_impl(repository_ctx):
     )
     repository_ctx.symlink(repository_ctx.attr.build_file, "BUILD")
 
-    if repository_ctx.os.name.startswith("windows"):
-      # Patch up breakpad on windows and add the dump_syms src.
-      repository_ctx.symlink(Label("@gapid//tools/build/third_party/breakpad:windows.patch"), "windows.patch")
-      repository_ctx.symlink(Label("@gapid//tools/build/third_party/breakpad:dump_syms_pe.cc"), "src/tools/windows/dump_syms/dump_syms_pe.cc")
+    # if repository_ctx.os.name.startswith("windows"):
```
kocdemir, line 26: Do we need a todo or an explanation note for the commented out section here?

gapir/cc/surface.cpp
```diff
@@ -168,13 +168,15 @@ void handleWindow(uint32_t width, uint32_t height) {
   }
 }
 
+static const int32_t stream_index = 0;
```
kocdemir, line 171: can't see where this is used, remove?

kocdemir (commented):
Changes LGTM, I'll test this on different machines for any compilation issues, bugs and stability.

tools/build/third_party/breakpad.bzl
yalcinmelihyasin: I will remove this. There used to be a patch but I think the patch is not needed as we updated the version and the issue is solved by the package itself. Thanks for picking!

gapir/cc/surface.cpp
yalcinmelihyasin: If I remember correctly this was one of the suspects in a problem that I had while trying to compile. I will retry and see if it's actually needed or not. Because it's static I can't be 100% sure.

tools/build/third_party/breakpad.bzl
yalcinmelihyasin: Done!

gapir/cc/surface.cpp
yalcinmelihyasin: Done!

kocdemir (approved):
LGTM, seems to be working fine on m1 and linux. I believe we can fix the remaining issues, if there are any, after merging this.

PR #1204, opened 2022-09-02
Update fuchsia_sdk.bzl

mangini:
Remove cipd_tag and sha, so the versions of these repos are rolled automatically.

### DIFF
--- a/tools/build/fuchsia/enabled/fuchsia_sdk.bzl
+++ b/tools/build/fuchsia/enabled/fuchsia_sdk.bzl
@@ -22,11 +22,6 @@ def fuchsia_sdk_dependencies(locals = {}):
     fuchsia_sdk_repository,
     name = "fuchsia_sdk_dynamic",
     locals = locals,
-    cipd_tag = "version:7.20220228.3.1",
-    sha256 = {
-      "linux": "7f886893f5758272da4f51aced35448fa83805aa0f2b363f2e538db4cf936462",
-      "mac": "e18ea6c85b436146ce9f93f6abcea35baea2ba23f9544a87e7b6f5632ade374e",
-    },
   )
   native.register_toolchains("@fuchsia_sdk_dynamic//:fuchsia_toolchain_sdk")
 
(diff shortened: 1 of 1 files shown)


yalcinmelihyasin:
I am not sure if this is a good idea. I think we should update these manually instead of removing the check sum

mangini:
Just to clarify, since I have not explained this well. This is not removing the checksum in the security sense. If not specified, the version of the fuchsia SDK is the one checked in and validated in the sdk-integration.git repository (Fuchsia Bazel SDK). It will be downloaded using the cipd binary tool and will have its checksum validated.

In this repo you already specify a [checked-in version of the sdk-integration repository](https://github.com/google/agi/blob/master/tools/build/fuchsia/enabled/workspace.bzl#L24) that you have to update manually and this CL is not removing that.

In the past, when the Bazel SDK was first introduced, you had to specify both the sdk-integration git commit and the Fuchsia SDK CIPD instance ID, which is redundant and error-prone, as most combinations of Fuchsia SDK and Bazel SDK other than the one checked-in will not work. A couple of months ago we moved to a model where the sdk-integration repository has a [manifest file](https://fuchsia.googlesource.com/sdk-integration/+/refs/heads/main/bazel_rules_fuchsia/fuchsia/manifests/core_sdk.resolved) and downstream users of the Bazel SDK just specify the git commit of the sdk-integration repository and get the appropriate Fuchsia SDK version.

You can still override the cipd instance id as you currently do, but this may cause several build problems if it is updated without upding to the corresponding Bazel SDK, which is the reason why I sent this CL (John was having trouble updating the Fuchsia SDK only without realizing that the Bazel SDK had to be upgraded too).

So, if this PR is merged, there is no automatic roll or anything on that sort. It will just remove the n x m mapping of Fuchsia SDK to Bazel SDK, and convert it to a 1:1 checked-in and valdidated mapping.

Let me know if you have more questions, happy to help.

yalcinmelihyasin (approved):
LGTM

PR #1211, opened 2022-09-13
Enable AGI's UI to collect energy breakdown using Trace Dialog

sonakshisaxena1:
Enable the users to collect energy breakdown in AGI's System Profiler using Trace Dialog configuration
Screenshot 2022-09-13 at 10 43 27 am


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -769,16 +769,19 @@ public BasicInputArea(Composite parent, Settings settings, Theme theme,
...
+      Composite batLabelGroup = withLayoutData(
+          createComposite(batGroup, withMargin(new GridLayout(3, false), 5, 0)),
           withIndents(new GridData(), GROUP_INDENT, 0));
-      batLabels[0] = createLabel(batGroup, "Poll Rate:");
-      batRate = createSpinner(batGroup, sBatt.getRate(), 250, 60000);
-      batLabels[1] = createLabel(batGroup, "ms");
+      batLabels[0] = createLabel(batLabelGroup, "Poll Rate:");
+      batRate = createSpinner(batLabelGroup, sBatt.getRate(), 250, 60000);
+      batLabels[1] = createLabel(batLabelGroup, "ms");
       if (caps.getHasPowerRail()) {
```
ttanatb, line 782: It might make more sense to change this so that the checkboxes are always added to the UI, but if power rail is not available:
- we gray out the checkboxes
- and add a label saying something like "This device does not provide information on power consumption and energy breakdown" (feel free to edit to make the flow feel right)

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -769,16 +769,19 @@ public BasicInputArea(Composite parent, Settings settings, Theme theme,
...
         batPowerRail = createCheckbox(batGroup, "Collect Power Rails", sBatt.getCollectPowerRail());
-        // TODO(b/241746026): Add the checkbox for Energy Breakdown
+        batEnergyBreakDown = createCheckbox(batGroup, "Collect Energy Breakdown", sBatt.getCollectEnergyBreakdown());
```
ttanatb, line 784: nit: make sure that the order of checkboxes match the order of rows in the system profile itself

ttanatb (approved):

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
kocdemir: I've recently added a new version of createCheckbox function with a tooltipText, can you please add extra information about this option as a tooltip. Similarly, a tooltip can be added to 'Collect Power Rails' option, to clarify the differences between the two. Feel free to add versions of other widget creation functions with tooltips in order to make newly added UI bits more understandable for the end user.

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
sonakshisaxena1: `updateBat()` keeps track on these. Majority of OEMs don't provide any information on Power Rails/Energy Breakdown so it was decided not to show those checkboxes instead. (Although the greyed boxes do look better!) I'll discuss on this further and get it in another PR, if necessary.

PR #1212, opened 2022-09-13
Display Energy Breakdown Graph in System Profiler's UI

sonakshisaxena1:
Displays a new section under Battery, to depict the energy consumed by different energy consumers, and different processes (uids) under them. 

Screenshot 2022-09-14 at 2 35 52 pm

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
@@ -32,10 +32,10 @@ public class PowerSummaryPanel extends TrackPanel<PowerSummaryPanel> {
   private static final double HOVER_MARGIN = 10;
   private static final double HOVER_PADDING = 4;
   private static final double CURSOR_SIZE = 5;
+  private static final double TRACK_HEIGHT = 80;
   private static Unit unit;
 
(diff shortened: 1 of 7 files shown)


gapic/src/main/com/google/gapid/perfetto/views/PowerSummaryPanel.java
```diff
@@ -0,0 +1,284 @@
...
+import com.google.gapid.perfetto.models.Selection;
+
+public class PowerSummaryPanel extends TrackPanel<PowerSummaryPanel> {
+  private static final double HOVER_MARGIN = 10;
+  private static final double HOVER_PADDING = 4;
+  private static final double CURSOR_SIZE = 5;
+  private static Unit unit;
+
+  protected final PowerSummaryTrack track;
+  protected static final double trackHeight = 80;
```
ttanatb: Does this need to be protected? It feels like it should be formatted similarly to `HOVER_MARGIN` or `HOVER_PADDING`

sonakshisaxena1: `trackHeight` has already been modified in https://github.com/google/agi/pull/1176/files as per your suggestions. I'll rebase this to include those changes. And I agree that it can be made private, so I'll make that change

PR #1216, opened 2022-09-15
handle missing energy tracks case

kocdemir:
System profiler is not working correctly since we didn't test the dependency update branch with the code submitted earlier. It crashes when a perfetto file is loaded without any energy tracks, or a new trace is performed. This CL adds a size check for energy tracks to get things running again. a better fix and UI can be added later. 

Screen Shot 2022-09-15 at 22 37 52


sonakshisaxena1:
Oh, missed this! LGTM now. Thanks for the fix Serdar!

sonakshisaxena1 (approved):

PR #1225, opened 2022-10-11
Loosen requirement for % compute time to be non-negative instead of strictly 0

ttanatb:
Based on back-and-forth in b/242828399


kocdemir (approved):
thanks!

PR #1227, opened 2022-10-12
An attempt to establish go mods throughout AGI and changing namespace…

GOOG-sergiu:
… (gapid => agi)

Note: go mod only indexes packages which have at least a go file declared. As such there is a need to add dummy.go files in .proto services. Although it might appear hacky, this is actually standard practice across many go repos.

When go mod changes we'll need to run:
`bazel run //:gazelle -- update-repos -from_file=go.mod`

UPDATE: ^ doing this in presubmit.sh solves this issue. Let me know if there are any concerns with this.

Not sure if there's a hook or a way to trigger bazel to do that on build?


kokoro/presubmit/presubmit.sh
```diff
@@ -123,7 +123,7 @@ function run_enumerate_lints() {
 
 function run_gazelle() {
   echo # TODO: figure out a way to make bazel not print anything.
-  $BAZEL run gazelle
+  $BAZEL run gazelle -- update-repos -from_file=go.mod
```
GOOG-sergiu, line 126: Modified the presubmit to read from go.mod

PR #1261, opened 2023-01-05
Add info in GPU Counter failures, misc fixes to validation

ttanatb:
- Fixed generic validator check. 
  - Previously it required every value to be over 0, but now it checks to make sure that not all of the value is 0.
- Made sure to include trace result in validation error.
- Adjusted the GPU counter check to list out missing counter (not just ones that failed the check). 
  - Made the logs way more expressive as well. 
  - Also trimmed duplicates from the check (querying counters can return duplicate counters as data sources).

Changing ones of the check in Adreno validation
Screenshot 2023-01-05 at 10 09 12 PM

Exynos currently doesn't have proper support for GPU profiling, this reports a full list of counters we couldn't trace
Screenshot 2023-01-05 at 10 10 43 PM


yalcinmelihyasin (approved):

yalcinmelihyasin:
LGTM but please rebase on latest master for the build.

PR #1277, opened 2023-03-07
Setup trace ffx unix pipe

rosasco-wk:
Create unix pipe and complete connectivity between AGI and running Fuchsia Vulkan component.  Launching component on Fuchsia will be handled in a succeeding change.  With this change, the header handshake between AGI and the Fuchsia component is working.

Pascal, this review will require more attention than prior changes.  I've noted areas where I'm looking for input if you have any.

### DIFF
--- a/cmd/agi/main.go
+++ b/cmd/agi/main.go
@@ -30,7 +30,7 @@ import (
 const (
 	versionPrefix = `version "`
 	googleInfix   = "-google-"
-	minJavaMajor  = 17
+	minJavaMajor  = 11
 	minJavaMinor  = 0
 	colorReset    = "\033[0m"
 	colorYellow   = "\033[33m"
--- a/cmd/agi/main.go
+++ b/cmd/agi/main.go
@@ -30,7 +30,7 @@ import (
 const (
 	versionPrefix = `version "`
 	googleInfix   = "-google-"
-	minJavaMajor  = 11
+	minJavaMajor  = 17
 	minJavaMinor  = 0
 	colorReset    = "\033[0m"
 	colorYellow   = "\033[33m"
--- a/core/cc/fuchsia/zircon_socket_connection.cpp
+++ b/core/cc/fuchsia/zircon_socket_connection.cpp
@@ -18,12 +18,17 @@
 
 #include "zircon_socket_connection.h"
 
-#include <thread>
-
 #include <errno.h>
 #include <zircon/status.h>
 
+#include <thread>
+
 #include "core/cc/log.h"
+
+#define ERR_IF_COND(cond, ...) \
+  if ((cond)) {                \
+    GAPID_ERROR(__VA_ARGS__);  \
+  }
 
 namespace core {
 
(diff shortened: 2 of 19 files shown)


cmd/agi/main.go
```diff
@@ -30,7 +30,7 @@ import (
 const (
 	versionPrefix = `version "`
 	googleInfix   = "-google-"
-	minJavaMajor  = 11
+	minJavaMajor  = 17
```
pmuetschard, line 33: You can't do that without also updating the build scripts to upgrade the bundled JRE. For Windows and MacOS we bundle the JRE with the application, look at the build scripts in the kokoro directory. For Linux, the .deb file should depend on a package providing the version of the JRE we need.

core/cc/fuchsia/zircon_socket_connection.cpp
```diff
@@ -39,11 +41,25 @@ size_t ZirconSocketConnection::send(const void* data, size_t size) {
...
+          "Failed to read %zu bytes from Zircon socket: %s, attempts: %d", size,
+          zx_status_get_string(status), attempts);
+      attempts++;
+    }
+    if (status == ZX_ERR_SHOULD_WAIT) {
+      GAPID_INFO("Zircon socket read: ZX_ERR_SHOULD_WAIT, waiting 6s ...");
+      std::this_thread::sleep_for(std::chrono::milliseconds(6000));
+    }
+  } while (status == ZX_ERR_SHOULD_WAIT);
+  // TODO(rosasco): Pascal, have any comments on how the above block functions?
```
pmuetschard, line 59: I don't know Fuchsia, but a few Google searches seem to show that you should use `zx_object_wait_one` https://fuchsia.dev/fuchsia-src/reference/syscalls/object_wait_one to wait until data is available?

gapic/src/main/com/google/gapid/models/TraceTargets.java
```diff
@@ -188,6 +188,7 @@ public boolean isTraceable() {
     }
 
     public Target getTraceTarget() {
+      System.out.println("GetTraceTarget, uri is: " + data.getTraceUri());
```
pmuetschard, line 191: remove or use the logger instead.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -190,16 +190,23 @@ public static void showTracingDialog(
...
+
       Tracer.Trace trace = Tracer.trace(client, shell, input.getValue(), progress);
+
       progress.setTrace(trace);
+
       if (progress.open() == Window.OK && progress.successful()) {
         models.capture.loadCapture(input.getValue().output);
       }
+    } else {
+      System.out.println("ERROR: loadDevicesAndShowDialog status is not OK");
```
pmuetschard, line 209: remove or use the logger. This just means that the user clicked cancel, BTW.

gapii/cc/spy.cpp
```diff
@@ -172,8 +172,10 @@ Spy::Spy()
 #else
     mConnection = ConnectionStream::listenSocket("127.0.0.1", "9286");
 #endif                                          // TARGET_OS
-    if (mConnection->write("gapii", 5) != 5) {  // handshake string
+    if (mConnection->write("gapii", 6) != 6) {  // handshake string
```
pmuetschard, line 175: `gapii` is only 5 letters and the other side is still only expecting 5 bytes here.

gapii/client/capture.go
```diff
@@ -193,21 +202,35 @@ func (p *Process) Capture(ctx context.Context, start task.Signal, stop task.Sign
 		}
 	}()
 
-	if p.conn == nil {
+	// TODO(rosasco): Pascal, why was this block conditionalized on p.Conn being nil ?
+	//                This assumes if p.Conn is initialized rather than the port, that the
+	//                header has already been sent.  Supposedly, it's ok to initialize either
+	//                p.Conn or p.Port prior to calling this Capture() method.
+	//                The Process struct is defined in "adb.go".
+	//                Is "p.Port == 0" a reasonable condition here (it is true on Fuchsia) ?
```
pmuetschard, line 210: Conclusion: Since you renamed the field and I don't see anywhere in this change where it used to be initialized, as long as everything builds, it means that we are safe and you can just remove the if.

I don't know. Maybe there used to be some codepath that initialized the process struct with a connection that is ready to trace? If that exists, I don't know if `p.Port == 0` is an ok check to ignore the existing connection and try to read the header.

Maybe the if was just defensive in an attempt to support a (future) case where the connection is already initialized - or a left over of a no longer supported past case. If so, you can just remove the if and "change" it so that it means that the connection is ready, but still want to read the header, which makes sense to me.

gapii/client/protocol.go
```diff
@@ -45,15 +46,23 @@ func readHeader(conn net.Conn) (msgType messageType, dataSize uint64, err error)
 	now := time.Now()
 	conn.SetReadDeadline(now.Add(time.Millisecond * 500)) // Allow for stop event and UI refreshes.
 	buf := make([]byte, messageHeaderSize)
-	if _, err = io.ReadFull(conn, buf); err == nil {
+	n := 0
+
+	if n, err = io.ReadFull(conn, buf); err == nil {
+		fmt.Printf("readHeader: Read %d bytes: %s", n, buf)
```
pmuetschard, line 52: remove or use the logger

gapii/client/protocol.go
```diff
@@ -45,15 +46,23 @@ func readHeader(conn net.Conn) (msgType messageType, dataSize uint64, err error)
...
+
 		// first header byte contains the message type
 		msgType = messageType(buf[0])
 		// next messageDataBytes contain the data size as little-endian unsigned integer
 		for i := uint(0); i < messageDataBytes; i++ {
 			dataSize += uint64(buf[i+1]) << (i * 8)
 		}
+	} else {
+		fmt.Printf("ERROR readHeader: Invalid message header.  Only %v bytes read and %d expected.", n,
```
pmuetschard, line 61: remove or use the logger

gapis/trace/fuchsia/BUILD.bazel
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2018 Google Inc.
+# Copyright (C) 2022 Google Inc.
```
pmuetschard, line 1: why? again below.

pmuetschard (commented):
Besides the JRE change, looks ok to me.

I don't mind the extra logging, but can you change the log messages to be a bit more readable and understandable without context? e.g. I wouldn't know what this meant if I saw it in the log: `GENERAL handleCommError POST <some err>`.

ttanatb (approved):

PR #1315, opened 2023-08-29
Update pytest version to 7.2.0

yalcinmelihyasin:


kocdemir (changes requested):
Let's fix the compilation issues on other platforms as well before submitting this one. Thank you!

kocdemir (approved):
Submitting this as some security vulnerabilities detected with the usage of py library with version 1.11.0. CI failures are related to the configuration changes of the build machines and can be dealt with later on.

PR #1355, opened 2024-08-08
Fix Linux build with clang

olehkuznetsov:
Fixes support of newer clang compilers with partial updated dependencies
Works with clang-13 and newer


sonakshisaxena1 (approved):
Tested it locally by updating to JAVA_11 and using `bazelisk build pkg`
LGTM!

PR #1380, opened 2024-10-18
Set AGI version to 3.3.2

olehkuznetsov:


m--koma (commented):
How do we decide which version this should be? 3.3.2 looks like a downgrade - where did it come from?

m--koma (approved):

